Issue #60, opened 2014-12-31
approx_eq_ulps() ?


mikedilger:
Choosing a good epsilon for comparing floats is tricky.   Floats close to 1.0 and floats way out there should probably be compared using quite different epsilons.   Leaving that up to the library consumer is too much of a punt, IMHO.

I think it's more useful to specify the maximum allowable number of floating-point representations separating the two numbers being compared, otherwise known as ULPs (units in the last place, or units of least precision).   IEEE754 data is stored so that when interpreted as 2-s complement integers, sort in the same order.  So casting to integer, and subtracting, can give you the number of representations separating two floats.

I've used something like this (f32 example):

```rust
impl FloatApproxCmp<i32> for f32 {
    fn ulps(self, rhs: f32) -> i32
    {
        // Setup integer representations of the input
        let ai32: i32 = unsafe { mem::transmute(self) };
        let bi32: i32 = unsafe { mem::transmute(rhs) };
        ai32 - bi32
    }

    fn approx_cmp(self, rhs: f32, ulps: i32) -> i32
    {
        // Get exact equality out of the way.
        // This nicely also handles -0 compared to +0
        if self==rhs { return 0; }

        // Handle differing signs as a special case, even if
        // they are very close
        if self>0_f32 && rhs<0_f32 { return 1; }
        if self<0_f32 && rhs>0_f32 { return -1; }

        let diff: i32 = self.ulps(rhs);
        match diff {
            x if x.abs() <= ulps => 0,
            x if x < 0 => -1,
            x if x > 0 => 1,
            _ => 0
        }
    }

    fn approx_eq(self, rhs: f32, ulps: i32) -> bool
    {
        let c :i32 = self.approx_cmp(rhs, ulps);
        c==0
    }
}
```

@aepsil0n

aepsil0n:
I think this is a very reasonable change for floating point numbers. I must admit, that I've been using some workaround to achieve something numerically similar in my projects using `nalgebra` (mostly multiplying epsilon by one of the numbers to compare, but the ULPs seem more efficient and and more adequate for floating point math).

As far as this library is concerned, whose main applications probably use some floating point type, I think we can do it like that. However, considering more abstract mathematical objects we should see how this relates to more general metric spaces.

sebcrozet:
That's an interesting approach. Though I am not sure we should do this by default (i-e. in `.approx_eq`) instead of the common epsilon-based solution as I think most people will have the latter in mind when using it.

PR #61, opened 2014-12-31
ApproxEq trait enhanced with ULPs method of specifying closeness:

mikedilger:
See issue #60.
Background reading:  https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

  approx_eq_ulps() allows specification of epsilon as an integer number
  of Units in the Last Place (ULPs) difference between the two floating
  point values

  default approx_ulps() is set to 8.

  approx_eq() function continues to use epsilon method, although I
  recommend further commits and a migration towards the ULPs method.


sebcrozet:
Looks good. See #60 for my comment about mighating `approx_eq` to the ulps method.

PR #65, opened 2015-01-03
Fixes for upstream

mikedilger:
Fixes some.
Punts w/ old_orphan_check on others, due to associated types not quite being ready.
Still not compiling, needs some thinking about 'Sized'

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -92,6 +92,8 @@ extern crate "rustc-serialize" as rustc_serialize;
 extern crate test;
 
 use std::cmp;
+use std::ops::Neg;
+
 pub use traits::{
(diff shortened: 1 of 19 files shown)


sebcrozet:
Thanks! Though I am going to merge #64 that was submitted first and makes things compile again. The `Sized` constraints it includes are not too restrictive (before this, we assumed everything was `Sized` anyway, so it should not break anything).

For the thinking about `Sized` I prefer to delay that until 1.0.alpha (next week) lands so that we don't have to face such a huge amount of breakage while thinking (and I will have more time to think about it too).

mikedilger:
Cool.

PR #70, opened 2015-01-07
Change Fn to FnMut

bfops:
`Fn` was the wrong way to go because a closure like `|y| x = y` isn't a valid `Fn`, but it's a valid `FnMut` (because it captures the mutable external state of `x`).


sebcrozet:
Ah, that's right. Thanks!

PR #74, opened 2015-01-10
Implement Arbitrary for (almost) all types

aepsil0n:
All types in nalgebra can now be used for quickcheck-style testing as an optional feature. `Rot4` and `Iso4` had to be left out, since `Rot4` apparently lacks a working constructor.

The test suite could, in principle, now be migrated to use quickcheck to, as this helps to detect and prevent a large range of bugs.

### DIFF
--- a/src/structs/ortho.rs
+++ b/src/structs/ortho.rs
@@ -2,6 +2,10 @@ use std::num;
 use traits::structure::BaseFloat;
 use structs::{Pnt3, Vec3, Mat4};
 
+#[cfg(feature="arbitrary")]
+use quickcheck::{Arbitrary, Gen};
+
+
 /// A 3D orthographic projection stored without any matrix.
 ///
 /// Reading or modifying its individual properties is cheap but applying the transformation is costly.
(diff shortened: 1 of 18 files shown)


src/structs/ortho.rs
```diff
@@ -232,3 +247,20 @@ impl<N: BaseFloat + Clone> OrthoMat3<N> {
...
+    }
+}
+
+
+/// Simple helper function for rejection sampling
+#[cfg(feature="arbitrary")]
+#[inline]
+pub fn reject<G: Gen, F: FnMut(&T) -> bool, T: Arbitrary>(g: &mut G, f: F) -> T {
+    use std::iter::repeat;
+    repeat(()).map(|_| Arbitrary::arbitrary(g)).filter(f).next().unwrap()
```
sebcrozet: This is a very interesting way of selecting the first valid element!

sebcrozet:
`Rot4` and `Iso4` don't have any constructor at the moment (they need to be build from pairs of quaternions) so it's OK to ignore them for now.

Thanks!

src/structs/ortho.rs
aepsil0n: Functional programming style… I just hope the compiler strips out all the fluff it generates. Otherwise this could be easily converted to a loop.

PR #75, opened 2015-01-10
fix warning from unneeded old_orphan_check

blevin:
trivial fix to remove deprecation warning in rustc 1.0.0-nightly (44a287e6e 2015-01-08 17:03:40 -0800):

src/lib.rs:86:12: 86:28 warning: feature is deprecated and will only be available for a limited time, please rewrite code that relies on it
src/lib.rs:86 #![feature(old_orphan_check)]
                         ^~~~~~~~~~~~~~~~


sebcrozet:
Thanks! I’m surprised this was unneeded.

PR #77, opened 2015-01-16
impl BaseNum for integers

arturoc:
BaseNum is described in the comment as basic integral numeric trait but it's only implemented for floats. also fixes the comment that was broken in a PR i sent some days ago with a find and replace for int -> isize


sebcrozet:
Correct. Thanks!

PR #83, opened 2015-01-25
Implement AssertEq for reference types

aepsil0n:
Just as the standard library's PartialEq is implemented for reference types, the ApproxEq trait should be implemented on them as well. This is mostly an ergonomic improvement for certain testing situations, where a method yields a reference. For non-copy types it allows using the assert_approx_… macros, which would otherwise not be possible.


sebcrozet:
Thanks!

PR #86, opened 2015-02-01
Rename similar in-place and copy methods

### LINKED ISSUE #52: API design of similar mutating and copy operations
There are some trait methods in `nalgebra` that show a certain symmetry: one method takes `&self` and returns a new object of the same type, while another takes `&mut self` and turns the object itself into the same thing that the other function would have returned. Examples for this are `Norm::normalize_cpy` and `Norm::normalize` or `Inv::inv_cpy` and `Inv::inv`.

Now the issue that I see there is that normally Rust goes for immutability by default, whereas this naming scheme implies that the mutable version is the more common one. Using the mutable version should be a conscious choice. You are mutating state that semantically does not have to be mutable, so this is typically a performance or memory optimization of some sort.

Of course, this is a bit of bike-shedding, but I find a naming scheme such as `Norm::normalize`/`Norm::normalize_self` would imho be more appropriate.

aepsil0n:
Fixes #52.

This renames pairs of methods like `normalize_cpy`, `normalize` to `normalize`
and `normalize_mut`. Note that the previous in-place operations had the same
name that now belongs to the copy operations. Therefore this is a breaking
change.

On a similar note, the `Quat::conjugate` method was also renamed to
`conjugate_mut` and a new copying method has taken its place. This appears to
be a similar issue (apart from the lack of the copy operation) and thus was
dealt with here, too.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,7 +31,7 @@ fn main() {
     let     a = Vec3::new(1.0f64, 1.0, 1.0);
     let mut b = Rot3::new(na::zero());
 
-    b.append_rotation(&a);
+    b.append_rotation_mut(&a);
 
     assert!(na::approx_eq(&na::rotation(&b), &a));
 }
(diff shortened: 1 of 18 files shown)


sebcrozet:
Thanks!

PR #87, opened 2015-02-03
Workaround internal compiler error

aepsil0n:
The associated type bound here is not interpreted correctly. This is likely due to the ICE described in rust-lang/rust#20413.

Note, that this may lead to a loss of type inference using that function.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,7 +98,7 @@ extern crate quickcheck;
 extern crate test;
 
 use std::cmp;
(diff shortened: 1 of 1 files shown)


aepsil0n:
The failure is an upstream bug in quickcheck. Build needs to be triggered again.

sebcrozet:
Thanks! I don't think that loosing type inference temporarily for this function is too problematic.

PR #93, opened 2015-02-16
Fix deprecated plugin syntax.

bfops:
This makes the test build with `rustc 1.0.0-nightly (b63cee4a1 2015-02-14 17:01:11 +0000)` for me.

### DIFF
--- a/tests/assert.rs
+++ b/tests/assert.rs
@@ -1,8 +1,8 @@
 //! Assertion macro tests
 
 #![feature(plugin)]
+#![plugin(nalgebra)]
 
-#[plugin]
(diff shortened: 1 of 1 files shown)


aepsil0n:
Hm… when I look at this, I don't quite understand, why this has been added in the first place. There is no need for plugins in the module at all, there are only macros.

bfops:
Ah you're right, it can just be removed. But it looks like the build is failing otherwise anyway..

aepsil0n:
Ok, I sat down fixing the rest of the issues. Thus closing in favour of #94.

Issue #85, opened 2015-02-01
#[repr(C)] for structs?


jarrett:
Should the vector and matrix structs have `#[repr(C)]`? Without that, I suspect the compiler could [rearrange the memory layout of the structs](http://doc.rust-lang.org/book/ffi.html#interoperability-with-foreign-code). That would cause weird and hard-to-debug errors when passing pointers to C APIs like OpenGL. E.g. when calling `glBufferData` or `gluniformmatrix4fv`.

I apologize if I'm mistaken about this. I'm new to Rust.

aepsil0n:
As far as I understand the code, the C ordering is intended and some unsafe parts of the implementation might even rely on it. So I am generally in favour. #97 even includes a fix for that.

cc @sebcrozet.

sebcrozet:
Yes, methods performing slice conversion assume the fields are ordered as they are declared. So I guess `#[repr(C)]` is needed. And I agree this is a must-have to ensure it would work well with e.g. opengl.

sebcrozet:
Fixed by #99.

PR #97, opened 2015-03-01
Nightly fixes, add repr(C)

jpernst:
Minor fix to macros that use `all`.
Changes deprecated suffixes in tests from `us` to `usize`.
Adds `#[repr(C)]` to structs that are most likely to be used in FFI scenarios (fixed size vectors, points, quats, and mats). It's unlikely that the compiler would reorder the fields when they all have the same size, but it doesn't hurt to be explicit.

### DIFF
--- a/src/structs/dmat.rs
+++ b/src/structs/dmat.rs
@@ -618,13 +618,13 @@ impl<N: ApproxEq<N>> ApproxEq<N> for DMat<N> {
 
     #[inline]
     fn approx_eq_eps(&self, other: &DMat<N>, epsilon: &N) -> bool {
-        let zip = self.mij.iter().zip(other.mij.iter());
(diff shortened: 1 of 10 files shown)


aepsil0n:
Thanks for the fixes!

I get the point about the FFI ordering, but I feel like it should be a separate PR. This is also because I would merge the fixes instantly, but would like @sebcrozet to look over the FFI part. There was an issue (namely #85) about this earlier btw.

sebcrozet:
The ffi part looks good except that to fix #85, `#[repr(C)]` should be added to `Iso` and `Rot` structures as well.

And as felt by @aepsil0n, it would be better to put thore ffi-related fixes on a separate PR too (so we can link it to the relevant issue without including unrelated compilation fixes).

jpernst:
Closing this to PR to split the changes into separate PRs

PR #98, opened 2015-03-01
Nightly fixes

jpernst:
Nightly fixes from my previous PR without the FFI stuff. Also fixes the deprecated closure syntax that just showed up in the latest nightly.


sebcrozet:
Thanks!

PR #99, opened 2015-03-01
Add [repr(C)] to structs likely to be used in FFI

jpernst:
Same as from previous PR, but now also includes `Iso` and `Rot`. Addresses #85. Depends on #98.


sebcrozet:
Thanks!

PR #110, opened 2015-03-31
Add type annotation required by rustc 2015-03-31

andersk:
Resolves this error, which is fallout from
https://github.com/rust-lang/rust/pull/23673:

```
src/structs/dmat.rs:501:43: 501:74 error: type annotations required: cannot resolve `<f64 as core::ops::Div<_>>::Output == f64` [E0284]
src/structs/dmat.rs:501         let normalizer: N    = Cast::from(1.0f64 / Cast::from(self.nrows));
                                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Signed-off-by: Anders Kaseorg <andersk@mit.edu>


tomaka:
Could this be uploaded on crates.io please? :)

Issue #113, opened 2015-04-03
trait core::ops::Neg not implemented for unsigned types


BradLyman:
Build fails on my system with the error that the trait core::ops::Neg is not implemented for the unsigned types (u8, u16, etc,...).
Removing the implementation of the BaseNum trait for these types seems to fix the problem, but I am not sure if this is the _right_ way to fix the problem.

Issue #32, opened 2014-11-16
Investigate moving to brendanzab/algebra as a basis for nalgebra


brendanzab:
We are currently in the process of simplifying the numeric traits in the standard libraries (see rust-lang/rfcs#369). Perhaps we could move to [algebra](https://github.com/brendanzab/algebra/) as a basis for nalgebra?

sebcrozet:
I started working on that.

This is currently being done in parallel to the improvement of the `algebra` crate itself.
A few of those traits have been implemented in d29ff5332930afdd3b1d468e712891da62fab3ef

brendanzab:
I wonder if it will be possible to allow for better integration between cgmath and nalgebra then... even then, perhaps I might consider switching back if I'm happy enough with the end result :)

sebcrozet:
Since #218 , we use [alga](https://github.com/sebcrozet/alga) as a basis for **nalgebra**. Closing this.

PR #114, opened 2015-04-03
Remove Neg from BaseNum

andersk:
As per rust-lang/rust#23945, `Neg` is no longer implemented for unsigned types.

### DIFF
--- a/src/structs/iso.rs
+++ b/src/structs/iso.rs
@@ -2,7 +2,7 @@
 
 #![allow(missing_docs)]
 
-use std::ops::{Add, Sub, Mul};
+use std::ops::{Add, Sub, Mul, Neg};
 
 use rand::{Rand, Rng};
 use structs::mat::{Mat3, Mat4, Mat5};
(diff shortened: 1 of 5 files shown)


aepsil0n:
I think we should discuss in that case, what BaseNum is supposed to mean. See also #113.

cc @sebcrozet

tomaka:
There are only two possibilities:
 - Remove the implementation of `BaseNum` on unsigned types, which means that you won't be able to have a `Mat4<u16>` for example.
 - Allow `Mat4<u16>`, but forbid all operations that may need to change the sign of the content of the matrix. That's what this PR does.

aepsil0n:
I am perfectly aware of that. The question is whether `u8`, `u16` etc. are supposed to be `BaseNum`. I tend to think that they ought to be as well, since that has been the case before.

But this still changes the abstract definition of `BaseNum`. Personally, I think that the algebraic definitions here should be more fine-grained anyhow. I just can't really judge, what exactly `BaseNum` is supposed to entail.

Anyway, to be practical I could probably merge this to get nalgebra to compile. Then we can think about proper abstractions later, see #32.

tomaka:
Ahhh, yes. Could you publish it too?

aepsil0n:
I'm in the process of fixing everything right now.

PR #115, opened 2015-04-03
FRAC_2_SQRTPI was renamed to FRAC_2_SQRT_PI

andersk:
The old name is deprecated.

Question: Do we want to rename `frac_2_sqrtpi()` to `frac_2_sqrt_pi()` as well for consistency, or keep the old name for compatibility, or add both names and deprecate the old one?


aepsil0n:
I don't think we're currently in a situation, where we keep backwards compatibility. Thus I'd say the method should also be changed for consistency.

PR #117, opened 2015-04-04
Use Float trait from num crate

aepsil0n:


sebcrozet:
Thanks! We might want to use their `Zero`, `One` and `Bounded` traits instead of ours as well.

PR #118, opened 2015-04-05
new dependency num. use traits from num.

johnpmayer:
old trait is deprecated

pretty simple but want your advice on where Clone ought to go (it used to be part of the Float I replaced)


johnpmayer:
duplicate of #117

PR #119, opened 2015-04-05
Feature/norm assoc type

johnpmayer:
Note: This subsumes #118 

This is part of a greater effort to remove old_impl_check from nphysics

Discussion - which other traits are going to be similarly transformed?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name    = "nalgebra"
-version = "0.2.9"
+version = "0.2.10"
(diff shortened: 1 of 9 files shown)


sebcrozet:
I’ve made some work a few weeks ago on using associated types for most trait that may be transformed. You'll find it on the `associated_types` branch. Naming needs to be discussed (currently I kind of followed a common pattern for all names with the name of the trait followed by `Type`). Also all the implementations are commented out at the moment.

I stopped working on it because it ICEd (and still does with the last nightly)...

johnpmayer:
Yep, still ICE

johnpmayer:
Opened https://github.com/rust-lang/rust/issues/24179

PR #124, opened 2015-05-06
Remove operator hacks

aepsil0n:
The LMul, RMul and Scalar* traits were only necessary due to language
limitations regarding trait bounds that are now gone. The Mat trait is now
expressed in terms of regular operator traits.

However, due to the removal of these traits this constitutes a breaking change.


sebcrozet:
We don't need them any more alright. Thanks!

PR #125, opened 2015-05-08
Add neg trait to BaseFloat

tinco:


sebcrozet:
We removed `Neg` from `BaseNum` at first because `uint` doesn't implement it (see #114 ), and I believe it is reasonable to assume real-like types to be able to assume negative values, so I'm OK with those changes.

PR #126, opened 2015-05-08
Rework Norm to use associated type

tinco:
This one builds on my previous PR, needed to make ncollide work in 1.0.0.beta4

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -320,13 +320,13 @@ pub fn center<N: BaseFloat, P: FloatPnt<N, V>, V: Copy>(a: &P, b: &P) -> P {
  */
 /// Returns the distance between two points.
 #[inline(always)]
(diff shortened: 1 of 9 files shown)


tinco:
No need to pull this in yet, it seems there's a whole lot more to it than just Norm.

sebcrozet:
Yes, ncollide requires you to use associated types almost everywhere they should be on nalgebra. You may want to get some inspiration from nalgebra's `associated_types` branch (which ICEs at the moment, and I could not manage to find exactly which piece of code causes this bug).

sebcrozet:
This is a very old PR and `Norm` now uses associated types, so I can close this.

Regarding other traits, it might be worth it to wait and see if we end up integrating [brendanzab/algebra](https://github.com/brendanzab/algebra) eventually.

PR #127, opened 2015-05-09
This reference should not be mut

tinco:
Found this little bug, it's picked up by a new linter in the nightly


Nemikolh:
Same problem here, is this going to be merged anytime soon?

sebcrozet:
Thanks! It is merged though I won't be able to publish it before a couple of days.

sebcrozet:
This version is now published on `crates.io`!

PR #134, opened 2015-06-13
Add 'rustc-serialize' and 'serde-serialize' features.

qrlpx:
This allows users to choose rustc-serialize or serde (or both or none) for serializiation. 
'rustc-serialize' is a default feature, serde is activated through the 'serde-serialize' feature. 
If there's any interest in it, I'll eventually do the same for ncollide and nphysics.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,13 +16,23 @@ name = "nalgebra"
 path = "src/lib.rs"
 
 [features]
+default = ["rustc-serialize"]
+serde-serialize = ["serde", "serde_macros"]
 # Generate arbitrary instances of nalgebra types for testing with quickcheck
 arbitrary = ["quickcheck"]
 
 [dependencies]
-rustc-serialize = "*"
 rand = "*"
 num = "*"
+
+[dependencies.rustc-serialize]
+optional = true
+
+[dependencies.serde]
+optional = true
+
+[dependencies.serde_macros]
+optional = true
 
 [dependencies.quickcheck]
 optional = true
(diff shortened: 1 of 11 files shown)


sebcrozet:
I'm a bit concerned about this feature-based approach.

If two libraries end up using the same version of `nalgebra` but with different  serialization methods (i-e, different features), they won't be able to cooperate smoothly (because cargo will identify both `nalgebra` libraries as different). And I don't feel very comfortable introducing this kind of potential incompatibility for a low level mathematical library like that.

erickt:
serde serializers and deserializers can live side by side with rustc-serialize encoders and decoders, so I don't think these need to be optional. There is some minor trickiness required to `#[derive(Serialize, Deserialize)]` in stable rust, but I'd be happy to help out with that if you're interested.

bluss:
@sebcrozet That's not my impression. Cargo will union the feature flags, and it will be just one compilation of nalgebra. (And features need to be additive for this reason).

bluss:
@erickt They should be optional though, since it slows down the compile too much otherwise?

Boscop:
I'm very interested in serialization support as well!

robo-hamburger:
Now that rust 1.15 is out and rust_derive works on stable this seems like a great idea.

vadixidav:
I am also looking for this right now, particularly now that serde_derive is supported on stable. This PR needs to be modified to use it though.

robo-hamburger:
I have an updated version of the pull request in a branch. Tests seem to pass but I haven't tried using the traits yet: https://github.com/robo-hamburger/nalgebra

sebcrozet:
Thanks for your contributions and sorry I've forgotten about this issue. As part of the massive changes that has been done for #218 (related issue: #207 ), I added the support for serde behind the `serde-serialize` feature. I've removed rustc-serialize though as it seems to be much less practical.

PR #139, opened 2015-06-20
Allow for non-consuming std operations on DMat. Added DMat multiplication test.

mitchmindtree:
Currently, the std operations (`Mul`, `Add` and `Sub`) consume both the lhs and rhs `DMat`s even though it is not always necessary or desired. This means a user must clone the `DMat` on the occasion that they only have access to the `DMat` by reference. As an example, the following isn't allowed:
```Rust
let NeuralNetwork { ref input_mat, ref weight_mat, .. } = *net;
let activity = *input_mat * *weight_mat; // error: cannot move out of borrowed content
```

This PR implements the std operations for `&'a DMat`, allowing the above example to work like so:
```Rust
let NeuralNetwork { ref input_mat, ref weight_mat, .. } = *net;
let activity = input_mat * weight_mat;
```

Specifically, the following are now possible:

```Rust
(&mat_a) * mat_b;
mat_a * (&mat_b);
(&mat_a) * (&mat_b);
mat_a + (&mat_b);
(&mat_a) + mat_b;
mat_a - (&mat_b);
(&mat_a) - mat_b;
```

I hope this is ok! Let me know your thoughts or if you'd prefer me to change anything :+1:

Edit: fixed example.

### DIFF
--- a/src/structs/dmat.rs
+++ b/src/structs/dmat.rs
@@ -265,7 +265,35 @@ impl<N> IndexMut<(usize, usize)> for DMat<N> {
 impl<N: Copy + Mul<N, Output = N> + Add<N, Output = N> + Zero> Mul<DMat<N>> for DMat<N> {
     type Output = DMat<N>;
 
+    #[inline]
     fn mul(self, right: DMat<N>) -> DMat<N> {
+        (&self) * (&right)
+    }
+}
+
+impl<'a, N: Copy + Mul<N, Output = N> + Add<N, Output = N> + Zero> Mul<&'a DMat<N>> for DMat<N> {
+    type Output = DMat<N>;
+
+    #[inline]
+    fn mul(self, right: &'a DMat<N>) -> DMat<N> {
+        (&self) * right
+    }
+}
+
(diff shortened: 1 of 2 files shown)


aepsil0n:
While we're at it, I guess the remaining variants `(&mat_a) + (&mat_b)` and `(&mat_a) + mat_b` and similar for `-` should be implemented, too, even though an allocation is necessary for the former, which is not the case for the other variants.

mitchmindtree:
@aepsil0n oh nice spot! I just added `(&mat_a) + mat_b` and `(&mat_a) - mat_b`.

The reason I haven't implemented `(&mat_a) + (&mat_b)` and `(&mat_a) - (&mat_b)` is to avoid any subtle extra allocations which would be required if taking both rhs and lhs by reference. Perhaps it is ok for a user to have to clone in addition and subtraction, i.e.
```Rust
let Foo { ref mat_a, ref mat_b } = *foo;
mat_a.clone() + mat_b
```
This way it makes the extra allocation more obvious (whereas a multiplication requires a new allocation either way).

sebcrozet:
Seems useful. I am also in favor to not adding `(&mat_a) + (&mat_b)` to prevent non-obvious allocations ; it is easy to forget that we have two references.

Merging, thanks!

PR #143, opened 2015-07-12
Fix multiplication of rectangular DMats

### LINKED ISSUE #142: Multiplication or rectangular matrices panics
Multiplication of a non-square `&DMat` and `DMat` panics with message
```
thread '<main>' panicked at 'assertion failed: self.ncols == right.nrows', /home/arttu/.cargo/registry/src/github.com-0a35038f75765ae4/nalgebra-0.2.19/src/structs/dmat.rs:297
```

For example:
```rust
     let a: DMat<f64> = DMat::new_zeros(1, 2);
     let b: DMat<f64> = DMat::new_zeros(2, 3);
     let c = (&a) * b;
```

aryla:
Fixes #142 and adds a test for it.

### DIFF
--- a/src/structs/dmat.rs
+++ b/src/structs/dmat.rs
@@ -285,11 +285,11 @@ impl<'a, N: Copy + Mul<N, Output = N> + Add<N, Output = N> + Zero> Mul<DMat<N>> 
 
     #[inline]
     fn mul(self, right: DMat<N>) -> DMat<N> {
-        right * self
+        self * (&right)
     }
 }
 
-impl<'a, N: Copy + Mul<N, Output = N> + Add<N, Output = N> + Zero> Mul<&'a DMat<N>> for &'a DMat<N> {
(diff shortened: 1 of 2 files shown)


sebcrozet:
Well spotted!

Though I'd prefer the fix to look more like the other implementations of `Mul` for `DMat`. I believe it could just be: `self * (&right)`. This might require looser lifetime bounds for the implementation on [line 292](https://github.com/sebcrozet/nalgebra/blob/master/src/structs/dmat.rs#L292), i.e `impl<'a, 'b, N: Copy + Mul<N, Output = N> + Add<N, Output = N> + Zero> Mul<&'b DMat<N>> for &'a DMat<N> {` instead of using `'a` twice.

aryla:
I amended the commit to use `self * (&right)`. It indeed required the addition of the `'b` lifetime.

sebcrozet:
Thanks!

PR #146, opened 2015-07-31
add an example cargo.toml dep

jaredly:


README.md
```diff
@@ -12,6 +12,14 @@ nalgebra
 An on-line version of this documentation is available [here](http://nalgebra.org).
 
 ## Using **nalgebra**
+
+Add this to your `Cargo.toml`:
+
+```toml
+[dependencies.nalgebra]
+git = "https://github.com/sebcrozet/nalgebra.git"
```
sebcrozet: I believe we should recommend the version on `crates.io` instead of the github one. So this should be something like:

```toml
[dependencies]
nalgebra = "0.2.*"
```

sebcrozet:
The README now includes the correct example. Closing this.

Issue #50, opened 2014-12-01
Use quickcheck for testing


aepsil0n:
[QuickCheck](https://github.com/BurntSushi/quickcheck) allows to test properties on arbitrary input. This prevents a larger class of bugs than example-based unit testing. My proposal is to use it for nalgebra's test suite. For convenience this involves implementing `quickcheck::Arbitrary` for most types. These impls should be public (i.e. not only compiled for the test suite), so that user code can use these implementations to write own tests that involve nalgebra types.

We may think about making this an optional feature, as it adds a dependency to nalgebra.

btw, I volunteer to do this at some point, unless there are any objections.

TODO
- [x] Implement `quickcheck::Arbitrary` for all types.
- [x] Migrate test suite to use quickcheck.

aepsil0n:
I've started to work on this. It should be done soon.

aepsil0n:
Now I realize how much time has passed, since I posted the last comment. I need to dig up this once again to see how it was going ;)

sebcrozet:
Fixed by #218. We now use quickcheck for testing.

PR #147, opened 2015-08-07
Implemented Cholesky decomposition

dshizzle:
I implemented the Cholesky decomposition for square symmetric positive definite matrices which size is known at compile time. As with the QR decomposition, DMat is not yet supported.

### DIFF
--- a/tests/mat.rs
+++ b/tests/mat.rs
@@ -659,6 +659,19 @@ fn test_cholesky_not_spd() {
 }
 
 #[test]
+fn test_cholesky_not_symmetric() {
+    
+    let a : Mat2<f64> = Mat2::<f64>::new(1.0, 1.0, -1.0, 1.0);
+
+    let result = na::cholesky(&a);
+
+    match result {
+        Ok(_) => assert!(false),
+        Err(_) => assert!(true),
+    }
+}
+
+#[test]
 fn test_cholesky_mat1() {
     test_cholesky_impl!(Mat1<f64>);
 }
--- a/tests/mat.rs
+++ b/tests/mat.rs
@@ -633,16 +633,13 @@ fn test_cholesky_const() {
 
     let result = na::cholesky(&a);
 
-    match result {
-        Ok(v) => {
-                    assert!(na::approx_eq(&v, &g));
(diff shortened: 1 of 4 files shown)


tests/mat.rs
```diff
@@ -600,3 +624,79 @@ fn test_ortho() {
...
+fn test_cholesky_not_spd() {
+    
+    let a : Mat3<f64> = Mat3::<f64>::new(1.0, 2.0, 3.0, 3.0, 2.0, 1.0, 1.0, 1.0, 1.0);
+
+    let result = na::cholesky(&a);
+
+    match result {
+        Ok(_) => assert!(false),
+        Err(_) => assert!(true),
+    }
```
aepsil0n: Perhaps `assert!(result.is_err())` is clearer and more concise here

tests/mat.rs
```diff
@@ -600,3 +624,79 @@ fn test_ortho() {
...
+    match result {
+        Ok(v) => {
+                    assert!(na::approx_eq(&v, &g));
+
+                    let recomp = v * na::transpose(&v);
+                    assert!(na::approx_eq(&recomp, &a));
+                 },
+
+        Err(_) => assert!(false),
+    }
```
aepsil0n: I would just unwrap the result, as you expect it to be `Ok(…)` and then do the inner assertions. Just trying to avoid nesting to make the tests more readable. ;)

aepsil0n:
In principle this looks ok. I think quickcheck-style tests would be great here, to thoroughly test the algebraic property this function should satisfy. Plus, you're testing it with random input anyway. There is already support for this in `nalgebra`, so I think this would be fairly straightforward. Anyhow, I don't see this as a precondition for a merge, but rather an idea for further improvements.

Also I think the `cholesky` function itself is quite long and thus could benefit from some refactoring. Generally, such refactoring should not be a performance issue, as the compiler is quite good at deconstructing abstractions.

sebcrozet:
Looks ok to me. I agree with @aepsil0n regarding quickcheck tests though the whole test suite still needs a migration anyway (cc. #50) so this is not blocking.

Merging, thanks!

PR #151, opened 2015-08-27
Fix wrong allocated vector length in RowSlice for DMat

AndersKaloer:
The RowSlice implementation for DMat allocates and returns a DVec of length self.nrows independent of the size of the slice.
These changes resolve the bug and add test cases for the DMat implementations of RowSlice and ColSlice.


sebcrozet:
Thanks!

PR #152, opened 2015-09-10
Removed Display trait requirement from Debug method for DMat.

Antsiscool:
DMat was unable to use the Debug trait if the value stored inside did not have the Display trait. This meant it could not be displayed or used in an assertion.


sebcrozet:
Thanks!

PR #162, opened 2015-11-14
Add Debug trait to Outer trait

MichaelRiss:
This PR is part of the https://github.com/sebcrozet/nphysics/issues/43 issue.
It enables me to use println! debugging in ncollide, for an example see https://github.com/MichaelRiss/ncollide/tree/with_debug_println.

So I thought, maybe this could be useful for somebody else.
But please be careful, this is probably not the right way to do it. I'm a rust newbie and figuring out where to touch the code to route the Debug trait successfully to the place I needed it, was admittedly banging randomly on the code until it finally worked.

### DIFF
--- a/src/structs/mat.rs
+++ b/src/structs/mat.rs
@@ -19,6 +19,7 @@ use traits::geometry::{ToHomogeneous, FromHomogeneous, Orig};
 use linalg;
 #[cfg(feature="arbitrary")]
 use quickcheck::{Arbitrary, Gen};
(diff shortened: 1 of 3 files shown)


sebcrozet:
Requiring `Debug` for something unrelated (the outer product) seems a bit overkill. For temporary debugging, my guess it that adding the `Debug` constraint on the relevant ncollide functions may be cleaner (though a bit more verbose).
Thanks for the attempt!

PR #163, opened 2015-11-14
Provide test case for a diagonal matrix resulting in NaN values from …

MichaelRiss:
…the QR eigenvalue computation.


MichaelRiss:
Closing as https://github.com/sebcrozet/nalgebra/pull/164 provides a better test case alongside the fix for the problem.

PR #164, opened 2015-11-14
Fix QR algorithm for diagonal matrices

dshizzle:
As pointed out by @MichaelRiss, my previous QR algorithm did not handle a possible division by zero which may arise at computing the eigenvalues of diagonal matrices. This pull request fixes this bug and adds a test case for random diagonal matrices.

This patch should pass the test case provided in #163 and fix the issue presented in sebcrozet/nphysics#43


sebcrozet:
Thanks!

PR #168, opened 2016-01-08
mat_macros: from_homogeneous_impl was using dimension of src not dst

arturoc:
for example:

```rust
let m4: Mat4<f32> = one();
let m3: Mat3<f32> = from_homogeneous(&m4);
```

used 4 as the boundary of the for loop which made it panic


sebcrozet:
Well spotted ! Thanks.

Issue #153, opened 2015-09-10
Implement Row & Col Traits for DMat


jramapuram:
It would be nice to just do ops like:

```rust
mymat.set_row(...);
let col = mymat.col(...);
```

It seems like the trait is implemented for everything besides `DMat`.

Issue #100, opened 2015-03-03
New matrix types and access methods


gnzlbg:
_Partially-dynamic matrices:_
Sometimes when working with dynamic matrices I know e.g. the number of columns but not the number of rows before hand. This can be used to speedup lookup and slices by relying on a compile time constant for offsetting indices. This can also speed matrix slices since it allows to return stack allocated slices instead of dynamic vectors.

_Dynamic matrices of bounded size:_
Sometimes I know the maximum size a matrixis going to have, but not the exact size (it is still dynamic). This knowledge allows matrices to be stack allocated. Is the same argument as for the already implemented dynamic vectors of bounded size.

_Mutable matrix slices_:
Right now the matrix methods `col_slice` and `row_slice` return an immutable `DVec<N>`. It would very useful to have a way to return a mutable slice to a matrix' sub-matrix (rows and cols are two common sub-matrices).

sebcrozet:
The second point is now implemented!

The last point is still interesting and worth the work. But I cannot think of a nice way of implementing the first one at the moment (and it seems less important than the rest).

sebcrozet:
The new version of nalgebra implements the first and third points. The second one will no longer be supported.

PR #169, opened 2016-01-10
Implement Row & Col Traits for DMat

oleglite:
Issue #153


sebcrozet:
Thanks!
Things will move a little once I finish with the second point of #100 (dynamic matrices with bounded sizes) though.

PR #178, opened 2016-04-07
Removed unneeded parentheses

Binero:
There were unneeded parentheses in for loops. This caused it to fail compiling. I ran a quick Regex over the repo to remove them.


Binero:
Actually, it seems like I was just very behind on the master branch. I thought I made a new fork but apparently it dug up an old one. This can be closed.

Issue #177, opened 2016-04-07
Use associated types when dealing with Vec/Pnt traits


Binero:
The traits NumPnt, NumVec, FloatPnt and FloatVec should really be using associated types, as PntX only maps to VecX, and vice versa.

PR #179, opened 2016-04-07
Made PntAsVec use associated types.

Binero:
This PR aims to solve #177. 

Do note that PR #126 also moves nalgebra to associated types, however PntAsVec is not included there.


Binero:
Forgot to commit a fix. Compile error should be gone now.

sebcrozet:
Thank!

Issue #176, opened 2016-04-06
Poll: rename everything to use full names, e.g., `Iso3` becomes `Isometry3`.


sebcrozet:
Currently, `nalgebra` tends to use 3-letter or 4-letter names for all the algebraic entities (e.g. Vec, Mat, Rot, Quat, etc.) Unfortunately, this naming convention is not very (not _at all_ actually) explicit for some structures like (direct) isometries `Iso` and the recently added similarity transformations `Sim`.

Thus, I now think that full names would be much better as they would make the code more readable, and would help users to find what he wants on the documentation. For example, I doubt that anyone reading the documentation would guess at first glance that `Sim` is a transformation.

So I would like to know if current users would be in favor of full names instead of the current ones! Of course, if the majority seems to agree with this change, it would break every single project using `nalgebra`.

aepsil0n:
I am in favour of this, as it makes the code more readable. The breaking change is something that cargo can deal with, I think, as long as you bump the major version number. (Of course, not when people depend on `"*"`.)

bjz:
I don't currently use nalgebra, but I would support this – to be honest, the abbreviations in the methods and types are part of the reason I continued development on cgmath. There is also the issue that `Vec` is something special in the `std`.

Binero:
Yes.

sebcrozet:
Alright, I am going to rename everything for the version 0.7.0!

sebcrozet:
Actually, this will be for the version 0.8.0. I had to bump it to 0.7.0 already because of #179 .

arturoc:
too late i guess but i had just had a surprise when updating a project and found out that the names had changed. the abbreviations were one of the reasons i chose nalgebra over cgmath. i'm mostly using nalgebra for opengl stuff and it makes so much sense to have Vec3 or Mat4 since that's the same naming that glsl uses also glm the de-facto c++ library for graphics math. 

Rust itself also uses a lot of abbreviations, like Vec over Vector or Rc, Ref... so it would seem that abbreviations are somehow more idiomatic

also this classes are usually used a lot of times if you are creating vectors of vec so having to type the full thing is pretty tedious and doesn't really add much in my opinion. 

i even use some utility functions like vec2, vec3, pnt2... so i can quickly create objects without having to call the full class + new, so Vec2::new(x,y)  -> vec2(x,y)

Jonesey13:
@arturoc said pretty much all of my thoughts on this. 

[...]

PR #183, opened 2016-04-14
Support `f64 * Vec3<f64>` and so on. Fixes #182.

### LINKED ISSUE #182: Support multiplying f64 * Vec3<f64> and so on
Rust doesn't allow `impl<N> Mul<Vec3<N>> for N`, but na could provide an `impl` for the most common types, `f32` and `f64`, just as a convenience. In my case, it'd be nice; I'm using `f32` exclusively.

If this is something you'd take, let me know. I'll send a PR.

jorendorff:
To be specific, support is added for `N op T<N>` where `N` is `f32` or `f64`, `op` is one of `+` `-` `*`, and `T` is one of the `Vec`, `DVec`, `Mat`, or `DMat` generic types. These are all cases where `T<N> op N` is already supported.

*   Except for `f32 * Vec3<f32>`, which works fine :) the rest of this is untested.

*   The code I'm touching in `src/mat_macros.rs` seems to be dead code! I can remove it or just leave it alone instead, just let me know what to do...

### DIFF
--- a/src/structs/dmat_macros.rs
+++ b/src/structs/dmat_macros.rs
@@ -634,6 +634,36 @@ macro_rules! dmat_impl(
             }
         }
 
+        impl Mul<$dmat<f32>> for f32 {
+            type Output = $dmat<f32>;
+
+            #[inline]
+            fn mul(self, right: $dmat<f32>) -> $dmat<f32> {
+                let mut res = right;
+
+                for mij in res.mij.iter_mut() {
+                    *mij = self * *mij;
+                }
+
+                res
+            }
+        }
+
+        impl Mul<$dmat<f64>> for f64 {
+            type Output = $dmat<f64>;
+
+            #[inline]
+            fn mul(self, right: $dmat<f64>) -> $dmat<f64> {
+                let mut res = right;
+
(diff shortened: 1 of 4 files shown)


jorendorff:
Hmm. `test_cholesky_mat4` passes for me locally, and the patch only adds new code. What do you think is going on here?

jorendorff:
...and it's passing now. OK then!

jorendorff:
Oh - also, if you'd like me to factor out the common code among the Add, Sub, and Mul cases, I can write a macro for that.

sebcrozet:
Yes, I restarted the builds on Travis and it passed. My guess is that, because the input matrix is generated randomly, there are times when it is too badly conditioned for the Cholesky decomposition to succeed… I am not sure.

Refactoring would be great indeed. Also, the macro could be parametrized by the scalar type so that it could easily be called one time per primitive type (f32, f64, but also for i32, u64, etc.).

sebcrozet:
Thanks! I'm going to do the refactoring you suggested now, so that I can start changing everything for #176 .

jorendorff:
OK, thanks!

PR #197, opened 2016-08-25
implement common operations for references

Limeth:
Implements operations for references.

Examples:
- `&Vector3<F> * F`
- `Vector3<F> * &F`
- `&Vector3<F> * &F`
- `-&Vector3<F>`

### DIFF
--- a/src/structs/common_macros.rs
+++ b/src/structs/common_macros.rs
@@ -1,5 +1,49 @@
 #![macro_use]
 
+macro_rules! ref_binop(
+    // Symmetric to `std::ops::forward_ref_binop!`
+    (impl $imp:ident, $method:ident for $t:ident; $($compN:ident),+) => {
+        impl<'a, 'b, N> $imp<&'a $t<N>> for &'b $t<N> where &'b N: $imp<&'a N, Output = N> {
+            type Output = $t<N>;
+
+            #[inline]
+            fn $method(self, right: &'a $t<N>) -> $t<N> {
+                $t::new(
+                    $(
+                        $imp::$method(&self.$compN, &right.$compN)
+                    ),+
+                )
+            }
+        }
+
+        impl<'a, N> $imp<&'a $t<N>> for $t<N> where N: $imp<&'a N, Output = N> {
+            type Output = $t<N>;
+
+            #[inline]
+            fn $method(self, right: &'a $t<N>) -> $t<N> {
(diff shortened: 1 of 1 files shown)


src/structs/common_macros.rs
```diff
@@ -79,156 +129,160 @@ macro_rules! pointwise_sub(
...
-        }
-
         impl<N: Copy + AddAssign<N>> AddAssign<N> for $t<N> {
             #[inline]
             fn add_assign(&mut self, right: N) {
                 $( self.$compN += right; )+
             }
         }
 
-        impl Add<$t<f32>> for f32 {
```
sebcrozet: This, and all the other implementations of Add/Mul/Div in this file that involve f32 or f64 explicitly must not be removed. They allow the user to write things like: `10.0f32 * v`. The other, generic, implementations allow the scalar to be only on the right-hand-side.

f32 and f64 and mentioned explicitly because it is not possible to write something generic like `impl<N> Mul<Vector3<N>> for N { ... }`

Limeth: Oh, I was wondering what it was for. Thanks for clarifying, I will put it back.

src/structs/common_macros.rs
```diff
@@ -79,156 +129,220 @@ macro_rules! pointwise_sub(
...
 
 
-macro_rules! pointwise_scalar_div(
-    ($t: ident, $($compN: ident),+) => (
-        impl<N: Copy + Div<N, Output = N>> Div<N> for $t<N> {
+macro_rules! ref_binop_scalar(
+    // Symmetric to `std::ops::forward_ref_binop!`
+    (impl $imp:ident, $method:ident for $t:ident; $($compN:ident),+) => {
+        ref_binop_scalar_exact!(impl $imp<f32>, $method for $t; $($compN),+);
+        ref_binop_scalar_exact!(impl $imp<f64>, $method for $t; $($compN),+);
```
Limeth: @sebcrozet I re-implemented the `f32 * T<f32>` ops, now it should support `&f32 * T<f32>`, `f32 * &T<f32>` and `&f32 * &T<f32>` aswell.

Limeth:
Fixes https://github.com/sebcrozet/nalgebra/issues/145.

sebcrozet:
Pretty cool. Thanks!

PR #209, opened 2016-10-23
Fix benchmark macro `bench_binop`

jmagnuson:
Previously, `cargo bench` had been failing out with, (e.g.)

```error[E0277]: the trait bound `&na::Quaternion<f32>: std::ops::Mul<na::Quaternion<f32>>` is not satisfied```


sebcrozet:
Thanks!

Issue #207, opened 2016-09-20
Trait reform: big plans for nalgebra 0.11.0


sebcrozet:
---

This discussion is also on the [user forum](http://users.nphysics.org/t/trait-reform-big-plans-for-nalgebra-0-11-0/37).
**Please**, share what you would like to see added to the next release of nalgebra (in particular for computer-graphics related features)!

---

After #187, the next release will be the next big step toward what will be a version 1.0 in the future. It will include many breaking changes (especially for generic programming) but this is the last time it will break that much!

This trait reform is also a good opportunity to reorganize the internals of nalgebra, get rid of some hidden aberrations (like the ability to change the `Rotation2` diagonal to arbitrary values), and take in account negative feedbacks (I actually read IRC logs) regarding its ergonomy for computer graphics applications.

The main goals are to:
- allow generic programming through solid trait-based mathematical foundations. The current traits are too granular and not organized (with inheritance) in a coherent manner.
- use the [approx](http://crates.io/crates/approx) for approximate equality testing (i.e. don't redefine our own).
- add computer-graphics-related features to 3x3 and 4x4 raw matrices. The computer graphics community is used to work with 4x4 matrix to represent a general 3D transformation (or 3x3 matrices for 2D). Thus, people are usually frustrated when they can't do anything CG-related with `Matrix3` and `Matrix4`. This will change but they will still not be usable in a generic context (except for one case, see bellow).
# Generic programming

The new trait hierarchy will:
- mostly be implemented in a separate crate: [alga](https://github.com/sebcrozet/alga). Note that this is based on the excellent work of @brendanzab and @WaDelma on [algebra](https://github.com/brendanzab/algebra) (I wanted to give a fresh reboot of this library with different design choices).
- follow more strict mathematical notions and names. If several names are possible for the same structure, the simpler one is chosen (e.g. trait for rotations will be called `Rotation` instead of `SpecialOrthogonalGroup`).
- provide the same (or more) features that already exist on `nalgebra`. You'll just have to change the name of the traits you import, and sometimes the methods names as well. For example the `SquareMatrix`, `Diagonal`, `Column`, and `Row` will be merged into the traits `SquareMatrix` and `SquareMatrixMut` (the latter allows modifications).

Implementing your own instance of the traits of `alga` will be a bit harder as you'll have to
[...]

PR #212, opened 2016-11-28
Made BaseNum to be a blanket implementation.

WaDelma:
I wanted to use nalgebra vectors and matrices with `Complex` from num crate, but because `BaseNum` isn't implemented for it, things traits like `Dot` don't apply vectors containing them.

This is breaking change: If crate using nalgebra has implemented the trait themselves this will cause them to break. So I don't expect this to be merged.


sebcrozet:
Thanks! Instead of a blancket impl. I'd prefer to explicitly implement `BaseNum` for `Complex` instead. `BaseNum` will disappear in #207 anyway. I'm closing this but will keep in mind that, in the new design, we have to be careful to allow complex number whenever possible.

Issue #160, opened 2015-11-12
Into & From implementations


Binero:
It'd be nice if there were Into and From implementation between arrays, matrices and other objects. This'd be more intuitive than having to go to the online doc and search for the exact function name.

sebcrozet:
I've made 58000e48383e0285da5863321195b4652a7f8c23 to address this issue by implementing `AsRef`, `AsMutRef` and `From` for `Mat*`, `Vec*`, and `Quat`. Though it does not implement the `Into` trait. The implementations are made for pointers because one may switch between the array and Vec/Mat/Quat representations for free (i-e without any copy). To obtain an actual copy, one can still use `.clone()` or the `*` operator on the reference returned.

I'm still unsure whether adding by-copy implementations would be useful. Unfortunately, we cannot perform partial moves on arrays, so an e.g. `impl<N> From<[N; 3]> for Vec3<N>` would require `N` to implement `Clone`, which is unfortunate (but, not very restrictive as N will almost always be a scalary, Copy-able type).

moll:
@sebcrozet: Might be a silly beginner question, but how you use the `from` functions with arrays?

I'm giving them an inline array reference (`Matrix4::from(&[...])`), yet am getting the following type error:
```
 expected `nalgebra::structs::matrix::Matrix4<_>`,
    found `&[[f32; 4]; 4]`
(expected struct `nalgebra::structs::matrix::Matrix4`,
    found &-ptr) [E0308]
```

This is with Rust v1.8.0 and Nalgebra 0.8.0. Thanks!

sebcrozet:
Because `From` is implemented for `&Matrix4` (to perform a shallow conversion) instead of `Matrix4`, you have to do something like this:

```rust
let mat: &Matrix4<f64> = From::from(&[ ... ]);
```

I realize that's not very ergonomic though… so it would be useful to add an implementation of `From<[[N; 4]; 4> for Matrix4<N>` as well in a future release of nalgebra.

sebcrozet:
@moll You will now be able to do what you intended (`Matrix4::from(&[...])`) with the version 0.8.1 of nalgebra!

moll:
Thanks!

mathstuf:
I'm interested in an instance of `From<Matrix4> for [[f32; 4]; 4]` (basically the opposite of the existing ones). Would that be acceptable?

sebcrozet:
[...]

Issue #36, opened 2014-11-21
Conversion to and from static arrays


aepsil0n:
This is motivated mainly to allow a seamless interaction with piston libraries. Piston internally uses https://github.com/PistonDevelopers/vecmath, which are generic functions on static arrays ([T, ..n] or [[T, ..n], ..m]). Thus to interface with the libraries it is crucial to be able to convert vectors and matrices to and from static arrays seamlessly.

I don't know whether this warrants changing the internal representation to arrays. Transitioning from cgmath I find this a bit cumbersome to work around.

tomjakubowski:
Maybe I'm missing something here, but isn't implementing `From<[T; N]>` and `Into<[T; N]>` for e.g. `VecN<T>` sufficient?

sebcrozet:
You're correct, implementing those two traits is the way to go! They did not exist (and neither did DSTs) when this issue was created.

sebcrozet:
Into and From are now implemented for matrices and vectors since v0.11.1. For other structures, the user can access their vector or matrix representations (and then use Into and From) if needed. Closing this.

PR #225, opened 2017-02-18
Implement AsRef, AsMut, Into, From for matrices.

### LINKED ISSUE #223: 0.11 removes as_ref for vector and matrices?
Seems like AsRef<[T;N]> is not implemented anymore for VectorN<T> which makes it harder to write generic code that will accept any vector type without an explicit dependency on nalgebra

sebcrozet:
They were mistakenly removed or incomplete.

Fix #223 #160 #36

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name    = "nalgebra"
-version = "0.11.0"
+version = "0.11.1"
 authors = [ "Sébastien Crozet <developer@crozet.re>" ]
 
 description = "Linear algebra library for computer physics, computer graphics and general low-dimensional linear algebra for Rust."
--- a/src/core/conversion.rs
+++ b/src/core/conversion.rs
@@ -1,7 +1,11 @@
+use std::ptr;
+use std::mem;
+use std::convert::{From, Into, AsRef, AsMut};
+use typenum::{self, Cmp, Equal};
 use alga::general::{SubsetOf, SupersetOf};
 
 use core::{Scalar, Matrix};
-use core::dimension::Dim;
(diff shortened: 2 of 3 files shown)


arturoc:
Since now Matrix4 (for example) implements both `AsRef<[[f32;4];4]>` and `AsRef<[f32;16]>` an API that accepts both as possible parameters will fail to disambiguate Matrix4 and requiring some casting.

For example:

```
trait AsUniform<T>{
    fn as_uniform(&self) -> Uniform;
}

impl<T: AsRef<[[f32;4];4]> AsUniform<AsRef<[[f32;4];4]> for T{
    fn as_uniform(&self) -> Uniform{
         ...
    }
}

impl<T: AsRef<[f32;16]> AsUniform<AsRef<[f32;16]> for T{
    fn as_uniform(&self) -> Uniform{
         ...
    }
}

fn uniform<T:AsUniform<U>, U>(uniform: &T){
    ...
}

...

let m: Matrix4<f32> = one();
uniform(&m);
```

used to work but will fail now requiring the call to be something like:

```
uniform(&m as &AsRef<[[f32;4];4]>);
```

which is much less convenient. 

i can't see any advantage of implementing both AsRef for 4x4 and 16 so perhaps is better to let APIs accepting AsRef to be the ones who implement both cases so they work with libraries that choose to use one or another?

arturoc:
btw, something similar happens with vectors which now implement AsRef for [f32;N] but also [[f32;N];1] so you also need to disambiguate them which in some very general macros or similar can be quite hard

sebcrozet:
Oh, I agree that can be a problem! I think I will just re-implement the old behavior, i.e., `[f32; N]` for vectors and `[[f32; NRows]; NCols]` for matrices.

PR #230, opened 2017-03-02
De-ambiguate AsRef/AsMut/Into implementations.

sebcrozet:
Addresses comments of #225.


aepsil0n:
cc @arturoc

arturoc:
I don't remember the implementation details but i think this should be ready to merge. Perhaps @sebcrozet has any feedback on why he didn't merged it before just in case

sebcrozet:
I didn't merge it right away because this is a breaking change. But I can merge it now and release it later (I plan to release a new "major 0.x" version mid-august).

PR #237, opened 2017-04-06
Allow for documentation strings in macro component_binop_impl.

JonasOlson:
The documentation currently says "Componentwise matrix multiplication.", no matter what the operation actually does. The attached commit fixes this by introducing more parameters to the macro, through which documentation strings can be passed. I'm not at all sure that this is the appropriate solution, but it's all I could come up with.


sebcrozet:
I think that's the right approach. Thanks!

PR #239, opened 2017-04-09
Fixing the documentation link in Cargo.toml

projektir:
This link doesn't seem to be correct and this is leading to the link on crates.io pointing to the wrong place, I believe.


sebcrozet:
Thanks for noticing! It's always hard to keep track of all the dead links…

PR #247, opened 2017-04-27
Relax invertibility check for Matrix::try_inverse

### LINKED ISSUE #244: Suggestion: Don't use approximate determinant check to determine invertibility
The current [try_inverse()](https://github.com/sebcrozet/nalgebra/blob/1471031ac45e8b69558259a44726087a3383e511/src/core/inverse.rs#L57) implementation for 2x2 and 3x3 matrices attempts to use an approximate comparison with the determinant as a check for whether the matrix is invertible. Unfortunately, this is not a reliable way to determine if a matrix is invertible, and will readily refuse to invert perfectly invertible matrices.

Consider the following 3x3 matrix:

```
[ a   0   0 ]
[ 0   a   0 ]
[ 0   0   a ]
```

where `a` is some arbitrary non-zero constant. This matrix is as invertible as it gets: it is diagonal and has a condition number of 1, and there are no numerical issues associated with either inverting it or solving for a given right-hand side. However, the determinant of this matrix is a<sup>3</sup>. If, say, `a = 1e-6`, this means that its determinant is `1e-18`, which will fail the relative test currently used.

What is the correct way to determine if the matrix is invertible? Well, in this case, your best bet is probably to just check if the determinant is exactly zero. If it is not, assume it is invertible at the cost of some possible numerical instability. Anything else would require some rigorous numerical analysis and possible non-negligible runtime overhead for additional checks.

I'm running into problems for my simple game because I have some objects with significantly differing masses, which leads to inertia tensors of similar form as the example I posted.

I'd be willing to make a PR for the change, but it of course presumes that we can agree on the remedy.

Andlon:
Resolves #244. I couldn't find any existing tests for `try_inverse` at all, so I took the liberty to add some very basic tests for a single example of a few different dimensions. Let me know if you don't want them!

### DIFF
--- a/src/core/inverse.rs
+++ b/src/core/inverse.rs
@@ -45,18 +45,19 @@ impl<N, D: Dim, S> SquareMatrix<N, D, S>
             match dim {
                 0 => true,
                 1 => {
-                    if relative_eq!(self.get_unchecked(0, 0), &N::zero()) {
+                    let determinant = self.get_unchecked(0, 0).clone();
+                    if determinant == N::zero() {
                         false
                     }
                     else {
-                        *self.get_unchecked_mut(0, 0) = N::one() / self.determinant();
+                        *self.get_unchecked_mut(0, 0) = N::one() / determinant;
                         true
                     }
                 },
                 2 => {
                     let determinant = self.determinant();
 
-                    if relative_eq!(&determinant, &N::zero()) {
+                    if determinant == N::zero() {
                         false
                     }
                     else {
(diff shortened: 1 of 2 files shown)


sebcrozet:
Thanks!

There are already tests on `tests/matrix.rs` that check that multiplying a random matrix by its inverse computed with `.try_inverse()` result in the identity matrix. But your tests are great as well so we can keep them!

Andlon:
Oh! I have no idea why I didn't find them, I thought I searched the `tests` folder for the keyword `inverse`. Huh, well, whatever! Happy to see it merged :-)

PR #251, opened 2017-05-04
Remove gratuitous indirection in serialization impls

### LINKED ISSUE #249: Use forwarding impls of Serialize/Deserialize for wrapper types
A unit quaternion currently encodes in JSON as something like `{"value":{"coords":{"data":[0.0,0.0,1.0,0.0]}}}`, and in TOML as something like `[[[[0, 0, 0, 1]]]]`. Even a `Vector3` ends up as `{"data":[0.0,0.0,0.0,1.0]}`. This doesn't seem useful, and is in particular very unfriendly for data intended to be written by hand. If Serialize/Deserialize implementations were written to forward to the ultimate storage, rather than automatically derived, the resulting encodings would be much friendlier. If nobody has any objections, I'm happy to do the grunt work of making it happen.

Ralith:
I think I got everything; let me know if not.

I deliberately passed on intelligent handling of nrows/ncols in `MatrixVec` for now because it seems considerably less trivial.

In the future, these impls could be replaced by a macro, or better yet, attribute-based support in serde for forwarding implementations.

Fixes #249

### DIFF
--- a/src/core/dimension.rs
+++ b/src/core/dimension.rs
@@ -7,9 +7,11 @@ use std::any::Any;
 use std::ops::{Add, Sub, Mul, Div};
 use typenum::{self, Unsigned, UInt, B1, Bit, UTerm, Sum, Prod, Diff, Quot};
 
+#[cfg(feature = "serde-serialize")]
+use serde::{Serialize, Serializer, Deserialize, Deserializer};
+
 /// Dim of dynamically-sized algebraic entities.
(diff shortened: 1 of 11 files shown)


sebcrozet:
Thanks! Everything seems OK. 
Though since I believe we could see this as a breaking change, I'll wait a bit before merging.

aepsil0n:
@sebcrozet I am going to merge this, too, as we are rolling up breaking changes for 0.13 anyway.

PR #260, opened 2017-05-20
Update dependencies

burtonageo:


aepsil0n:
I have doubts, whether the build failure is actually related to the dependency update. Can you reproduce this to see what's causing it?

@sebcrozet I think bumping dependency versions should cause our major version to be bumped to 0.13, right? btw, just to be clear, in `0.x.y`, do we call `x` major and `y` minor? pre 1.0 this whole semver business seems a little uncertain to me.

sebcrozet:
No, the failure is not caused by the dependencies update. I think this is because quickcheck generated a cornecase example that caused the test to fail. I restarted the build and it dies not fail any more but we'll need to investigate why it failed.

In the mean time, merging this should be ok.

Yes, pre 1.0, we call `x` major and `y` minor in `0.x.y`. We need to bump the major version of nalgebra if we change the major version of its dependencies indeed.

sebcrozet:
I'll merge this now and make a major release mid-august.

PR #265, opened 2017-06-09
Enable serde on num-complex

corngood:
I needed this to serialise `UnitComplex`.  Does it seem reasonable?


aepsil0n:
Seems like a natural fit to me, merging.

PR #270, opened 2017-07-25
Mint integration

kvark:
see https://github.com/kvark/mint

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,10 @@ script:
     - cargo --version
     - cargo build --verbose
     - cargo build --verbose --features arbitrary
+    - cargo build --verbose --features mint
     - cargo build --verbose --features serde-serialize
     - cargo build --verbose --features abomonation-serialize
-    - cargo test --verbose --features "debug arbitrary serde-serialize abomonation-serialize"
+    - cargo test --verbose --features "debug arbitrary mint serde-serialize abomonation-serialize"
     - cd nalgebra-lapack; cargo test --verbose
 
 env:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,8 @@ alga           = "0.5"
 matrixmultiply = "0.1"
 serde          = { version = "1.0", optional = true }
 serde_derive   = { version = "1.0", optional = true }
-abomonation     = { version = "0.4", optional = true }
+abomonation    = { version = "0.4", optional = true }
+mint           = { version = "0.4.2", optional = true }
 
 [dependencies.quickcheck]
 optional = true
--- a/src/core/conversion.rs
+++ b/src/core/conversion.rs
@@ -280,8 +280,61 @@ macro_rules! impl_from_into_mint_1D(
 
 // Implement for vectors of dimension 2 .. 4.
 #[cfg(feature = "mint")]
-impl_from_into_mint_1D!(    // Column vectors.
+impl_from_into_mint_1D!(
     U2 => Vector2[2];
     U3 => Vector3[3];
     U4 => Vector4[4];
+);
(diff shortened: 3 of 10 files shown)


sebcrozet:
@aepsil0n Not yet. I'm not yet very convinced behind the actual usefulness of mint either. Also, I'd like to investigate if merging this would introduce ambiguities when calling `.as_ref` (see comments from @arturoc on #225).

kvark:
We can remove as_ref for mint types if that's a concern

sebcrozet:
@kvark Actually, my concern applies to `Into/From` as well. If nalgebra already has an implementation of those traits, introducing second implementations for mint types causes ambiguities if the context does not allow the compiler to infer easily the return type, e.g.: `println!("{}", v.into())`. So this code will compile with  mint disabled (because only one implementation of `Into` is found), but won't once mint is enabled. Thus just enabling mint may cause some breakage because of the ambiguities introduced by the appearance of those implementations.

kvark:
@sebcrozet I see. Do you have a policy (in this project) to only convert `Into`/`From` single destination? AFAIK, it's completely normal for libstd to have multiple different conversions for types.

Ralith:
I think it's overwhelmingly more common for conversion traits to be used in cases where their context makes them unambiguous (e.g. passing to a differently typed interface). If you want a specific, monomorphic conversion, perhaps a dedicated method for same would be a better way to provide it?

torkleyy:
> So this code will compile with mint disabled (because only one implementation of `Into` is found), but won't once mint is enabled.

I don't see how that is true. If that was the case, the compiler had to prove that no dependent library of `nalgebra` implements `Into<??>` for a nalgebra type. Try the code out yourself, it will fail to compile, even if your types are private; a type annotation is **always** required in

```rust
println!("{:?}", my_value.into());
```

The compiler can never infer something based on some trait only being implemented with one exact type. If this was the case, why do we have orphan rules? Otherwise a new dependency could always break your code.

arturoc:
The problem with having more than one conversion only for this kind of traits is not of directly calling the method but when another api accepts those 2 conversions too. For an example see my comment in https://github.com/sebcrozet/nalgebra/pull/225

I don't think Into/From would have the same problem since usually they are used as:

```
fn some_function<Into<Something>>(..)
```

which would accept anything that converts into the destination so that shouldn't have the same problem as the AsRef one.

That said i wonder if mint itself could have features for interoperability with nalgebra, cgmath... instead of the other way around given that it seems a much newer project. Also i've found that for basic cases like vectors, matrices... usually implementing AsRef for the underlying memory structure is enough to be compatible with several libraries.

aepsil0n:
@arturoc fyi I raised roughly the same concerns over in brendanzab/cgmath#418, may be relevant to read the response by @kvark.

kvark:
@sebcrozet what's the consensus w.r.t this PR? I need to know if it's worth rebasing, or you don't care.

sebcrozet:
@kvark Sorry for the very late answer. I've been busy and forgot about this PR. Given the discussion on this thread, and the fact that it seems common for the stdlib to provide several implementations of conversion traits, I'm OK to merge this if its rebased (and mint remains optional)! I understand this can be convenient in some cases.

kvark:
@sebcrozet thanks for the answer. The code is rebased now.
I've got to say, the internal interfaces look nicer now, great to see `nalgebra` being simplified inside!

sebcrozet:
Thanks ! Hopefully it will become more and more simple as new language features get stabilized (especially const generics and specialization obviously).

torkleyy:
:tada: :heart: Wow, nice to see this merged ;)

PR #272, opened 2017-07-28
Fix TCategory implementation for TAffine

### LINKED ISSUE #271: panic 'Matrix index out of bounds.' in check_homogeneous_invariants (impl TCategory for TAffine)
The following code:
```rust
let local_transform : Affine3<f32> = try_convert(Matrix4::<f32>::identity()).unwrap();
```
results in a panic 'Matrix index out of bounds.' in check_homogeneous_invariants: [src/geometry/transform.rs, line 104](https://github.com/sebcrozet/nalgebra/blob/7a62b68c38ca4689a81694b5da505abf8872a4b2/src/geometry/transform.rs#L104-L104)

```rust
fn check_homogeneous_invariants<N, D, S>(mat: &SquareMatrix<N, D, S>) -> bool
        where N: Scalar + Field + ApproxEq,
              D: DimName,
              S: Storage<N, D, D>,
              N::Epsilon: Copy {
        mat.is_invertible()                   &&
        mat[(D::dim(), D::dim())] == N::one() &&
        (0 .. D::dim()).all(|i| mat[(D::dim(), i)].is_zero())
    }
```

Shouldn't `D::dim()` be `D::dim()-1` ?

aepsil0n:
Closes #271.

### DIFF
--- a/src/geometry/transform.rs
+++ b/src/geometry/transform.rs
@@ -100,9 +100,10 @@ impl TCategory for TAffine {
               D: DimName,
               S: Storage<N, D, D>,
               N::Epsilon: Copy {
-        mat.is_invertible()                   &&
-        mat[(D::dim(), D::dim())] == N::one() &&
-        (0 .. D::dim()).all(|i| mat[(D::dim(), i)].is_zero())
+        let last = D::dim() - 1;
+        mat.is_invertible()           &&
+        mat[(last, last)] == N::one() &&
+        (0 .. last).all(|i| mat[(last, i)].is_zero())
     }
 }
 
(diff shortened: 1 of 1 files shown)


aepsil0n:
cc @sebcrozet 

While the fix is obvious, I was wondering a little about the testing strategy here. This part of the code has not been tested yet. There is a lot of generative testing going on out-of-crate, so is this the general strategy, how stuff like this should be tested?

sebcrozet:
Thanks for the fix! Yes, I tend to prefer out-of-crate testing so that they are all grouped in a single folder. I usually test things by generating random inputs with quickcheck. On my ongoing work, I added ways to make quickcheck generate, e.g., only invertible matrices for testing. Let's merge this for now, and add more tests in a few weeks.

sebcrozet:
The issue to remember to add test: #273

PR #310, opened 2018-01-17
Row vector cross product

### LINKED ISSUE #308: Vector and RowVector cross product
use nalgebra::Vector3;
    let p1 = Vector3::new(1.0, 2.0, 3.0);
    let p2 = Vector3::new(1.0, 5.0, 7.0);
    print!("{}", p1.cross(&p2));
    -1
    -4
     3

    use nalgebra::RowVector3;
    let p1 = RowVector3::new(1.0, 2.0, 3.0);
    let p2 = RowVector3::new(1.0, 5.0, 7.0);
    print!("{}", p1.cross(&p2));
    [ -1 -4 -4 ]

I don't understand why the results are different! The `Vector3` is the right answer. What does `RowVector3` is calculating? As a standard user (and non-math expert), I'm expecting the results to be equal. Is there a reason not to?

nilgoyette:
Fix #308 

This is not related to this issue, but I tried fixing the `soooo ugly!` problem. We would need a tool to get the first, second and third element, like `.x`, `.y` `.z` They are not available in this function. Do you now where they are defined? I might try the `FIXME: use specialization instead of an assertion.` when I have some free time.

### DIFF
--- a/src/core/matrix.rs
+++ b/src/core/matrix.rs
@@ -879,7 +879,7 @@ impl<N: Scalar + Ring, R: Dim, C: Dim, S: Storage<N, R, C>> Matrix<N, R, C, S> {
 
                 *res.get_unchecked_mut(0, 0) = ay * bz - az * by;
                 *res.get_unchecked_mut(0, 1) = az * bx - ax * bz;
-                *res.get_unchecked_mut(0, 2) = az * bx - ax * bz;
+                *res.get_unchecked_mut(0, 2) = ax * by - ay * bx;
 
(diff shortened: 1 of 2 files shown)


aepsil0n:
Thank you for the contribution. :)

The "soooo ugly" issue may also be rooted in the fact that we do not statically know in this context that we are dealing with a 3D vector. There is only a dynamic check to make sure. If that was the case, it would probably be a lot easier to create specialized methods hiding away all those unchecked calls.

Issue #295, opened 2017-10-23
Vector Debug should be more readable


Boscop:
Vector's Debug currently looks like this:
`Matrix { data: [1, 0], _phantoms: PhantomData }`

It should be more readable, without PhantomData, because this causes a lot of visual noise when printing arrays of Vectors.

aepsil0n:
I have found that a little annoying, too.

While the standard library provides a derive feature for `Debug`. There does not seem to be any convention, that it should be used though. The only thing I found [in the API guidelines](https://github.com/rust-lang-nursery/api-guidelines/blob/master/src/debuggability.md), said that it should be implemented and non-empty.

From my point of view, there are two main options that make sense:

1. Remove only `_phantoms`: `Matrix { data: [1, 0] }`
2. More radically, reduce it to the array itself: `[1, 0]`

I added the second suggestion, because it would be much more concise for printing small lists of small vectors. To justify: to me `Matrix` seems mostly like a semantic wrapper around the data contained. At the same time its completely transparent what's inside. The downside would be, that you could mix it up with an actual array. I am not sure that is worth avoiding for debug output, though.

cc @sebcrozet

Ralith:
I'd like to see 2; the long multi-line format provided by Display is rarely useful to me, and this would provide a convenient alternative. I expect it would also be easier to implement, as you'd only need to forward to the contained data's existing Debug impl.

PR #311, opened 2018-01-18
Remove phantom data from matrix debug output

aepsil0n:
Addresses #295.

### DIFF
--- a/src/core/matrix.rs
+++ b/src/core/matrix.rs
@@ -68,13 +68,21 @@ pub type MatrixCross<N, R1, C1, R2, C2> =
 /// dynamically-sized column vector should be represented as a `Matrix<N, Dynamic, U1, S>` (given
 /// some concrete types for `N` and a compatible data storage type `S`).
 #[repr(C)]
-#[derive(Hash, Debug, Clone, Copy)]
+#[derive(Hash, Clone, Copy)]
 pub struct Matrix<N: Scalar, R: Dim, C: Dim, S> {
     /// The data storage that contains all the matrix components and informations about its number
     /// of rows and column (if needed).
     pub data: S,
 
     _phantoms: PhantomData<(N, R, C)>
+}
+
+impl<N: Scalar, R: Dim, C: Dim, S: fmt::Debug> fmt::Debug for Matrix<N, R, C, S> {
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        formatter.debug_struct("Matrix")
+            .field("data", &self.data)
+            .finish()
+    }
 }
 
 #[cfg(feature = "serde-serialize")]
--- a/tests/core/matrix.rs
+++ b/tests/core/matrix.rs
@@ -62,6 +62,13 @@ fn iter() {
     assert!(it.next().is_none());
 }
 
+#[test]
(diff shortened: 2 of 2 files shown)


aepsil0n:
cc @Ralith @Boscop

Does this address your issue appropriately? I also considered the flattened structure and column layout, i.e. `[1, 3, 2, 4]` vs. `[[1, 2], [3, 4]]`. However, this is probably a separate issue not be addressed in this context. And perhaps it is ok like this anyway, as it is closer to a debug view.

jswrenn:
I'd favor a `Debug` that displayed the nalgebra wrapping type. I appreciate that information when I'm using debugging output, and it seems to be a Rust convention; there's even a [`DebugStruct`](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html) builder API for debugging output that requires the struct name. I'd recommend an implementation like this:
```rust
impl<N: Scalar, R: Dim, C: Dim, S: fmt::Debug> fmt::Debug for Matrix<N, R, C, S> {
    fn fmt(&self, fmt: &mut fmt::Formatter) -> Result<(), fmt::Error> {
        fmt.debug_struct(&format!("Matrix<{},{}>", self.rows(), self.cols()))
         .field("data", &self.data)
         .finish()
    }
}
```

or this, if we really want to elide that there's a field called `data`:
```rust
impl<N: Scalar, R: Dim, C: Dim, S: fmt::Debug> fmt::Debug for Matrix<N, R, C, S> {
    fn fmt(&self, fmt: &mut fmt::Formatter) -> Result<(), fmt::Error> {
        fmt.debug_tuple(&format!("Matrix<{},{}>", self.rows(), self.cols()))
         .field(&self.data)
         .finish()
    }
}
```

Ralith:
Although I like this as-is, the point about Rust convention is legitimate, so I won't be disappointed if it goes that way instead. It looks like if I really want maximally concise output, I can just format the `data` member directly, since it's public. That's convenient enough for me.

jswrenn:
Does `println!("{:?}", foo.as_slice())` fit your use-case, @Ralith? It reliably prints the data values on a single line, regardless of the backing storage type.

Ralith:
Yep, that would work too.

aepsil0n:
@jswrenn that is a valid concern. I get why it makes sense to clearly distinguish between an plain array and an nalgebra matrix.

What is your reasoning for including the number of rows in columns in your example? After all, our motivation was to reduce noise and that information is not present at the moment and nobody seems to be missing it.

The most conservative change would probably be to simple remove the phantom data part similar to your first suggestion: `Matrix { data: [1, 2, 3, 4] }`. I still have a slight preference for `Matrix([1, 2, 3, 4])`, but given the option so use `as_slice()` instead, I don't care too much either.

jswrenn:
> What is your reasoning for including the number of rows in columns in your example? After all, our motivation was to reduce noise and that information is not present at the moment and nobody seems to be missing it.

Yeah, on second thought, it's a poor addition.

I think the `Debug` output _should_ broadcast that there's a `data` field, since part of the role of the debug output is discoverability--`data` _is_ a public field, and it's one you access in the normal course of using `nalgebra`.

aepsil0n:
Cool, seems like we agree. I'll change the PR accordingly.

PR #333, opened 2018-04-27
Incoherent fundamental impls

### LINKED ISSUE #332: Fix warnings due to the incoherent_fundamental_impls lint.
This lint is going to become a hard error with https://github.com/rust-lang/rust/pull/49799. Fixing it will require a breaking change that may not affect too much users: changing the name of the matrix slice constructor from `new` to something else. Those were introduced by the version 0.14.0 of nalgebra.

sebcrozet:
Fix #332

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [0.15.0] - WIP
 ### Modified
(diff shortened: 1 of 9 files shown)


sebcrozet:
@Ralith Thanks for the suggestion. Though since https://github.com/rust-lang/rust/pull/49799 has not landed yet, I won't introduce a breaking change just now. Instead I'llmark the problematic constructors as deprecated, and suggest using the new ones.

PR #336, opened 2018-05-03
Add UnitQuaternion::{new_eps, from_scaled_axis_eps}.

### LINKED ISSUE #335: Unit quaternion constrution from axis-angle: allow customizing the epsilon.
Currently, when an unit quaternion is built from an axis-angle using, e.g., `::new`, the identity is returned whenever the provided scaled axis is "too small". By default, "too small" means that its magnitude is smaller than the `default_epsilon()` provided by the `approx` crate. In some configurations this epsilon may be too large.

We should add another constructor, e.g., `::new_eps` that takes the epsilon as input.

sebcrozet:
Fix #335

### DIFF
--- a/src/geometry/quaternion_construction.rs
+++ b/src/geometry/quaternion_construction.rs
@@ -351,7 +351,7 @@ impl<N: Real> UnitQuaternion<N> {
 
     /// Creates a new unit quaternion rotation from a rotation axis scaled by the rotation angle.
     ///
-    /// If `axisangle` has a magnitude smaller than `N::default_epsilon()`, this returns the indentity rotation.
+    /// If `axisangle` has a magnitude smaller than `N::default_epsilon()`, this returns the identity rotation.
     #[inline]
     pub fn new<SB>(axisangle: Vector<N, U3, SB>) -> Self
     where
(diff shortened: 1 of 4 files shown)


src/geometry/quaternion_construction.rs
```diff
@@ -351,7 +351,7 @@ impl<N: Real> UnitQuaternion<N> {
 
     /// Creates a new unit quaternion rotation from a rotation axis scaled by the rotation angle.
     ///
-    /// If `axisangle` is zero, this returns the indentity rotation.
+    /// If `axisangle` has a magnitude smaller than `N::default_epsilon()`, this returns the indentity rotation.
```
nilgoyette: I hope you mean "identity"? Also copied many times below.

src/geometry/quaternion_construction.rs
```diff
@@ -364,7 +364,20 @@ impl<N: Real> UnitQuaternion<N> {
...
+        SB: Storage<N, U3>,
+    {
+        let two: N = ::convert(2.0f64);
+        let q = Quaternion::<N>::from_parts(N::zero(), axisangle / two).exp_eps(eps);
+        Self::new_unchecked(q)
+    }
+
+    /// Creates a new unit quaternion rotation from a rotation axis scaled by the rotation angle.
+    ///
+    /// If `axisangle` has a magnitude smalle than `N::default_epsilon()`, this returns the indentity rotation.
```
nilgoyette: smalleR

src/geometry/quaternion_construction.rs
```diff
@@ -373,6 +386,18 @@ impl<N: Real> UnitQuaternion<N> {
     {
         Self::new(axisangle)
     }
+
+    /// Creates a new unit quaternion rotation from a rotation axis scaled by the rotation angle.
+    ///
+    /// If `axisangle` has a mangnitude smaller than `eps`, this returns the indentity rotation.
```
nilgoyette: maNgnitude

Issue #334, opened 2018-05-01
Slow matrix-vector multiplication


robsmith11:
I was trying to speed up my neural network calculations (which involve many matrix-vector multiplications and found that simply writing the multiplication with a reference to the matrix rather than the matrix itself produced a 36% reduction in run-time.

Is there any reason why `m*v` should be so much slower than `&m*v`?  If not, would it be possible to make them both fast?

Example:
```
extern crate nalgebra;
use nalgebra::*;

fn main() {
    let mut s = 0f64;

    let m:MatrixMN<f64,U15,U20> = zero();
    let mut v:VectorN<f64,U20> = zero();
    let mut r:VectorN<f64,U15> = zero();

    for _ in 0..10000000 {
        r = &m*v;   // fast
        // r = m*v; // slow
        s += r.iter().sum::<f64>();
    }

    println!("{}", s);
}
```

jswrenn:
Judging by a diff between the outputs of [`cargo asm`](https://crates.io/crates/cargo-asm) on the fast and slow versions of your example, it looks like the primary difference between them is the slow one involves an extra `memcpy`.

The fact that rather significant `memcpy`s can sneak into code using `nalgebra` is an unfortunate side-effect of `Matrix` implementing `Copy`. This issue is sorta tracked by #282, which mostly focuses on the benefits of being able to support non-Copy internal types. I would really like to see `Copy` removed from nalgebra's container types because of subtle performance pitfalls like the one you're describing.

sebcrozet:
That's an interesting observation @jswrenn. I'm actually surprised that the optimizer does not remove the `memcpy`, especially since the corresponding [implementation](https://github.com/sebcrozet/nalgebra/blob/master/src/core/ops.rs#L567) of `Mul` does use references internally. So I expected the compiler not to generate any `memcpy` after inlining the `mul` method.

I agree removing `Copy` would avoid this kind of pitfalls. Though we still need it (for convenience) for small vectors and matrices. So a solution would be to implement `Copy` for all matrices/vectors up to a small dimension, e.g., 6 which seems to be the threshold where the extra `memcpy` impacts performances negatively on benchmarks.

jswrenn:
I also started a branch yesterday that completely removes `Copy`. I shared @robsmith11's intuition, but after reading the documentation, I'm having doubts:

[...]

PR #337, opened 2018-05-04
Implement Copy for matrices and vetors up to dimension 6 only.

sebcrozet:
See the discussion in #334.

I'm not going to merge this before a week because I which to let anyone strongly opposed to this change to provide a feedback.

### DIFF
--- a/src/core/matrix_array.rs
+++ b/src/core/matrix_array.rs
@@ -20,7 +20,7 @@ use typenum::Prod;
 use generic_array::{ArrayLength, GenericArray};
 
 use core::Scalar;
-use core::dimension::{DimName, U1};
+use core::dimension::{DimName, U1, U2, U3, U4, U5, U6};
 use core::storage::{ContiguousStorage, ContiguousStorageMut, Owned, Storage, StorageMut};
 use core::allocator::Allocator;
 use core::default_allocator::DefaultAllocator;
@@ -103,7 +103,6 @@ macro_rules! impl_copy {
     };
 }
 
-use core::dimension::{U2, U3, U4, U5, U6};
 impl_copy!(U1, U1);
 impl_copy!(U1, U2);
 impl_copy!(U1, U3);
--- a/src/core/matrix_array.rs
+++ b/src/core/matrix_array.rs
@@ -97,16 +97,54 @@ where
     }
 }
 
-impl<N, R, C> Copy for MatrixArray<N, R, C>
-where
-    N: Copy,
-    R: DimName,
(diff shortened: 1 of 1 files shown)


jswrenn:
As I described [here](https://github.com/sebcrozet/nalgebra/issues/334#issuecomment-386647661), I oppose this change on the basis of conforming to the documented scope of `Copy`.

As an alternative: We should keep only the `&Matrix` × `&Matrix` binop implementation, and remove the `Matrix` × `Matrix`, `Matrix` × `&Matrix` and `&Matrix` × `Matrix` implementations. These other implementations just call out to the `&Matrix` × `&Matrix` implementation and we _assumed_ that the copy would therefore be optimized away. [It is **not** optimized away](https://github.com/sebcrozet/nalgebra/issues/334#issuecomment-386668587), and thus these other implementations are silent performance traps.

Removing these extraneous binop implementations would not significantly degrade ergonomics. For one, the problem can be resolved by users by adding a single character (`&`). Second, the error message users would receive if we took this approach would be much more helpful than the error message they would receive it we removed `Copy`. When these binop implementations are removed, rustc produces a very helpful error message to users who inadvertently pass `Matrix` rather than `&Matrix` as an operand to a binop that recommends that they pass a reference instead. For most users, _this_ is the right thing to do, _not_ add a `.clone()`.

In short this approach:
 - better corresponds with how the Rust documentation says `Copy` should be used
 - presents more helpful error messages to users

sebcrozet:
Closing this as it was not a good idea.

Nufflee:
> Closing this beacause it was not a good idea.

@sebcrozet excuse my ignorance but do you mind expanding on why exactly this would be a bad idea? Assuming that the compiler bug is fixed or all binops are changed to take `&Matrix`, why not implement `Copy` for small matrices whose element type is also `Copy`? I feel like doing so would improve the ergonomics of `nalgebra-glm` (I only really have experience with it) as most functions wouldn't have to take `Vec{N}`s by borrow any more.

Nufflee:
@Andlon: ah, I see, thanks. It remains a mystery to me why functions such as [nalgebra_glm::vec2_to_vec3](https://docs.rs/nalgebra-glm/latest/nalgebra_glm/fn.vec2_to_vec3.html) take arguments by borrow then.

sebcrozet:
@Nufflee It’s mostly for being coherent with the the original `nalgebra` API. However, I agree that, as far as `nalgebra-glm` is conserned, we should keep it simpler and not take any references.

PR #353, opened 2018-06-08
Adds conversions to and from mint Point types.

### LINKED ISSUE #352: Add mint conversions for Point types
In https://github.com/sebcrozet/nalgebra/blob/master/src/base/conversion.rs there are conversions to/from `mint` types for `Vector` and `Matrix` types, but not for `Point` types.

Will happily make a PR for this if you want.

edit: Quaternion types would be nice too.

icefoxen:
Resolves #352

### DIFF
--- a/src/geometry/point_conversion.rs
+++ b/src/geometry/point_conversion.rs
@@ -6,6 +6,16 @@ use base::dimension::{DimName, DimNameAdd, DimNameSum, U1};
 use base::allocator::Allocator;
 
 use geometry::Point;
+#[cfg(feature = "mint")]
+use mint;
+#[cfg(feature = "mint")]
+use base::dimension::{U2, U3};
+#[cfg(feature = "mint")]
+use std::convert::{AsMut, AsRef, From, Into};
+#[cfg(feature = "mint")]
+use base::storage::{Storage, StorageMut};
+#[cfg(feature = "mint")]
+use std::mem;
 
 /*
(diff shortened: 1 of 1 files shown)


src/geometry/point_conversion.rs
```diff
@@ -67,3 +76,57 @@ where
...
 }
+
+
+
+#[cfg(feature = "mint")]
+macro_rules! impl_from_into_mint_1D(
+    ($($NRows: ident => $PT:ident, $VT:ident [$SZ: expr]);* $(;)*) => {$(
+        impl<N> From<mint::$PT<N>> for Point<N, $NRows>
+        where N: Scalar,
+              DefaultAllocator: Allocator<N, $NRows> {
```
sebcrozet: This `DefaultAllocator` bound can be removed. This is never necessary when the dimension of the vector/matrix is given as a concrete type.

sebcrozet:
Thank you for this PR @icefoxen! Please remove the `DefaultAllocator` bound. After that this will be ready for merge.

icefoxen:
Done, thank you!  Funny that `rustc` doesn't warn you about unused/unnecessary trait bounds.

sebcrozet:
Thanks! I believe there was [some discussion](https://github.com/rust-lang/rust/issues/8366) already regarding such a warning. This would, for example, be problematic for trait like marker traits that only bring additional semantic information while not needed for the compilation.

PR #355, opened 2018-06-16
Changed loops order in from_fn_generic

tforgione:
Hi!

First, thank you for your librairies, they are great!

After investigating the performances of [the code from a friend on the forum](https://discourse.nphysics.org/t/efficient-ways-of-computing-a-centered-gradient/204), I realized that the `from_fn` function was quite slow due to the fact that matrices are stored by columns, but the iteration is done by rows.

I also added a benchmark to see the performance improvements:

Original version:
```
test core::matrix::mat100_from_fn :     7,736 ns/iter (+/- 246)
test core::matrix::mat500_from_fn : 1,085,050 ns/iter (+/- 192,820)
```

This PR version:
```
test core::matrix::mat100_from_fn:   2,796 ns/iter (+/- 189)
test core::matrix::mat500_from_fn: 650,736 ns/iter (+/- 16,264)
```

### DIFF
--- a/benches/core/matrix.rs
+++ b/benches/core/matrix.rs
@@ -192,7 +192,7 @@ fn mat100_from_fn(bench: &mut Bencher) {
     bench.iter(|| {
         DMatrix::from_fn(100, 100, |a, b| {
             a + b
-        });
+        })
     })
 }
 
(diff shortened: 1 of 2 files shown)


sebcrozet:
Good catch. Thanks!

PR #359, opened 2018-07-04
Uniformly distributed random rotations, unit vectors

Ralith:
This replaces the presumably non-uniform `Standard` distribution implementations for `UnitQuaternion` and `Rotation3` with uniform sampling algorithms taken from Graphics Gems III. While I was at it I also dropped in an implementation for `Unit<VectorN>`.

The implementation bounds for the pre-existing distributions were modified to replace `Standard` with `OpenClosed01`, which is arguably a breaking change. Although this is approximately a no-op for f32/f64, it's unclear what `Standard` might mean for more exotic types, and hence I feel it's appropriate to make the requirements of the algorithms used more explicit.

An alternative approach would be to introduce one or more new unit structs implementing `Distribution` for these algorithms, preserving the pre-existing `Standard` impls for backwards compatibility. This would also have the benefit of providing a clearer point to hang documentation off of, and making code that relies on the uniformity guarantee clearer. On the other hand, a non-uniform `Standard` distribution for these types strikes me as a landmine: users will invariably either want uniformity or a specific non-uniformity best accomplished with a documented or custom distribution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,11 @@ documented here.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [0.16.0] - WIP
+## Modified
+  * Adjust `UnitQuaternion`s and `Rotation3`s generated from the `Standard` distribution to be uniformly distributed.
 ### Added
   * Add construction of a `Point` from an array by implementing the `From` trait.
+  * Add support for generating uniformly distributed random unit column vectors using the `Standard` distribution.
 
 ## [0.15.0]
 The most notable change of this release is the support for using part of the library without the rust standard
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [0.16.0] - WIP
 ## Modified
-  * Adjust `UnitQuaternion`s and `Rotation3`s generated from the `Standard` distribution to be uniformly distributed.
+  * Adjust `UnitQuaternion`s, `Rotation3`s, and `Rotation2`s generated from the `Standard` distribution to be uniformly
+    distributed.
 ### Added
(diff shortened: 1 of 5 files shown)


sebcrozet:
Thanks! I am OK with the breaking change since the current implementation is not specific enough to be useful anyway.

The use of `StandardNormal` breaks compilation with `no_std` though (none of `StandardNormal` and `Normal` seem exported if the stdlib is disabled. See this [source](https://github.com/rust-lang-nursery/rand/blob/master/src/distributions/mod.rs#L183-L184).) If there is no alternative ways to achieving the same result with distributions compatible with `no_std`, we should guard the `Unit<VectorN<...>>` impl (as well as the `use rand::distributions::StandardNormal`) with a `#[cfg(features = "std")]`.

Ralith:
There are (superior?) alternative ways for the 2D and 3D cases (which would be nice to implement once specialization is a thing, perhaps), but I'm not aware of a good solution for the n-dimensional case. I'll guard it.

Ralith:
It'd be nice to have the uniformity stated somewhere in the documentation, but other than for `Rotation3` I'm not sure where would make sense.

sebcrozet:
I believe the documentation should go on the trait impls (above the `fn sample` methods). If I'm not mistaken this will override the default documentation provided by the trait on the generated doc.

The `use rand::distribution::StandardNormal` had to be guarded as well. I discovered I could edit the file myself directly on github. So I tried it to fix this (hence the additional commit). Hope you didn't mind!

Ralith:
No problem! Confirmed the docs work as you said; pretty hidden away, but at least the info's there. I also tweaked `Rotation2` to be uniform and adjusted the interface of the already-uniform `UnitComplex` impl for consistency.

sebcrozet:
Thank! Looks like one of the travis builds still fails, but it seems unrelated to your contribution. I'll merge this and investigate this separately.

PR #361, opened 2018-07-09
Add imax/imin()

bemeurer:
Currently the equivalents to numpy's `argmax()` and `argmin()` only exist for the min/max _absolute_ values in a Matrix. This implements non-abs variations of nalgebra's `aimax()` and `aimin()`.

### DIFF
--- a/src/base/blas.rs
+++ b/src/base/blas.rs
@@ -14,6 +14,27 @@ use base::storage::{Storage, StorageMut};
 use base::{DefaultAllocator, Matrix, Scalar, SquareMatrix, Vector};
 
 impl<N: Scalar + PartialOrd + Signed, D: Dim, S: Storage<N, D>> Vector<N, D, S> {
+
+    /// Computes the index of the vector component with the largest value.
+    #[inline]
+    pub fn imax(&self) -> usize {
+        assert!(!self.is_empty(), "The input vector must not be empty.");
+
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thanks! `imin` and `imax` are good to have.

Points addition on the other hand is not desirable because of the semantic we currently give to points. Could you remove the new Add impls and only keep imin/imax on this pull request?

jswrenn:
The `Point` op impls are a little inconsistent. [`Neg`](http://nalgebra.org/rustdoc/nalgebra/geometry/struct.Point.html#impl-Neg) is implemented for `Point`, as is [`Sub<Point>`](http://nalgebra.org/rustdoc/nalgebra/geometry/struct.Point.html#impl-Sub%3CPoint%3CN%2C%20D%3E%3E), but _not_ `Add`. So you can `p1 - -p2` but you can't `p1 + p2`.

The `Sub<Point>` implementations should probably be removed, too.

Ralith:
Subtracting points is meaningful. Note that the resulting value is a `Vector`, not another `Point`.

bemeurer:
@sebcrozet My bad, I was going to make 2 PR's but screwed up. Now this one only has the imax/min(). Cheers!

sebcrozet:
@bemeurer Thanks! It looks like the travis failure is not related to your work (and also occurs recently on master) so I'll merge it.

Issue #369, opened 2018-08-18
Compile issues on  nightly


BaxterEaves:
Updated nightly this morning and rebuilt a package that has nalgebra as a dependency and I got a lot of `not found` type compile errors:

```
 Compiling nalgebra v0.16.0
error[E0432]: unresolved import `geometry::Isometry`
  --> /Users/baxtereaves/.cargo/registry/src/github.com-1ecc6299db9ec823/nalgebra-0.16.0/src/base/cg.rs:15:16
   |
15 | use geometry::{Isometry, IsometryMatrix3, Orthographic3, Perspective3, Point, Point3, Rotation2,
   |                ^^^^^^^^ no `Isometry` in `geometry`. Did you mean to use `isometry`?

error[E0432]: unresolved import `geometry::Isometry`
  --> /Users/baxtereaves/.cargo/registry/src/github.com-1ecc6299db9ec823/nalgebra-0.16.0/src/geometry/rotation_conversion.rs:13:16
   |
13 | use geometry::{Isometry, Point, Rotation, Rotation2, Rotation3, Similarity, SuperTCategoryOf,
   |                ^^^^^^^^ no `Isometry` in `geometry`. Did you mean to use `isometry`?

error[E0432]: unresolved import `geometry::Similarity`
  --> /Users/baxtereaves/.cargo/registry/src/github.com-1ecc6299db9ec823/nalgebra-0.16.0/src/geometry/rotation_conversion.rs:13:65
   |
13 | use geometry::{Isometry, Point, Rotation, Rotation2, Rotation3, Similarity, SuperTCategoryOf,
   |                                                                 ^^^^^^^^^^ no `Similarity` in `geometry`. Did you mean to use `similarity`?

error[E0432]: unresolved import `geometry::Isometry`
  --> /Users/baxtereaves/.cargo/registry/src/github.com-1ecc6299db9ec823/nalgebra-0.16.0/src/geometry/quaternion_conversion.rs:11:16
   |
11 | use geometry::{Isometry, Point3, Quaternion, Rotation, Rotation3, Similarity, SuperTCategoryOf,
   |                ^^^^^^^^ no `Isometry` in `geometry`. Did you mean to use `isometry`?

error[E0432]: unresolved import `geometry::Similarity`
  --> /Users/baxtereaves/.cargo/registry/src/github.com-1ecc6299db9ec823/nalgebra-0.16.0/src/geometry/quaternion_conversion.rs:11:67
   |
[...]

PR #374, opened 2018-08-31
Add keyword "math" and category "science"

lucatrv:


sebcrozet:
Thanks!

CI fails on nightly because of #369 .
Merging this anyway as this change is unrelated.

PR #397, opened 2018-09-21
Add magnitude synonym functions for ease of use

Xaeroxe:
No new math here, just adding some common synonyms for the `norm()` functions.  I didn't know to look for the name `norm()` when browsing the docs so I got pretty confused.  The name `magnitude` is much more common in other math libraries I've used so these names are more intuitive to me.


sebcrozet:
Thanks!

PR #399, opened 2018-09-23
nalgebra-glm: add a GLM-like API to nalgebra.

sebcrozet:
This adds the `nalgebra-glm` crate, which exposes and API similar to the the one from the C++ [GLM](https://glm.g-truc.net/) library. It is mostly focused in computer graphics, using homogeneous coordinates.

Because all types of `nalgebra-glm` are aliases to types from `nalgebra`, transition and interoperations between both is immediate.


sebcrozet:
> It's kind of handy but i've been thinking lately of switching it to the way you are doing it here cause i suspect it would make compiling faster but haven't done any tests yet

I don't know about compilation times. What I did is mostly to follow the GLM types so people find what they may already be used to.

> I have some functions and macros to create vectors with a similar syntax to glm/glsl:

The macro constructors is very interesting, I did not think of that! That's definitely something worth considering. I will create an issue for this.

> I'll give this a try with projects where i'm using na and let you know if i find anything else.

Great, feedbacks are welcome! I am working on a documentation page on `nalgebra.org` and will release a first version later this afternoon.

arturoc:
Couple more comments from a quick look:

- Would it make sense to export aliases for the Point* types? Point2 -> Pnt2..., I've come to find them really useful to create apis where a parameter is clearly a position rather than a vector

- You have Quaternion -> Quat but the most used is usually UnitQuaternion, I use it as UnitQuat in https://github.com/arturoc/na

- Might re-exporting some types from alga/nalgebra make sense so people don't have to import all the 3. For example you have a Number trait here but alga::Real is sometimes useful cause some operations only might make sense on real types so re-exporting it from this crate would allow to create apis using na-glm that accept only reals. Unit is also useful

- glm has t_vec2, t_vec3 for generic types perhaps this crate could have TVec2<T>, TVec3<T>, TPnt2<T> as well so one can use the most common Vec2 but also have the possibility for generic types

- exporting `type Vec = Vector<N...>` is kind of problematic cause the one can't import the whole crate without collisions with std's Vec

- Rotation3 and Isometry3 are sometimes useful, although the equivalent functions to create rotation/translation matrices in this crate are really nice the original types cover some more cases, what about aliasing them to Rot3, Iso3?

arturoc:
One more thing, you are defining a zero function in constants: https://github.com/rustsim/nalgebra/pull/399/files#diff-d90d7ea9f3ae947aea5858e76bf257dcR136  that only works for numbers but nalgebra already has one that works for anything that implements identiy over addition or something like that so re-exporting that one would be much more flexible

sebcrozet:
> Would it make sense to export aliases for the Point* types? Point2 -> Pnt2..., I've come to find them really useful to create apis where a parameter is clearly a position rather than a vector

For now I prefer not to add them. The current goal is to get close to the glm API, which does not include points. Including point aliases would require to actually use them through the API of nalgebra-glm, which will make things more complicated for newcomers that are not familiar with points. Those that like to manipulate points explicitly can always depend on nalgebra too.

> You have Quaternion -> Quat but the most used is usually UnitQuaternion, I use it as UnitQuat in https://github.com/arturoc/na

Yes, and most functions from GLM actually assume the quaternion is normalized. Maybe we should just take `UnitQuat` as argument everywhere instead of just `Quat`. I am still hesitant on this since that would diverge a bit from the original GLM API.

> Might re-exporting some types from alga/nalgebra make sense so people don't have to import all the 3. For example you have a Number trait here but alga::Real is sometimes useful cause some operations only might make sense on real types so re-exporting it from this crate would allow to create apis using na-glm that accept only reals. Unit is also useful

Thank you for the suggestion, I have re-exported some traits and structs: `Real, Scalar, U1, U2, U3, U4, DefaultAllocator`.

> glm has t_vec2, t_vec3 for generic types perhaps this crate could have TVec2, TVec3, TPnt2 as well so one can use the most common Vec2 but also have the possibility for generic types

I did not know about those aliases. I've added them.

> exporting type Vec = Vector<N...> is kind of problematic cause the one can't import the whole crate without collisions with std's Vec

Correct, I've replaced `Vec` by `TVec` (and `Mat` by `TMat`) for consistency. I think it makes since since this looks like `TVec2`, but without the dimension specified.

> Rotation3 and Isometry3 are sometimes useful, although the equivalent functions to create rotation/translation matrices in this crate are really nice the original types cover some more cases, what about aliasing them to Rot3, Iso3?

Same comment as for defining aliases for points. I am still not sure if they are desired as they may be perceived as elements that complicate the API of nalgebra-glm.

> One more thing, you are defining a zero function in constants: https://github.com/rustsim/nalgebra/pull/399/files#diff-d90d7ea9f3ae947aea5858e76bf257dcR136 that only works for numbers but nalgebra already has one that works for anything that implements identiy over addition or something like that so re-exporting that one would be much more flexible

I've changed this too, thanks!

I will merge and release this now as soon as the CI passes.

PR #403, opened 2018-09-23
Fix typo in GLM Mat4 docs

17cupsofcoffee:
Was reading through the docs for the new GLM library (which looks fantastic, by the way - looking forward to trying it out!) and spotted a very small typo/copy and paste error  - it says that `Mat4` is a 4x3 matrix, rather than a 4x4 matrix.


sebcrozet:
Thanks!

PR #405, opened 2018-09-24
Typo fixes.

waywardmonkeys:


sebcrozet:
Thanks a bunch! I will merge this this evening. Did you use a specific tool/IDE plugin to do this spellchecking on comments only?

waywardmonkeys:
I have spell checking enabled in vim and it does it. I was just scrolling through each file ... I should get a better workflow some day, but it is also a nice way to quickly familiarize myself with a codebase.

sebcrozet:
> it is also a nice way to quickly familiarize myself with a codebase.

I see. Interesting approach!

PR #411, opened 2018-09-25
Introduce Unit::from_ref_unchecked

Ralith:
Just an idea I had after writing an API that takes `&Unit<Vector3<f64>>`; not sure if it's a good one. Note that `#[repr(transparent)]` was stabilized in 1.28.


sebcrozet:
That's interesting. I didn't know about `#[repr(transparent)]` but it seems appropriate here. Thanks!

PR #414, opened 2018-09-25
Implement map_with_location

### LINKED ISSUE #375: Add a method to map elements on a matrix, providing the current (row, column) index.
This method would be similar to `.map()`, except the closure would take an extra argument with the current row and column indices currently being mapped. See the original suggestion on the [forum](https://discourse.nphysics.org/t/questions-remarks-around-computation-efficiency/243).

mborst:
Fix #375.


src/base/matrix.rs
```diff
@@ -360,6 +360,29 @@ impl<N: Scalar, R: Dim, C: Dim, S: Storage<N, R, C>> Matrix<N, R, C, S> {
         res
     }
 
+    /// Returns a matrix containing the result of `f` applied to each of its entries. Unlike `map`,
+    /// `f` also gets passed the row and column index, i.e. `f(value, row, col)`.
+    #[inline]
+    pub fn map_with_location<N2: Scalar, F: FnMut(N, usize, usize) -> N2>(&self, mut f: F) -> MatrixMN<N2, R, C>
```
sebcrozet: The argument order should be `FnMut(usize, usize, N)` so this is more in line with the rest of the API that generally takes indices as the first arguments, and more similar to what would give `.enumerate().map(f)`.

mborst:
Done.

PR #417, opened 2018-09-26
Improve cross-linking and cross-referencing in docs.

waywardmonkeys:
I think this is a good start on how the docs could improve. I'll keep doing more of this if you like this direction.


sebcrozet:
That's a great idea! It would be definitely useful if you take some time to add more of those. Thank you for your time!

One small note: links pointing from the nalgebra-glm doc to the nalgebra doc will not work on the [website](https://www.nalgebra.org/rustdoc_glm/nalgebra_glm/index.html). But that's probably something that has to be fixed post-documentation-generation on a script dedicated to the documentation integration on the website. A solution that would work on both the offline rustdoc-generated doc and the online one would be too complicated or impossible.

PR #420, opened 2018-09-28
Add `Into` impls for converting `MatrixSlice`s into `MatrixMN`s

### LINKED ISSUE #342: Add an Into impl for matrix slices to convert them to plain matrices.
Currently, to convert a matrix slice to a plain matrix that owns its data, e.g., a `VectorSlice3` into a `Vector3`, one has to call either `.clone_owned()` or `.into_owned()`. For this to be more idiomatic, we should add an `Into` implementation with the same effect as `.into_owned()` to matrix slices.

jswrenn:
Resolves #342 by adding a bunch of `From` impls.

The fact that this couldn't be accomplished with one, more generic `From` impl has some consequences:
 - All the additional impls add visual noise to the rustdoc.
 - If you get used to using `.into()` for this purpose, you might be surprised that it stops working when you try to write more generic code over `MatrixMN`s (and you need to use the `into_owned` method instead).

### DIFF
--- a/src/base/conversion.rs
+++ b/src/base/conversion.rs
@@ -5,12 +5,16 @@ use std::convert::{AsMut, AsRef, From, Into};
 use std::mem;
 use std::ptr;
 
+use std::ops::Mul;
+use typenum::Prod;
+use generic_array::ArrayLength;
+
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thanks!

PR #425, opened 2018-10-04
Fix the glm::sign function to match its documentation.

### LINKED ISSUE #422: glm: `sign` docs don't reflect reality
Rust's `signum` function returns `-1.0` or `1.0` and don't return `0.0` for zero values.

The docs for the `sign` function in `glm` indicate that it should return `0` when the input value is 0.

sebcrozet:
Fix #422.


waywardmonkeys:
Great to see that you added comments and doc tests to this like I've been doing. :)

PR #427, opened 2018-10-04
Vector and quaternion alias improvements.

waywardmonkeys:
By aliasing to the aliases defined within glm rather than going
directly to nalgebra types / aliases, we're able to improve the
experience of linking through the docs in the glm docs.

For example, with TVec1 defined in terms of TVec rather than Vector1,
then the docs link from TVec1 to TVec and we're able to better
explain the concepts at the glm level.


waywardmonkeys:
No idea why this failed on Travis ...

sebcrozet:
Good idea.
The fail on travis is nothing worry about (unrelated to your work). The cholesky-related tests happen to fail randomly on the CI, likely due to some badly conditioned random matrix. I will have to investigate this at some point.

sebcrozet:
Thanks!

PR #431, opened 2018-10-05
clippy: Use println!(f), without empty string.

waywardmonkeys:


sebcrozet:
Didn’t know this was possible. Thanks!

PR #440, opened 2018-10-08
Doc improvements

waywardmonkeys:


sebcrozet:
Thanks!

Tell me if you would prefer that I keep a single PR open instead of merging them progressively (which I do so I can update the API reference on [nalgebra.org](https://nalgebra.org) regularly).

waywardmonkeys:
No need to keep a single PR open ... I like that we're landing things regularly and that I know that you're looking at things. :)

PR #441, opened 2018-10-08
Change the SVD methods to return a Result instead of panicking

### LINKED ISSUE #358: Make SVD::solve return an Option insetad of panicking.
In `svd.rs`:

>    /// Returns `None` if the singular vectors `U` and `V` have not been computed.

But it panics, and it actually can't return `None` because it doesn't return an `Option`. I don't know whether the comment or the code needs to change.

JD557:
Fix #358.

I changed the SVD methods to return a `Result` instead of an `Option`, since there are multiple types of errors. Knowing the error message might be useful.

I'm pretty new to rust and nalgebra, so I'm not sure if I'm doing everything right. For example, some questions in my mind right now:
* Should I also change the return types of the methods in `nalgebra-lapack/src/svd.rs`?
* Is is OK to return a `Result<MatrixMN<N, R, C>, &'static str>`:
  * Should I return a custom error type?
  * Is it OK for the lifetime of that string slice to be `static` (I'm pretty much a noob when it comes to lifetimes)? Should I return a `String` instead?

### DIFF
--- a/src/linalg/svd.rs
+++ b/src/linalg/svd.rs
@@ -485,89 +485,97 @@ where
 
     /// Rebuild the original matrix.
     ///
-    /// This is useful if some of the singular values have been manually modified.  Panics if the
-    /// right- and left- singular vectors have not been computed at construction-time.
-    pub fn recompose(self) -> MatrixMN<N, R, C> {
-        let mut u = self.u.expect("SVD recomposition: U has not been computed.");
-        let v_t = self.v_t
-            .expect("SVD recomposition: V^t has not been computed.");
+    /// This is useful if some of the singular values have been manually modified.
+    /// Returns `Err` if the right- and left- singular vectors have not been
+    /// computed at construction-time.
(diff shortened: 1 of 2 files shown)


JD557:
I've looked again at the lapack codebase and I don't think it makes sense to return `Result` there, as the API calls never fail.

I just skimmed through the logic, but it appears that there are no `compute_u`/`compute_v` flags and the code works just fine if `epsilon < 0.0`.

What's your opinion on this?
* Should I keep the lapack API unchanged ("if it ain't broke don't fix it")?
* Should I change the return type to `Result<MatrixMN<N, R, C>, &'static str>` and never fail?  (`Result<MatrixMN<N, R, C>, !>` would probably be more correct, but it appears that the bottom type is still experimental) 
* Should I change the lapack API to check that `epsilon >= 0`?

sebcrozet:
@JD557 Thanks! Let's keep the Lapack impl the way it is for now.
Since this is a breaking change, I will merge this to the `dev` branch until the next point release is made.

PR #455, opened 2018-10-27
Add lerp to vectors.

### LINKED ISSUE #453: Add a lerp method to vectors.
This already exists on [nalgebra-glm](https://www.nalgebra.org/rustdoc_glm/nalgebra_glm/fn.lerp.html) and could be added to nalgebra too.

sebcrozet:
Fix #453

### DIFF
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -1,4 +1,4 @@
-use num::Zero;
+use num::{One, Zero};
 use num_complex::Complex;
 #[cfg(feature = "abomonation-serialize")]
 use std::io::{Result as IOResult, Write};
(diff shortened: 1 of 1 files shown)


jswrenn:
Minor optimization: Take `self` by-move:
```rust
pub fn lerp<S2: Storage<N, D>>(self, rhs: &Vector<N, D, S2>, t: N) -> VectorN<N, D>
where DefaultAllocator: Allocator<N, D>
{
    self.axpy(t, rhs, N::one() - t);
    self
}
```

This avoids incurring the cost of cloning `self` when the library user may not need the original value of self again. (And if they do, they get it for free, ergonomically, just by using whatever identifier name they've called `lerp` on, since `Vector` is `Copy`.)

Issue #378, opened 2018-09-03
Add a method to create a matrix of vector from a Vec.


sebcrozet:
For dynamically-sized matrices and vectors, this would be much more efficient than the current solutions (`::from_row_slice` and `::from_column_slice`) which copy the content of a slice. See the suggestion on the 5th point of this [post](https://discourse.nphysics.org/t/questions-remarks-around-computation-efficiency/243).

sebcrozet:
Note that the current way of doing this is quite complicated: `DMatrix::from_data(MatrixVec::new(Dynamic::new(nrows), Dynamic::new(ncols), vec))` where `nrows, ncols, vec` are the dimension of the matrix and the Vec.

mpizenberg:
Oh good to know there is a way to do it! I never looked into `MatrixVec` internals yet.

sebcrozet:
Fixed by #464.

PR #464, opened 2018-11-07
Introduces a `from_vec` constructor for `MatrixMN`.

jswrenn:
Addresses #378.

The implementation of `from_vec` simply wraps a call to `from_iterator_generic`. Because the standard library's implementation of `into_iter` [is specialized](https://github.com/rust-lang/rust/issues/46084#issuecomment-345524508) for the `.into_iter().collect<Vec<_>>()` case, this is equivalent to re-using the original Vec as storage.

### DIFF
--- a/src/base/construction.rs
+++ b/src/base/construction.rs
@@ -596,6 +596,13 @@ macro_rules! impl_constructors(
             ) -> Self {
                 Self::from_distribution_generic($($gargs, )* distribution, rng)
             }
+
+            /// Creates a matrix backed by a given vector.
+            #[inline]
+            #[cfg(feature = "std")]
+            pub fn from_vec($($args: usize,)* data: Vec<N>) -> Self {
+                Self::from_vec_generic($($gargs, )* data)
+            }
         }
 
         impl<N: Scalar, $($DimIdent: $DimBound, )*> MatrixMN<N $(, $Dims)*>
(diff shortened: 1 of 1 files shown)


jswrenn:
Note: this addition _definitely_ needs to be accompanied by an update to the [_Matrix Construction_](https://nalgebra.org/vectors_and_matrices/#matrix-construction) section of the guide but the PR, as it stands, doesn't include that.

sebcrozet:
Thanks @jswrenn! I did not know about this specialization, it's great.

A couple of details though: the method you added should be named `from_vec_generic` because it takes its dimension arguments explicitly with types `R` and `C`.

This is allow you to add other methods called `from_vec` that will take the right number of `usize` arguments for the required dimensions not known at runtime (for example one `usize` for `DVector::from_vec`, two `usize` for `DMatrix::from_vec` and no dimension arguments at all for statically-sized matrices). For example you may take a look at the [from_element](https://github.com/rustsim/nalgebra/blob/master/src/base/construction.rs#L337-L340) implementation (which also has a corresponding [from_element_generic](https://github.com/rustsim/nalgebra/blob/master/src/base/construction.rs#L38-L43) method). Something similar should be added for `from_vec` (inside on the same macro as `from_element`:

```rust
#[inline]
pub fn from_vec($($args: usize,)* vec: Vec<N>) -> Self {
    Self::from_vec_generic($($gargs, )* vec)
}
```

sebcrozet:
I have created an issue for adding `from_vec` to the list of construction methods on the user guide: https://github.com/rustsim/nalgebra.org/issues/19 (do not feel obligated to fix it if you don't have the time).

sebcrozet:
Thanks for adding doc-tests too!

Issue #471, opened 2018-11-13
Provide to_homogeneous for MatrixN


grtlr:
There already is a `to_homogeneous` method for `VectorN`, it would be great if nalgebra would provide the same for `MatrixN` (and `SquareMatrix`) to allow homogeneous coordinates in higher dimensions.

Such method would increase the dimensions of `MatrixN` to *N+1* and set `mat[(N,N)] = 1`.

I could try to implement such a method, but right now I am still trying to better understand the structure of nalgebra. @sebcrozet would you be interested in such a method?

sebcrozet:
That would be useful, yes! The implementation and trait bounds would likely look very much like the `.to_homogenous()` method from the `Rotation` type [there](https://github.com/rustsim/nalgebra/blob/master/src/geometry/rotation.rs#L198-L208).

Note however that since there already is a `to_homogeneous` for `VectorN`, you may end up with a duplicated method error (because `VectorN` itself is an alias to a matrix). One way to solve this may be to add the `D: IsNotStaticOne` trait bound where `D` is the dimension type parameter.

grtlr:
Cool! My plan would be to move `to_homogeneous()` from `Rotation` to `MatrixN`. `Rotation` stores a `MatrixN` internally, so then we can just call this method there instead.

As you have predicted, I am running into problems with duplicated method errors. Adding `D: IsNotStaticOne` does not seem to fix this problem. I have added it as a trait bound for the method, and the `impl` block, but still get the errors. Do you have any other ideas on how to solve this problem? Oh and by the way, what is the actual purpose of `IsNotStaticOne`?

sebcrozet:
The `IsNotStaticOne` trait is implemented by `Dynamic` and by all type-level integer except `U1`. So may sometimes be enough to disambiguate situations where other impls of a trait already exist for vectors because vectors are just matrices with a number of columns equal to `U1`.

Would you mind showing what you tried so far so I can try help figuring out how to deal with this? By opening a PR maybe?

grtlr:
I think that I got the disambiguation to work, but now I'm having trouble to get the `Allocator` to work with `DimNameSum`. Here is the PR: #475. Could you maybe take a short look at it? Thank you!

PR #475, opened 2018-11-18
RFC: to_homogeneous for MatrixN

grtlr:
First attempt at #471. It seems that the disambiguation using `IsNotStaticOne` is working, but I can't get it to compile. 

Currently, I get the following error message:

```
error[E0277]: cannot multiply `<D as base::dimension::DimName>::Value` to `<D as base::dimension::DimName>::Value`
   --> src/base/matrix.rs:833:1
    |
833 | / impl<N: Scalar + One, D: DimNameAdd<U1> + IsNotStaticOne> MatrixN<N, D> {
834 | |
835 | |     /// Yields the homogeneous matrix for this matrix, i.e., appending an additional dimension and
836 | |     /// and setting the diagonal element to `1`.
...   |
844 | |
845 | | }
    | |_^ no implementation for `<D as base::dimension::DimName>::Value * <D as base::dimension::DimName>::Value`
    |
    = help: the trait `std::ops::Mul` is not implemented for `<D as base::dimension::DimName>::Value`
    = help: consider adding a `where <D as base::dimension::DimName>::Value: std::ops::Mul` bound
    = note: required because of the requirements on the impl of `base::allocator::Allocator<N, D, D>` for `base::default_allocator::DefaultAllocator`
```

I think this has to do with how the `Allocator` is set up, but I can't get it to work.

### DIFF
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -20,7 +20,7 @@ use alga::general::{ClosedAdd, ClosedMul, ClosedSub, Real, Ring};
 
 use base::allocator::{Allocator, SameShapeAllocator, SameShapeC, SameShapeR};
 use base::constraint::{DimEq, SameNumberOfColumns, SameNumberOfRows, ShapeConstraint};
-use base::dimension::{Dim, DimNameAdd, DimAdd, DimNameSum, DimSum, IsNotStaticOne, U1, U2, U3};
+use base::dimension::{Dim, DimAdd, DimSum, IsNotStaticOne, U1, U2, U3};
 use base::iter::{MatrixIter, MatrixIterMut};
 use base::storage::{
     ContiguousStorage, ContiguousStorageMut, Owned, SameShapeStorage, Storage, StorageMut,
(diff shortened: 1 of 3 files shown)


src/base/matrix.rs
```diff
@@ -830,6 +830,22 @@ impl<N: Scalar, D: Dim, S: Storage<N, D, D>> SquareMatrix<N, D, S> {
     }
 }
 
+impl<N: Scalar + One + Zero, D: Dim + DimAdd<U1> + IsNotStaticOne, S: Storage<N, D, D>> Matrix<N, D, D, S> {
```
sebcrozet: If I am not mistaken, `DimAdd` already inherits from `Dim` so you don't need the latter here.
```suggestion
impl<N: Scalar + One + Zero, D: DimAdd<U1> + IsNotStaticOne, S: Storage<N, D, D>> Matrix<N, D, D, S> {
```

sebcrozet:
Sorry, I forgot to answer your question:

> While this would be no problem, I'm not sure if it makes sense - what could this be used for, and what is your preference on this?

Let's just keep this for square matrices for now and see if someone complains about it in the future (this will give us a hint about the possible use-case).

sebcrozet:
All good, thanks for your contributions and all the changes!

PR #480, opened 2018-11-19
Add as_ptr and as_mut_ptr methods

jturner314:
This is useful for moving across a matrix/slice in unsafe code and for converting slices to other types (e.g. `ndarray::ArrayView`).

Edit: If you'd like to guarantee that the pointer is aligned and non-null for empty matrices too, that would be great. I just didn't know if you'd like to make that guarantee, so I left it out.


sebcrozet:
Thanks!

> If you'd like to guarantee that the pointer is aligned and non-null for empty matrices too, that would be great. I just didn't know if you'd like to make that guarantee, so I left it out.

How would this be done?

jswrenn:
It always struck me as a little odd that [`ptr`](https://www.nalgebra.org/rustdoc/nalgebra/base/storage/trait.Storage.html#tymethod.ptr) is a method on `Storage`, instead of [`ContiguousStorage`](https://www.nalgebra.org/rustdoc/nalgebra/base/storage/trait.ContiguousStorage.html).

jturner314:
> Having it be a method on `Storage` seems like it introduces the implicit requirement that it's possible to construct a pointer to the start of the storage space of the matrix, which would preclude implementing, for instance, a `HashMapStorage`.

I assume you're thinking about a sparse representation with indices as keys and matrix elements as values? From what I can tell, `Storage` is restricted to accessing elements by offsetting a pointer. (For example, it has a `strides` method.) Note that getting a pointer and shape/strides is necessary for non-contiguous matrix types like `MatrixSlice` in addition to contiguous matrix types, and a trait like `Storage` allows operating in a uniform way on both contiguous and non-contiguous matrices. If sparse representations are desired in the future, I'd suggest adding a separate, more general trait.

sebcrozet:
Thanks @jturner314! I will merge this now for today's release.
The discussion about alignment is worth keeping in mind so I will create an issue for that.

Issue #479, opened 2018-11-19
Add method to convert a Matrix into a Vec.


sebcrozet:
See https://github.com/rustsim/nalgebra/issues/473#issuecomment-440038489

The goal is to add an `.into_vec` method that would either unwrap the underlying `Vec` storage, or would perform a copy if the storage is not `MatrixVec`.

Note that one way of achieving this would be to add an implementation of `IntoIter` for `Matrix`. Then the `matrix.into_iter().collect()` will do the trick and will not perform any copy for matrices with a `MatrixVec` storage because of [that optimization](https://github.com/rust-lang/rust/issues/46084#issuecomment-345524508).

jswrenn:
> Note that one way of achieving this would be to add an implementation of `IntoIter` for `Matrix`.

Confusingly, this [already sort of exists](https://github.com/rustsim/nalgebra/blob/85bfae53900d0b5415a8375d881875765aa49cdf/src/base/conversion.rs#L74-L94), but [doesn't appear in the rendered rustdoc](https://www.nalgebra.org/rustdoc/nalgebra/index.html?search=IntoIterator).

So, a generic-but-inefficient conversion implementation would be:
```
#[cfg(any(feature = "std", feature = "alloc"))]
impl<N, R, C, S> Into<Vec<N>> for Matrix<N, R, C, S>
where
    N: Scalar,
    R: Dim,
    C: Dim,
    S: Storage<N, R, C>,
{
    fn into(self) -> Vec<N> {
        use std::iter::FromIterator;
        Vec::from_iter(self.into_iter().cloned())
    }
}
```

The `.cloned()` is needed because `self.into_iter()` produces an iterator of references; that's a consequence of the fact that `IntoIterator` is only implemented for `&Matrix`. Implementing `IntoIterator` for `Matrix` where `type IntoIter = MatrixIter` isn't possible. At any rate, doing so wouldn't actual enable that optimization since that optimization only applies to [`vec::IntoIter`](https://doc.rust-lang.org/std/vec/struct.IntoIter.html).

LU15W1R7H:
I'm also interested in this feature. Why isn't it posible to implement `IntoIterator` for `Matrix` where `type IntoIter = MatrixIter`? It should be possible to convert the underlying `Vec` storage into an interator, right?

LU15W1R7H:
[...]

PR #481, opened 2018-11-20
Implement `Into<Vec<N>>` for `MatrixVec<N, R, C>`

jswrenn:
Very narrowly addresses @jturner314's [proximate issue](https://github.com/rustsim/nalgebra/issues/473#issuecomment-440038489) of getting the `Vec` out of a `Matrix` backed by a `Vec`.

This makes possible something that was previously impossible. However, this change isn't enough on it's own to achieve the ergonomic goals of #479.


sebcrozet:
Thanks!

PR #485, opened 2018-11-23
Group swizzle methods by dimensional requirement.

jswrenn:
This is semantically equivalent, but improves the rendered documentation.

Previously, each swizzle method was contained in its own `impl`.


sebcrozet:
Interesting idea, thanks!

PR #490, opened 2018-12-02
Overloaded Indexing Operations

jswrenn:
This P.R. overloads matrix indexing in the same manner that the standard library overloads slice indexing.

## Modes of Indexing
Matrices can be indexed in three ways, each with different safety guarantees:
  - `get` produces `Some` reference to the element(s) at the index, or `None` if the index is out of bounds
  - `index` produces a reference to the element(s) at the index, or panics if the index is out of bounds
  - `get_unchecked` produces a reference to the element(s) at the index, without bounds checking

Each of these has a corresponding `_mut` counterpart that produces a mutable reference to the element(s) at the given location.

## Types of Indices
Matrices can be indexed by a `usize`, or with any pair of `usize`, `Range<usize>`, `RangeFrom<usize>`, `RangeFull`, `RangeInclusive<usize>`, `RangeTo<usize>`, and `RangeToInclusive<usize>`.

Indexing by `usize` and `(usize, usize)` produces a reference to the element at that location. All other two-dimensional indices produce a `MatrixSlice`.  

Some examples:

#### Two-Dimensional Indices
```rust
let matrix = Matrix2::new(0, 2,
                          1, 3);

assert_eq!(matrix.index((0, 0)), &0);
assert_eq!(matrix.index((1, 0)), &1);
assert_eq!(matrix.index((0, 1)), &2);
assert_eq!(matrix.index((1, 1)), &3);
```

#### Linear Address Indexing
```rust
let matrix = Matrix2::new(0, 2,
                          1, 3);

assert_eq!(matrix.get(0), Some(&0));
assert_eq!(matrix.get(1), Some(&1));
assert_eq!(matrix.get(2), Some(&2));
assert_eq!(matrix.get(3), Some(&3));
```

#### Index to a Row
```rust
let matrix = Matrix2::new(0, 2,
                          1, 3);

assert!(matrix.index((0, ..))
    .eq(&Matrix1x2::new(0, 2)));
```

#### Index to a Column
```rust
let matrix = Matrix2::new(0, 2,
                          1, 3);

assert!(matrix.index((.., 0))
    .eq(&Matrix2x1::new(0,
                        1)));
```

#### Index to a Partial Row
```rust
let matrix = Matrix3::new(0, 3, 6,
                          1, 4, 7,
                          2, 5, 8);

assert!(matrix.index((0, ..2))
    .eq(&Matrix1x2::new(0, 3)));
```

#### Index to a Partial Column
```rust
let matrix = Matrix3::new(0, 3, 6,
                          1, 4, 7,
                          2, 5, 8);

assert!(matrix.index((..2, 0))
    .eq(&Matrix2x1::new(0,
                        1)));
```

#### Index to a Range of Rows
```rust
let matrix = Matrix3::new(0, 3, 6,
                          1, 4, 7,
                          2, 5, 8);

assert!(matrix.index((1..3, ..))
    .eq(&Matrix2x3::new(1, 4, 7,
                        2, 5, 8)));
```

#### Index to a Range of Columns
```rust
let matrix = Matrix3::new(0, 3, 6,
                          1, 4, 7,
                          2, 5, 8);

assert!(matrix.index((.., 1..3))
    .eq(&Matrix3x2::new(3, 6,
                        4, 7,
                        5, 8)));
```

## Future Work
This PR can be extended in a non-breaking way to permit:
 - Overloaded linear address indexing.
 - Further overloading two-dimensional indexing whereby the row and column components are ranges of `Dim` (and thus the output can be a `MatrixSlice` of size known at compile-time). This just requires some tricky code-generation work that the current macros can't handle.

### DIFF
--- a/src/base/indexing.rs
+++ b/src/base/indexing.rs
@@ -334,8 +334,8 @@ pub trait MatrixIndexMut<'a, N: Scalar, R: Dim, C: Dim, S: StorageMut<N, R, C>>:
 }
 
 /// # Indexing Operations
-/// ## Indicies to Individual Elements
-/// ### Two-Dimensional Indicies
+/// ## Indices to Individual Elements
+/// ### Two-Dimensional Indices
 /// ```
 /// # use nalgebra::*;
 /// let matrix = Matrix2::new(0, 2,
@@ -359,7 +359,7 @@ pub trait MatrixIndexMut<'a, N: Scalar, R: Dim, C: Dim, S: StorageMut<N, R, C>>:
 /// assert_eq!(matrix.get(3), Some(&3));
 /// ```
 ///
-/// ## Indicies to Individual Rows and Columns
+/// ## Indices to Individual Rows and Columns
 /// ### Index to a Row
 /// ```
 /// # use nalgebra::*;
@@ -381,7 +381,7 @@ pub trait MatrixIndexMut<'a, N: Scalar, R: Dim, C: Dim, S: StorageMut<N, R, C>>:
 ///                         1)));
 /// ```
 ///
-/// ## Indicies to Parts of Individual Rows and Columns
+/// ## Indices to Parts of Individual Rows and Columns
 /// ### Index to a Partial Row
 /// ```
 /// # use nalgebra::*;
@@ -408,7 +408,7 @@ pub trait MatrixIndexMut<'a, N: Scalar, R: Dim, C: Dim, S: StorageMut<N, R, C>>:
 ///     .eq(&Matrix2x1::new(1,
 ///                         2)));
 /// ```
-/// ## Indicies to Ranges of Rows and Columns
+/// ## Indices to Ranges of Rows and Columns
 /// ### Index to a Range of Rows
 /// ```
 /// # use nalgebra::*;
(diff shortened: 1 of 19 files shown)


sebcrozet:
This is amazing! I will take a closer look and get back to you this week.

jswrenn:
9f9e5759282cdf0106c20ec8415e691d5d633ce7 implements the groundwork for indexing with typenums and adds a proof-of-concept `MatrixIndex` implementation for index pairs where one of the components is a `ops::RangeFrom<impl DimName>`.

To support the other range types, we're going to need to do some work to surface more of typenum's type-level operators as `Dim` operators (namely `IsLess`). This seems like a significant undertaking, and I don't think I'll get to it in this PR.

src/base/indexing.rs
```diff
@@ -0,0 +1,714 @@
...
+    /// if out of bounds.
+    #[doc(hidden)]
+    #[inline(always)]
+    fn index_mut(self, matrix: &'a mut Matrix<N, R, C, S>) -> Self::OutputMut {
+        self.get_mut(matrix).expect("Index out of bounds.")
+    }
+}
+
+/// # Indexing Operations
+/// ## Indicies to Individual Elements
```
sebcrozet: I believe the plural of `index` is `indices`.

sebcrozet:
Thanks!

PR #493, opened 2018-12-04
typofix: `DimNameNimimum` -> `DimNameMinimum`

jswrenn:


sebcrozet:
Thanks! That's a very unfortunate typo as fixing it is technically a breaking change (though I doubt anyone actually use this trait besides nalgebra itself). This will released at the same time as #490 then.

Issue #470, opened 2018-11-12
Storage types are inconsistently named.


jswrenn:
The `Storage` trait is implemented for four types:
 1. `SliceStorage`
 2. `SliceStorageMut` 
 3. `MatrixArray`
 4. `MatrixVec`.

These names are natural to me now, but they weren't when I first started using the library.

`MatrixArray` and `MatrixVec` are easily confusable with the myriad of other `Matrix` types. If you search for `Matrix` in the documentation, you're presented with a pretty overwhelming collection of types with `Matrix` in their name. 90% of them are just aliases of `Matrix`. The exceptions, of course, are `MatrixVec` and `MatrixArray`. Trying to do linear algebra operations directly on these types ends in misery.

A savvy user might then conclude "Oh, if I see `Matrix` followed by a datatype, it's actually an underlying storage type." This rule-of-thumb works well enough, but again there are types that confound this pattern: `MatrixSlice` and `MatrixSliceMut` which _are_ aliases for `Matrix`.

Perhaps a more consistent set of names for the Storage types would be:
 1. `SliceStorage`
 2. `SliceStorageMut` 
 3. `ArrayStorage`
 4. `VecStorage`

sebcrozet:
I strongly agree with you @jswrenn! The current names can be confusing and the ones you propose are good ones.

sebcrozet:
THis was fixed by #496. Closing.

PR #496, opened 2018-12-05
Storage type name reform.

jswrenn:
This implements the renaming proposed in #470:
 - `MatrixVec` → `VecStorage`
 - `MatrixArray` → `ArrayStorage`

`MatrixVec` and `MatrixArray` remain as deprecated type aliases to `VecStorage` and `ArrayStorage`, respectively, so this may not actually be a breaking change.


sebcrozet:
Glad this is not a breaking change. Thanks!

Issue #498, opened 2018-12-09
Add features required for the migrassion of rusty-machine to nalgebra


sebcrozet:
See [that issue](https://github.com/AtheMathmo/rusty-machine/issues/199) from rusty-machine.
So far, the following features are missing from nalgebra:

- [ ] Simplify the construction of `DVector` by addressing https://github.com/rustsim/nalgebra/issues/377 This will require breaking changes.
- [ ] Add statistics functions: variance, mean.
- [ ] Add more general norms by addressing https://github.com/rustsim/nalgebra/issues/258
- [ ] Add row and column iterators.

ralfbiedert:
Not sure if this belongs here, but I'm the author of [ffsvm](https://github.com/ralfbiedert/ffsvm-rust) (a SVM classifier) and I also considered migrating to nalgebra for my math, and to simplify data exchange.

We use ffsvm in our company as a module in VR and gaming libraries, where literally every microsecond (and allocation) counts.

Right now there are 2 features missing in nalgebra that prevent me from making a switch. I generally believe they are essential for any performance sensitive SVM implementation; rusty-machine, ffsvm, or otherwise:

- [ ] sparse vectors
- [ ] properly aligned SIMD vectors

**Sparse vectors** are basically needed for high-dimensional SVMs, like word lists, where a vector might have +30k entries, out of which only very few are actually used. Using dense vectors here would be a big performance hit.

Conversely, for "fully dense" SVMs, properly aligned **SIMD vectors** are the way to go. When I started looking into ffsvm about a year ago I evolved the code from "naive" over "manual vectorization" to "actual SIMD":

![perf](https://raw.githubusercontent.com/ralfbiedert/ffsvm-rust/master/docs/performance_history.v4.png)

At least back then, there was a 2x speed difference between "trusting the compiler" and "do SIMD yourself". I wouldn't be surprised if that difference is up to >4x even today when considering various compiler flags, target platforms, ... On top of that (not properly reflected in the graph because the test vectors were so big) is a performance impact when dealing with unaligned SIMD data.

If I had to prioritize between these I'd pick SIMD over sparse vectors.

Long story short, I have seen you have tracking issues for these items already. However, I wanted to post this as more of a note that I would also love to align with rustsim / nalgebra, but those issues are holding me back.

sebcrozet:
[...]

PR #499, opened 2018-12-09
Add features required by rusty-machine.

sebcrozet:
See #498.


sebcrozet:
@jswrenn I believe the convention is for the iterator type to have the same name as the method returning it. In our case the methods are `row_iter()`, `row_iter_mut()` etc, hence the `RowIter`, `RowIterMut`, etc names.

I didn't call those methods `rows` and `rows_mut` in the first place because those names are already used for slicing the matrix.

jswrenn:
I forgot about those methods! You're right about the convention. The name conflict is unfortunate.

sebcrozet:
I'm merging this for today's release.
#499 is not fully fixed yet so we will open another PR to continue the work.

Issue #460, opened 2018-10-31
unwrap vs into_inner


whatisaphone:
This is a slightly bikesheddy issue, sorry in advance :)

In the Rust ecosystem, `unwrap` is used for operations that can panic. As such, when you see the word "unwrap" in code, it's a 🚩 red flag 🚩 that should jump out at you as a possible bug when you're reading code.

nalgebra seems to use `unwrap()` as the method name for unwrapping a wrapper type (e.g. getting a `Quaternion` from a `UnitQuaternion`). Those calls always succeed, because there's never a case where `UnitQuaternion` does not contain a `Quaternion`. That usage does not line up with the rest of the Rust ecosystem.

The name that's been adopted for "unwrapping" operations which always succeed is `into_inner`:

https://doc.rust-lang.org/std/cell/struct.Cell.html#method.into_inner
https://doc.rust-lang.org/stable/std/?search=into_inner

Another example from a third-party crate:

https://indiv0.github.io/lazycell/lazycell/struct.LazyCell.html#method.into_inner

Can I suggest deprecating your functions named `unwrap` and adopting `into_inner` instead?

sebcrozet:
Thank you for noticing this! I agree we should deprecate `unwrap` in favor of `into_inner` as this seems to be the unspoken convention for this.

sebcrozet:
By the way @whatisaphone, do not hesitate to report other things that look unidiomatic to you regarding the nalgebra API. The nalgebra crate existed before the Rust community started to settle down on those kind of convention, so there might be a few things that do not comply with the new trends yet.

Those kinds of details are what makes the life of users much simpler.

whatisaphone:
Glad to help! I'll open another issue if I notice anything else.

sebcrozet:
This was fixed by #500. Closing.

PR #500, opened 2018-12-09
Rename `unwrap` to `into_inner`.

jswrenn:
Implements #460 for `Unit`, `Rotation`, `Transform`, `Orthographic3` and `Perspective3`. In all cases, `unwrap()` is retained, but deprecated.


sebcrozet:
Thanks!

PR #512, opened 2018-12-27
Re-export alga docs to improve discoverability of trait implementations

jswrenn:
Implements the suggestion I made in https://github.com/rustsim/nalgebra/issues/511#issuecomment-450224279.

This provides a one-stop place to find out which traits from `alga` are implemented for types from `nalgebra`.

For example, here's how rustdoc renders `alga_documentation::linear::AffineTransformation`:
screenshot--2018 12 27-15-18-27

Once [`doc_cfg`](https://github.com/rust-lang/rust/pull/53076) is stabilized, the `deprecated` notice should be superceded (or augmented) by `#[cfg(rustdoc)]`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,6 +148,15 @@ use num::Signed;
 
 pub use alga::general::{Id, Real};
 
+/// Documentation for traits implemented from `alga`.
+///
+/// For documentation purposes only; do not rely on items from this module.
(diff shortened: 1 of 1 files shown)


Andlon:
Since `nalgebra` these days is based primarily upon `simba`, my understanding is that this PR is no longer relevant. I'll therefore close it.

Issue #514, opened 2018-12-28
std::iter::Sum for DVector and DMatrix


grtlr:
Currently, there is no way to use ``std::iter::Sum`` with an iterator over matrices (or vectors). I believe the reason for this is that #264 implements the trait in terms of ``DimName`` and not the more general ``Dim`` type.

This makes a lot of sense because then the dimensions of the resulting matrix are known in advance. For dynamic matrices, we would have to check (or make the assumption) that all elements have the correct dimensions. A straightforward approach would be to have the first element determine the size of the allocation. The problem with this approach is that there might not be a first element (the sequence is empty):

    let mut it = seq.iter();
    let x = it.next().unwrap().clone();
    let res = it.fold(x, |acc, v| acc + v);

Do you see a way to make ``std::iter::Sum`` work with the dynamic types?

PR #517, opened 2018-12-29
`iter::Sum` for `DMatrix`

jswrenn:
This is one potential way to address #514. It tackles the problem in three steps:
1. The foundation of this approach is to [introduce](https://github.com/rustsim/nalgebra/commit/ccdd393700e83e3093b1986e67a9778c18a2b1cc) a notion of `Default` for dimensions. Statically known dimensions default to their value (this is already the behavior of the `Default` implementation provided by typenum). The `Default` of a `Dynamic` dimension is 0.
2. Using this, we can [then](https://github.com/rustsim/nalgebra/commit/1dd762d17460ec7fa6765010ed84bd9e29b0c293) lift `DimName` requirements on `Matrix` constructors that required them; any dynamic dimensions simply default to zero size.
3. [Finally](https://github.com/rustsim/nalgebra/commit/d60f138f626bc8faa8bc2eb37536435e7fdc6bc9), we can lift the `DimName` requirements on functions that used these constructors, like the implementation of `iter::Sum`.

This PR only lifts the `DimName` requirement on `iter::Sum`, but the approach is broadly applicable and we can use it to lift `DimName` requirements in many parts of nalgebra.

### DIFF
--- a/src/base/construction.rs
+++ b/src/base/construction.rs
@@ -681,14 +681,14 @@ impl_constructors!(Dynamic, Dynamic;
  * Zero, One, Rand traits.
  *
  */
-impl<N, R: DimName, C: DimName> Zero for MatrixMN<N, R, C>
+impl<N, R: Dim, C: Dim> Zero for MatrixMN<N, R, C>
 where
     N: Scalar + Zero + ClosedAdd,
     DefaultAllocator: Allocator<N, R, C>,
 {
     #[inline]
     fn zero() -> Self {
-        Self::from_element(N::zero())
+        Self::zeros_generic(R::default(), C::default())
     }
 
     #[inline]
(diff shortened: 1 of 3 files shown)


grtlr:
That was a really quick response, thank you very much!

I have not found any test cases for the implementation of `Sum` in the current code base. If you want, I could write up some tests for the static and the dynamic case (as well as `DVector`)?

grtlr:
@jswrenn I have created a PR on your branch. If you have any comments regarding the formatting, feel free to let me know.

jswrenn:
I've opened a separate pull request implementing the panic-on-empty behavior: #552.

Sorry for the slow response; I've been recovering from some medical issues.

sebcrozet:
@jswrenn No worry. Take care!

PR #525, opened 2019-01-16
Fix out-of-bounds panic during deserialization; produce error instead

### LINKED ISSUE #522: Deserialization using Serde may panic
Deserialization of a `ArrayStorage` can panic if the input length is too large. This is due to [that indexing](https://github.com/rustsim/nalgebra/blob/master/src/base/array_storage.rs#L334). The value of `curr` should be checked and an error of type `V::Error::invalid_length` should be returned.

jswrenn:
Fixes #522


sebcrozet:
Thanks @jswrenn!

Issue #472, opened 2018-11-15
Rename ::new_observer_frame to ::face_towards


sebcrozet:
I find the `::face_towards` name more representative of what the method does. This name was suggested by @magnonellie [there](https://github.com/amethyst/amethyst/issues/1134) (in the context of the Amethyst game engine).

PR #527, opened 2019-01-16
Rename ::new_observer_frame to ::face_towards

manuelmauro:
Addresses #472


Ralith:
This could be made non-breaking by retaining `new_observer_frame` as a deprecated wrapper function.

manuelmauro:
Thanks! I didn't know what was the policy before v1.0. I pushed some more code, let me know if it needs something else.

Issue #520, opened 2019-01-05
Implement convolution.


sebcrozet:
This has been suggested on the [forum](https://discourse.nphysics.org/t/convolutions-in-nalgebra/299), taking the [convolve](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.convolve.html) function from scipy as reference.

npapapietro:
Does the fft method required for this? Is direct sufficient based on nalgebra's applications?

sebcrozet:
@npapapietro An implementation based on the FFT would be the best, but a direct method would be good too for a first implementation. It's always better to have the feature even if it's slow, than not having the feature at all!

sebcrozet:
Fixed by #539.

PR #539, opened 2019-02-10
Implement convolution #520

npapapietro:
https://github.com/rustsim/nalgebra/issues/520

First time contributing to library and relearning rust. Will take any and all advice.


ChristopherRabotin:
Just browsing through here, and by no means am I knowledgeable on nalgebra, but I was wondering why you're using DVector in your tests instead of the statically allocated Vector<N, T>?

sebcrozet:
Hi @npapapietro !

Thank you for this pull request! I will have more time to give you a feedback on this next Saturday but here are some preliminary remarks:

- The convention within nalgebra is to use `N` (meaning "number") for the scalar type parameter. We already use `R` for the number of rows of a matrix.
- Another convention: because you are involving two vectors with different dimensions, you should use `D1` and `D2` for their dimension types (instead of `D` and `E`).
- Another convention: because you are involving two storages, you should use `S1` and `S2` for their type parameters (instead of `S` and `Q`).

Those conversions are quite arbitrary but necessary to keep things homogeneous within the crate.

- Your functions should not return a dynamic vector `Matrix<R, Dynamic, U1, VecStorage<R, Dynamic, U1>>`. Instead, it should return a vector that may possibly have a static size if `D1` and `D2` (that is `D` and `E` in your code) both have static sizes. I will provide more details on this next Saturday. @ChristopherRabotin That's why `DVector` had to be used on the tests.
- Those free functions should be methods instead. Something like this:

```rust
impl<N: Real, D1: Dim, S1: Storage<N, D1>> Vector<N, D1, S1> {
    pub fn convolve_full<D2: Dim, S2: Storage<N, D2>>(&self, kernel: Vector<N, D2, S2>) -> ... {
    }
}
```

I'll get back to you next saturday about what to put as the return type to take my comment into account.

npapapietro:
> Just browsing through here, and by no means am I knowledgeable on nalgebra, but I was wondering why you're using DVector in your tests instead of the statically allocated Vector<N, T>?

Dvector means any sized vector can be passed in while I needed to choose set inputs so I can assert the correct correct answer.

sebcrozet:
@npapapietro As promised, here are the trait bounds you need to use in order to be able to return a non-dynamic vector (I've taken into account some of my comments above regarding nalgebra conventions too). I've taken the example of the first function, `convolve_full`:

```rust
pub fn convolve_full<N: Real, D1: Dim, D2: Dim, S1: Storage<N, D1>, S2: Storage<N, D2>>(
    vector: Vector<N, D1, S1>,
    kernel: Vector<N, D2, S2>,
) -> VectorN<N, DimSum<D1, D2>>
    where D1: DimAdd<D2>,
          D2: DimAdd<D1, Output = <D1 as DimAdd<D2>>::Output>, // Needed because of the recursive call.
          DefaultAllocator: Allocator<N, DimSum<D1, D2>> {
```

The `DimSum` and `DimAdd` constructs are what will build a type representing the sum of both dimensions `D1` and `D2`. Also note that the `Allocator` trait requires an `use allocator::Allocator;`.

In addition to that, here is how you can initialize your result vector:

```rust
    let result_len = vector.data.shape().0.add(kernel.data.shape().0);
    let mut conv = VectorN::zeros_generic(result_len, U1);
```

This will effectively only allocate dynamically if either `vector` or `kernel` is a dynamic vector too. Otherwise, the result will be statically allocated on the stack since its dimension will be known at compile-time.

npapapietro:
@sebcrozet Thanks for the feedback, I have been spending a lot time reading very deep into the dimension trait setups. I think I understand it much better than before. However I still am getting hung on how to declare a return type of 
D1 + D2 - 1. I've tried to write it as 
```rust
->  VectorN<N, DimDiff<DimSum<D1, D2>, U1>>
```
and various other combinations as well as adding to the ```where``` clause to try and get the output as a Dim or DimName.  I think either I'm missing something or this can't be done in this way because there are too many unknowns at compile time.

sebcrozet:
@npapapietro Sorry I didn't realize you needed `D1 + D2 - 1` instead of `D1 + D2`. Here is what you need:

```rust
pub fn convolve_full<N, D1, D2, S1, S2>(
    vector: Vector<N, D1, S1>,
    kernel: Vector<N, D2, S2>,
) -> VectorN<N, DimDiff<DimSum<D1, D2>, U1>>
    where N: Real,
          D1: DimAdd<D2>,
          D2: DimAdd<D1, Output = <D1 as DimAdd<D2>>::Output>, // Needed because of the recursive call.
          DimSum<D1, D2>: DimSub<U1>, // Needed to subtract `U1` from `D1 + D2`.
          S1: Storage<N, D1>,
          S2: Storage<N, D2>,
          DefaultAllocator: Allocator<N, DimDiff<DimSum<D1, D2>, U1>> {
```

And for initializing your result:

```rust

    let result_len = vector.data.shape().0.add(kernel.data.shape().0).sub(U1);
    let mut conv = VectorN::zeros_generic(result_len, U1);
```

So the key here is to add `DimSum<D1, D2>: DimSub<U1>,` in the `where` clause to tell the compiler you can subtract 1 to the result of `D1 + D2`. This is quite complicated, but unfortunately necessary until Rust gains builtin support of const-generics.

npapapietro:
I was able to follow what you had written and apply it to the "same" padding and get the correct answers as well as constructing the valid case, but the issue  is an implementation from another crate, when passing in VectorN's.
```rust
pub fn convolve_valid<N, D1, D2, S1, S2>(
    vector: Vector<N, D1, S1>,
    kernel: Vector<N, D2, S2>,
) -> VectorN<N, DimSum<DimDiff<D1, D2>, U1>>
where
    N: Real,
    D1: DimSub<D2>,
    D2: DimSub<D1, Output = DimDiff<D1, D2>>,
    DimDiff<D1, D2>: DimAdd<U1>,
    S1: Storage<N, D1>,
    S2: Storage<N, D2>,
    DefaultAllocator: Allocator<N, DimSum<DimDiff<D1, D2>, U1>>
```

generates this error:
```rust
error[E0277]: the trait bound `typenum::uint::UTerm: typenum::private::PrivateSub<typenum::uint::UInt<typenum::uint::UTerm, typenum::bit::B1>>` is not satisfied
  --> examples\convolution.rs:12:21
   |
12 |     let expected3 = convolve_valid(vec,ker);
   |                     ^^^^^^^^^^^^^^ the trait `typenum::private::PrivateSub<typenum::uint::UInt<typenum::uint::UTerm, typenum::bit::B1>>` is not implemented for `typenum::uint::UTerm`
   |
   = note: required because of the requirements on the impl of `typenum::private::PrivateSub<typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UTerm, typenum::bit::B1>, typenum::bit::B0>>` for `typenum::uint::UInt<typypenum::uint::UInt<typenum::uint::UInt<typenum::uint::UTerm, typenum::bit::B1>, typenum::bit::
   = note: required because of the requirements on the impl of `std::ops::Sub<typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UTerm, typenum::bit::B1>, typenum::bit::B0>, typenum::bit::B0>>` for
`typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UTerm, typenum::bit::B1>, typenum::bit::B1>`
   = note: required because of the requirements on the impl of `nalgebra::DimSub<nalgebra::U4>` for `nalgebra::U3`
   = note: required by `nalgebra::convolve_valid`
```

Also there's some issues with Travis not building, not sure why it's happening.

sebcrozet:
> but the issue is an implementation from another crate, when passing in VectorN's.

@npapapietro Could you provide the code that result in this error? The error suggests you are trying to subtract `U4` from `U3`, which is indeed invalid (because that would yield -1). So my guess is that in your example the kernel size is greater than your vector size?

> Also there's some issues with Travis not building, not sure why it's happening.

The errors from Travis look like this:

```
26 |     let vec_d = DVector::from_vec(4,vec![1.0,2.0,3.0,4.0]);
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 1 parameter
```

which means `DVector::from_vec` takes only one parameter. You should remove the first parameter which is is not needed anymore in the latest version of nalgebra.

npapapietro:
Ha! I think that change for DVector::from_vec happened right after my last pull from upstream. 

As for the main error I'm facing.  Convolve with Valid inputs has an output of D1 - D2 + 1. 

```rust

pub fn convolve_valid<N, D1, D2, S1, S2>(
    vector: Vector<N, D1, S1>,
    kernel: Vector<N, D2, S2>,
) -> VectorN<N, DimSum<DimDiff<D1, D2>, U1>>
where
    N: Real,
    D1: DimSub<D2>,
    D2: DimSub<D1, Output = DimDiff<D1, D2>>,
    DimDiff<D1, D2>: DimAdd<U1>,
    S1: Storage<N, D1>,
    S2: Storage<N, D2>,
    DefaultAllocator: Allocator<N, DimSum<DimDiff<D1, D2>, U1>>
{

    let vec = vector.len();
    let ker = kernel.len();

    if vec == 0 || ker == 0 {
        panic!("Convolve's inputs must not be 0-sized. ");
    }

    if ker > vec {
        return convolve_valid(kernel, vector);
    }
    let result_len = vector.data.shape().0.sub(kernel.data.shape().0).add(U1);
    let mut conv = VectorN::zeros_generic(result_len, U1);

    for i in 0..(vec - ker + 1) {
        for j in 0..ker {
            conv[i] += vector[i + j] * kernel[ker - j - 1];
        }
    }
    conv
}

```

Here is how I'm using it:
```rust
    let vec = Vector4::new(1.0,2.0,3.0,4.0);
    let ker = Vector3::new(1.0,2.0,2.1);

    let vec_d = DVector::from_vec(vec![1.0,2.0,3.0,4.0]);
    let ker_d = DVector::from_vec(vec![1.0,2.0]);
    
    let expected_d = convolve_valid(vec_d,ker_d); //Works fine
    let expected_s = convolve_valid(vec,ker); //Throws previously mentioned error

```

sebcrozet:
Oh I see. Then I guess instead of `DimSum<DimDiff<D1, D2>, U1>` you should use `DimDiff<DimSum<D1, U1>, D2>>` to avoid the invalid subtraction. And the:

```rust
    D1: DimSub<D2>,
    D2: DimSub<D1, Output = DimDiff<D1, D2>>,
    DimDiff<D1, D2>: DimAdd<U1>,
```

would become something like (just a guess, I don't have my computer with me rigth now to test this):

```rust
    D1: DimAdd<U1>,
    D2: Dim,
    DimSum<D1, U1>: DimSub<D2>
```

npapapietro:
That works I think I tried that, but it breaks the symmetry of the last function. Meaning if you accidentally pass in a ker.len() > vec.len() it will swap them for you. I'm not entirely sure what the use case for mixing up your kernels(filters) and your target vector is.  If the feeling is that is not needed, I can go ahead and remove them from all 3 functions so that the Convolve methods all work in the same way.

sebcrozet:
> I'm not entirely sure what the use case for mixing up your kernels(filters) and your target vector is. If the feeling is that is not needed, I can go ahead and remove them from all 3 functions so that the Convolve methods all work in the same way.

@npapapietro If the use-case is not clear, then let's not allow this. So I think removing this automatic swap is a option.

npapapietro:
@sebcrozet Sounds good, I think the pull request is ready for more feedback/review.

sebcrozet:
@npapapietro Thank you for addressing my comments. The last point of my first comment was not addressed though: declaring those as inherent methods instead of free functions. Would you mind doing this? This PR will be ready to merge afterward.

npapapietro:
Done

sebcrozet:
Great, thanks @npapapietro!

Issue #487, opened 2018-11-26
Dual Quaternions for Rigid Body Transforms


mxkramer:
[Dual quaternions](https://en.wikipedia.org/wiki/Dual_quaternion) can represent all of rigid body transformations (translation + rotation), with some advantages over matrixes:

- Better numerical stability
- Smaller memory footprint (8 values, two algebraic constraints)
- Small processing speed improvement (one paper I found said ~10% over matrices)

They also keep the simplicity of combining transformations is just multiplication.

A downside is that they only work for 3 dimensions (if I understand correctly), but given a large portion of users likely will be using only 3D geometry (for graphics, etc), it might be worth it. Also, there is already support for non-dual quaternions, so there is some interest.

Many of you probably already know this, just thought I'd document the case for them as I understand it. 

If people are interested in this, I'll create a fork and add the feature? This is my first time contributing to an open-source project on github, so I may need a little guidance... Or, if someone else is interested in writing it that would work as well.

sebcrozet:
Hi! Adding dual quaternions would be a very interesting addition to nalgebra! As far as I know, nobody is working on this right now so your contribution is welcome! I'd be curious to see how dual quaternions compare to the existing `Isometry3` structure which are representations of rigid-body transformations with 7 values and one algebraic constraint (one unit quaternion and one 3D vector).

If you need guidance, I suggest you open a pull-request even if you have not yet finished, and request for reviews when you need them. You should also see how quaternions are currently implemented. In particular, following the same file structure you should add:

- `geometry/dual_quaternions.rs` containing the definition of the dual quaternion `struct` and operations which are not constructors nor operator overloading.
- `geometry/dual_quaternions_ops.rs` containing operator overloading.
- `geometry/dual_quaternion_construction.rs` containing static functions for constructing the dual quaternions.

Of course don't hesitate to ask questions on github, or on the nalgebra [discord chanel](https://discord.gg/vt9DJSW)!

arturoc:
dual quaternions are also really useful to avoid some artifacts while doing mesh skinning compared to using matrices

mxkramer:
Should I work on the master branch or create a new feature branch?
I'm also curious to see how they compare. I know they're used a lot in IK and robotics due to the numerical stability. Maybe after writing them I'll do some benchmarks for speed and stability (since using `Isometry3` has the numerical stability in angle from the quaternion already).

sebcrozet:
> Should I work on the master branch or create a new feature branch?

[...]

PR #544, opened 2019-02-15
Starting work on Issue #487

adamnemecek:
https://github.com/rustsim/nalgebra/issues/487


Nivek92:
What happened to this?

mxkramer:
I checked out the code out of curiosity, it looks like it's not compiling at the moment (but nearly all of the dual quaterinon-related code is commented out anyway). The error is coming from `base/cg.rs` though, so I'm not sure why that is. It's a borrow checker error but I can't recall any changes in borrowck in the last year that would have changed it?

chinedufn:
See #810

PR #545, opened 2019-02-15
Use -> Self and Self:: where possible

adamnemecek:


Ralith:
These probably don't all need to be separate commits.

adamnemecek:
I was doing it little by little as sometimes I made a change that fucked things up and going back way annoying. Should I squash it?

Ralith:
Yes. In the future, you can amend and force-push the same commit repeatedly when iterating on a single logical change like this. For more advanced cases, where multiple commits are necessary and you find yourself making changes out of order, [fixup and autosquash](https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html) are useful tools.

sebcrozet:
Thanks @adamnemecek! A lot of this code date back from before `Self` was allowed at all those places. So what you are doing here is great to modernize all this.

Let us know when you think this is ready to review/merge.

adamnemecek:
I opened a new request where I squashed the commits.

https://github.com/rustsim/nalgebra/pull/546

PR #550, opened 2019-02-19
added Quaternion accessors for r,i,j,k, refactored conjugate to use imag

adamnemecek:


jswrenn:
Quaternion already have field accessors via [the `Deref<Target=IJKW>` impl](https://www.nalgebra.org/rustdoc/nalgebra/geometry/struct.Quaternion.html#impl-Deref).

adamnemecek:
So what exactly does the access look like? Also, it doesn’t seem like the best ergonomics to put this in deref.

jswrenn:
Something like this:
```rust
let q = Quaternion::new(1.0, 2.0, 3.0, 4.0);
assert_eq!(q.w, 1.0)
```

PR #552, opened 2019-02-19
`iter::Sum` for `DMatrix`

### LINKED ISSUE #514: std::iter::Sum for DVector and DMatrix

jswrenn:
Fixes #514. Supersedes #517.

Panic if the iterator is empty.

### DIFF
--- a/src/base/ops.rs
+++ b/src/base/ops.rs
@@ -11,7 +11,7 @@ use base::allocator::{Allocator, SameShapeAllocator, SameShapeC, SameShapeR};
 use base::constraint::{
     AreMultipliable, DimEq, SameNumberOfColumns, SameNumberOfRows, ShapeConstraint,
 };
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thanks!

PR #553, opened 2019-02-19
added geometric algebra operations, wedge & antiwedge product, parall…

adamnemecek:
…el & perpendicular bisectors

Code is taken from https://github.com/peakbook/Quaternions.jl/blob/master/src/Quaternions.jl#L116

### DIFF
--- a/src/geometry/quaternion.rs
+++ b/src/geometry/quaternion.rs
@@ -506,6 +506,30 @@ impl<N: Real> Quaternion<N> {
     pub fn normalize_mut(&mut self) -> N {
         self.coords.normalize_mut()
     }
+
+    /// Calculates the wedge product
+    #[inline]
+    pub fn wedge(&self, other: &Self) -> Self {
+        (self * other - other * self) / ::convert(2.0f64)
+    }
+
+    /// Calculates the antiwedge product
+    #[inline]
+    pub fn antiwedge(&self, other: &Self) -> Self {
+        (self * other + other * self) / ::convert(2.0f64)
+    }
+
+    /// Calculates the parallel bisector
+    #[inline]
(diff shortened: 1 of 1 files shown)


Ralith:
I thought these were operations on/yielding (bi)vectors, not quaternions.

adamnemecek:
Correct but if both operands are Quaternions then the bivector is a quaternion.

Ralith:
Can you cite a more authoritative source for the definitions/interpretations you're using than some javascript library?

adamnemecek:
Which ones? Or all of them? Also it's Julia thank you very much.

sebcrozet:
Thank you for this PR!

> Also, can you take my word for it?

That's not how we work. In order to contribute to this library and add some operations that are not so widely known, you have to either by add clear and comprehensive tests or provide precise bibliographic references (papers/book) instead of linking to the code from another code-base (no matter the language it is originally implemented in) so that we can be able to review your code properly.

> I'm going to contribute some more to this library, and if this is broken, I will fix it.

If you need to implement more features in order to be sure what is added to this PR is not broken, then please add those feature in this same PR so that when we merge it everything is complete and in good shape.

sebcrozet:
@Ralith We discussed this further with @adamnemecek (I'm not familiar with those operations either), and a pretty good description of those operations can be found on the book "Foundations of Game Engine Development, Volume 1: Mathematics" from Eric Lengyel. You can for example see the equation 4.89, page 175.

@adamnemecek Would you mind adding this reference (including the right equation number) in the comments of each operator?

adamnemecek:
Superceded by https://github.com/rustsim/nalgebra/pull/566

PR #559, opened 2019-02-25
Correct typo in quaternion documentation

bwhetherington:
Correct "looses" to "loses" in quaternion.rs documentation.


sebcrozet:
Thanks!

PR #563, opened 2019-03-15
Quaternionic division + refactoring

adamnemecek:


src/geometry/quaternion.rs
```diff
@@ -757,6 +772,15 @@ impl<N: Real> Quaternion<N> {
     }
 }
 
+impl<N: Real> Inv for Quaternion<N> {
```
sebcrozet: Please remove this `impl` and use `.try_inverse` directly on the `left_div` and `right_div` methods above.
If we want to implement more num traits, this should be done on a separate PR that addresses all the nalgebra types instead of just one.

src/geometry/quaternion.rs
```diff
@@ -537,6 +537,29 @@ impl<N: Real> Quaternion<N> {
...
+    /// # use nalgebra::Quaternion;
+    /// let a = Quaternion::new(0.0, 1.0, 2.0, 3.0);
+    /// let b = Quaternion::new(0.0, 5.0, 2.0, 1.0);
+    /// let result = a.left_div(&b).unwrap();
+    /// let expected = Quaternion::new(0.4, 0.13333333333333336, -0.4666666666666667, 0.26666666666666666);
+    /// assert_relative_eq!(expected, result, epsilon = 1.0e-7);
+    /// ```
+    #[inline]
+    pub fn left_div(&self, other: &Self) -> Option<Self> {
+        other.inv().map(|inv| self * inv)
```
sebcrozet: Shouldn't this be `self.inv().map(|inv| inv * other)` instead? I'm guessing "left division" refers to: `a \ b = a⁻¹b` where `a` is `self` and `b` is `other`.

src/geometry/quaternion.rs
```diff
@@ -537,6 +537,29 @@ impl<N: Real> Quaternion<N> {
...
+    }
+
+    /// Right quaternionic division.
+    #[inline]
+    pub fn right_div(&self, other: &Self) -> Option<Self> {
+        other.inv().map(|inv| inv * self)
```
sebcrozet: Shouldn't this be `other.inv().map(|inv| self * inv)` instead? I'm guessing "right division" refers to: `a / b = ab⁻¹` where `a` is `self` and `b` is `other`.

adamnemecek: You mean the names should be flipped?

adamnemecek: https://prnt.sc/myskju

https://www.google.com/books/edition/Pythagorean_Hodograph_Curves_Algebra_and/vrJwyyo0dkcC?hl=en&gbpv=1&dq=%22right+division%22+quaternion&pg=PA66&printsec=frontcover

They define:
Left division = a⁻¹b
Right division = ba⁻¹

However notice that they are diving b by a, if we have a by b then we have

left div = b⁻¹a (or inv * self)
right div = ab⁻¹ (or self * inv)

sebcrozet: I also meant that `self` should always be the first operand of the division. So, in pseudo-code, the right-division is `self * other.inverse()` and the left-division is `self.inverse() * other`.

Also, in your comments, you cannot mention a formula involving `A` and `B` without saying what `A` and `B` are. So you should add "where `A` is `self` and `B` is `other`." to your comments.

sebcrozet: Actually [that ref](https://pursuit.purescript.org/packages/purescript-quaternions/4.0.0/docs/Data.Quaternion#v:leftDiv) has a good reason why we should not do what I suggested, and keep `self.left_div(other)` equal to `other.inverse() * self`. The reason is that, from a abstract algebra point of view, we want the "left division" operator to return the same result as the "rigth division" operator. Thus we want that `self.right_div(other) == self.left_div(other)`.

src/geometry/quaternion.rs
```diff
@@ -836,6 +835,16 @@ impl<N: Real + fmt::Display> fmt::Display for Quaternion<N> {
...
+impl<N: Real + fmt::Debug> fmt::Debug for Quaternion<N> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "Quaternion {:?} − ({:?}, {:?}, {:?})",
+            self[3], self[0], self[1], self[2]
+        )
+    }
+}
+
```
Ralith: This is unidiomatic; `Debug` impls are normally more or less valid Rust syntax for the displayed object. This change is inconsistent with the rest of the lib and doesn't seem to add much value over the derived impl.

adamnemecek: I hate how the dq is displayed without it.

sebcrozet: I agree with @Ralith, this should be removed. We already have a `Display` impl that does exactly that. So all you need to do to have a nicer display is to use the `{}` formater.

PR #567, opened 2019-03-23
Complex number support

### LINKED ISSUE #281: Implement all decompositions for complex types as well.
Currently, matrix decompositions work with reals only. Handling complex numbers would be very useful as well.

### LINKED ISSUE #503: Complex number support
Currently, the `Real` trait pops up 846 times in a search. In the vast majority of those cases, there is no reason why a complex number would not work. Certain operations such as floor/ceil, inequalities, etc. only make sense on real numbers, but the vast majority of functions defined by the `Real` trait also apply to complex numbers. It seems to me that many, many parts of nalgebra rely on `N` being `Real` despite only relying on functions that also make sense for complex numbers. For example, it is impossible to get the norm of a matrix with complex numbers, which is almost a deal breaker for me for the entire library:

https://github.com/rustsim/nalgebra/blob/0f66403cbbe9eeac15cedd8a906c0d6a3d8841f2/src/base/matrix.rs#L1265-L1283

(As a workaround, I am currently just copy-pasting the above code into my project, which works just fine even for complex numbers).

This operation does not require a real number at all, nevertheless all of the numerous requirements of the `Real` trait, including the `'static` lifetime. If I understand correctly, this means that any number put into a matrix will be leaked and never deallocated; and this is just one requirement of `Real`. The same goes for a number of other operations: the `Real` trait is overkill and overly restrictive. The solution to this seems to be mainly a change inside `alga`: separating `Real` into multiple traits and only requiring the needed traits in `impl` blocks. For example, if the traits were separated into `Real`, where `Complex` is the [supertrait](https://doc.rust-lang.org/stable/book/ch19-03-advanced-traits.html#using-supertraits-to-require-one-traits-functionality-within-another-trait) of `Real`, then taking the norm of a matrix would be constrained to `Complex`, allowing for both real and complex numbers. This would seem to fix #281 and a number of other problems with complex numbers.

This is no small task as it would require manually going through much of the library and updating trait constraints, but I believe that it is an important and necessary change. I think that this should at least be an RFC, both for whether making this change is important or not, and what exactly the new traits would be.

**EDIT**: That was a bad example, I just saw the `rusty_machine` branch and c531a34 in #499 😄. However this is still an issue until that is merged, and for many other algorithms and operations. But don't let this issue block #499 because this might require some discussion and work to implement for any changes that are agreed on.
**EDIT 2**: Actually, it's still not fixed on the `rusty_machine` branch, *and* now for some reason copy-pasting the code doesn't work with that branch because the dot product now requires the `ClosedAdd` and `ClosedMul` traits, which `num_complex::Complex` doesn't implement.

sebcrozet:
This adds complex number support almost everywhere a `Real` was accepted before. This includes decomposition of complex matrices as well as BLAS operations involving conjugate-transposition.

This relies on ongoing work on the alga  crate on the [complex branch](https://github.com/rustsim/alga/tree/complex).

Fix #281 
Fix #503


Coder-256:
Wow this is an enormous PR, I totally underestimated the magnitude of this change. Thank you for all this, it will definitely be a huge help in my project and surely many others!

That being said, I have just a few questions. For many of these functions that now use `Complex`, wouldn't that be a breaking change since people now have to wrap everything with `Complex::new(n, 0)`? Or are there different functions that take complex numbers? (I'm finding it hard to figure this out myself through all the code; especially because the documentation won't build). Additionally, I am concerned about the implications of this on memory allocation if now you need to allocate double the memory for unused fields. On the topic of type safety, what happens to the return type functions that return real values for real inputs and complex values for complex inputs, such as the trigonometric functions? (or even simple linear functions; in fact probably most functions). For example, I want `sin(a)` to return a real value rather than having to assert that a complex return has no imaginary part. Lastly, it is really hard to read this behemoth of a diff even when cherry-picking, so would it be possibly to very briefly sum up what has changed?

Again, I really appreciate the open-source work that went into this project and it will definitely be useful no matter what! 😄

Ralith:
> For many of these functions that now use Complex, wouldn't that be a breaking change since people now have to wrap everything with Complex::new(n, 0)?

`Complex` is a trait, not a struct, and per https://github.com/rustsim/alga/commit/05763823b73c7161b07aec588fc758a97f6b9289 every type that implements `Real` (e.g. `f32`) also implements `Complex`.

Coder-256:
@Ralith Thank you, I see you're right, I did not realize there was a new trait because I was mixing it up with the `Complex` struct from `num_complex`. That is a clever way to implement things. However that now adds even more confusion to my question about return types.

Ralith:
I think everything's continuing to return the same type it's given as input, so e.g. if you want to get complex results for real inputs you'll have to lift your reals into a complex type with an explicit zero imaginary component first yourself.

Coder-256:
@Ralith What I mean is, if we now change the `sin` function to accept any type conforming to `Complex`, then what is the return type? Is it the type that I pass it? In that case, what about functions that actually need to return a complex number regardless of whether the input is real or not? (ex. finding the square root of a negative real)

Ralith:
> Is it the type that I pass it?

I think so.

> In that case, what about functions that actually need to return a complex number regardless of whether the input is real or not?

For types that cannot represent the imaginary component, they'll probably do the same thing they do today. If you don't want that behavior, use a type that can represent the imaginary component.

sebcrozet:
@Coder-256 I confirm what @Ralith said. Everything will work just like before for anything that are not complex numbers. If you use `f64`, you will get operations on real matrices and vectors that return the same float-based results as before. The novelty here is that if you want to use `num_complex::Complex`, you can, and it will perform all the operations on the complex field.

> What I mean is, if we now change the sin function to accept any type conforming to Complex, then what is the return type? Is it the type that I pass it?

If you have a function that takes a type `N: Complex`, and outputs `N`, then the return type is the same as the input type. If you have `N = f64`, it will output a `f64`. If `N = num_complex::Complex<f64>`, it will output a `num_complex::Complex<f64>`. One particularity though is that some operations, like the modulus, take a complex number and return a real number. In that case the signature looks like this: `modulus<N: Complex>(val: N) -> N::Real` where `::Real` is an associated type provided by the new `Complex` trait.

> In that case, what about functions that actually need to return a complex number regardless of whether the input is real or not? (ex. finding the square root of a negative real)

The square root of a negative real does not exist if you are not working on the complex field. So for this square root to exist, you need to provide a complex number (with an imaginary part equal to zero) in the first place.

I understand having a trait named `Complex` can be quite confusing. Would you @Coder-256 or @Ralith, or perhaps @jswrenn have a better name idea?

jswrenn:
Name collisions between different numeric libraries are probably inevitable. I think `Complex` is an appropriate name. I think referring to it in the docs as `alga::Complex` (rather than just `Complex`) would clear up most of the confusion.

(This would probably be a good thing to do with other traits that come from `alga`, too.)

sebcrozet:
@jswrenn Thank you for your input.
@WaDelma You contributed a lot to `alga` too, do you have an opinion about whether or not we should name `Complex` the alga trait for complex numbers or if another name would be more suitable?

Coder-256:
@sebcrozet IMHO naming the trait `Complex` is a bit confusing. I initially thought that it was meant only for complex numbers. How about `ComplexField`? That implies that it accepts any numeric type in the complex field (C). Maybe `RealField` too for consistency. (just an idea)

sebcrozet:
@Coder-256 Thank you for the suggestions, I think those are actually great names!

WaDelma:
`ComplexField` sounds good to me.

adamnemecek:
U go gurl!!!

PR #568, opened 2019-03-23
Fix matrix slerp function

IcanDivideBy0:
I'm currently using the slerp function to interpolate between frames in a skeletal animation,

Here is the actual result:
bugged

Comparing the nalgebra slerp with the gl-matrix slerp (which worked fine on the same animation), I noticed [some differences](https://github.com/toji/gl-matrix/blob/master/src/quat.js#L210-L216) with the nalgebra  implementation.

Same animation after the fix:
working

Let me know if you'd like some test around this case

### DIFF
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -1480,6 +1480,19 @@ impl<N: Scalar + Zero + One + ClosedAdd + ClosedSub + ClosedMul, D: Dim, S: Stor
 
 impl<N: Real, D: Dim, S: Storage<N, D>> Unit<Vector<N, D, S>> {
     /// Computes the spherical linear interpolation between two unit vectors.
+    ///
+    /// # Examples:
+    ///
+    /// ```
+    /// # use nalgebra::geometry::UnitQuaternion;
+    ///
+    /// let q1 = UnitQuaternion::from_euler_angles(std::f32::consts::FRAC_PI_4, 0.0, 0.0);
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thank you for this fix!

> Let me know if you'd like some test around this case

If you can find a simple test for this, then please add it, preferably as a doc-test.

IcanDivideBy0:
Any quaternion pair that have a negative cross product will fall into the bug.
I tried to write the test, but i can't manage to get the rustdoc command working 

```sh
rustdoc --test src/base/matrix.rs
```

gives me tons of errors about `unresolved import nalgebra` with both the stable and nightly branch. I'm not really familiar with this tool, am I missing something obvious?

sebcrozet:
@IcanDivideBy0 You need to use `cargo` to run the tests, otherwise it will only see the one rust file without considering the whole crate. `cargo test --doc slerp` will run every doctests containing `slerp` on its name.

IcanDivideBy0:
Ok, the test is done. The second quaternion could have been  just `PI` and not `-PI` with the same result, but, having it negative have two purposes:
- first the dot product will be negative, which is the core reason of this fix
- secondly, it shows the `slerp` function will interpolate with the shortest path on the sphere surface, not just raw angles values

sebcrozet:
Nice! Thanks a bunch. This will be released with nalgebra 0.18 at the end of next week.

Issue #543, opened 2019-02-12
Using atan2 instead of acos for rotation angles?


mpizenberg:
![screenshot from 2019-02-12 15-58-18](https://user-images.githubusercontent.com/2905865/52669257-e2f47100-2f15-11e9-8eef-c141760e91c7.png)

I've recently faced a case where I wanted to compare very similar 3D rotations (UnitQuaternion). One considered as "ground truth", the other obtained by computations from data. To measure the quality of the computed one, I used the `angle_to()` method, itself based on the `angle()` method. This method was giving me a finite number of different angles, as can be seen in the left plot of the figure, which was weird.

As explained in the [wikipedia page of quaternions for rotations][wiki-quat], a quaternion basically is an axis-angle representation, `q = cos(a/2) + sin(a/2) * v` where `v` is the imaginary part (3 components). Currently nalgebra retrieve the angle through the real part by computing `angle = 2.0 * acos(|w|)`. However, `acos` appears to be less precise for small angles, i.e. when `w` tends to 1. So I replaced the `angle()` method by the following `angle` function in my code, using the arctangent instead, and obtained the right plot of the figure.

```rust
fn angle(uq: UnitQuaternion<f32>) -> f32 {
    let w = uq.into_inner().scalar();
    2.0 * uq.into_inner().vector().norm().atan2(w)
}
```

(Note that I didn't use `abs` here since I didn't really care about the sign for my case). It appears that the arctangent behaves a lot better for small angles than arccos. When I have time, I'll try to make proper tests that could confirm this for nalgebra. I could also do a PR if this doesn't break stuff.

[wiki-quat]: https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation

simonpuchert:
This is indeed a problem with `acos` (and finite precision).
The values you get are all of the form `acos(1 - n * 2^-22)` with `n=0,1,2,...`, consistent with what one would expect from `f32`.

Your `atan2`-based implementation does not need `abs`, since `y.atan2(x)` will be in [0, pi] as long as y is non-negative (and norms are non-negative).
This certainly looks like a better solution than the existing one in principle, although I can't comment on the implementation details; there are seemingly a lot of different ways to extract the "scalar"/"vector" parts of a quaternion.

sebcrozet:
Fixed by #569.

PR #569, opened 2019-03-25
Using atan2 instead of acos for rotation angles.

simonpuchert:
I have implemented the solution discussed in #543.


sebcrozet:
Thanks!

mpizenberg:
@simonpuchert As you mentioned last time, we should not use `abs(w)` but `w` to get the full range of angles between 0 and pi. In this PR you left the abs from previously. Wouldn't this cause issues for angles > pi/2 ?

PR #574, opened 2019-04-02
Add some infinite and reversed perspectives.

### LINKED ISSUE #573: Projection: implement reversed-Z and infinite zFar
Reversed-Z and infinite zFar are two techniques that improves depth precision, mitigating issues like Z-fighting. The implementation is described [here](http://dev.theomader.com/depth-precision/).

These approaches are gradually gaining popularity as default configuration in game engines. Especially, infinite zFar makes things simpler and should be the go-to configuration to be used.

Reversed-Z has a different Z representation: thus it's backward incompatible and needs a different API. Infinite zFar should be implemented as an addition to the API.

sebcrozet:
This adds the following to nalgebra-glm:

- infinite_perspective_rh_no
- infinite_perspective_rh_zo
- ~reversed_perspective_rh_no~
- reversed_perspective_rh_zo
- ~reversed_infinite_perspective_rh_no~
- reversed_infinite_perspective_rh_zo

Addresses #573

### DIFF
--- a/nalgebra-glm/src/ext/matrix_clip_space.rs
+++ b/nalgebra-glm/src/ext/matrix_clip_space.rs
@@ -43,7 +43,7 @@ use na::{RealField};
 //
 //pub fn infinite_perspective_lh<N: RealField>(fovy: N, aspect: N, near: N) -> TMat4<N> {
 //    unimplemented!()
-//}ç
+//}
 //
 //pub fn infinite_ortho<N: RealField>(left: N, right: N, bottom: N, top: N) -> TMat4<N> {
 //    unimplemented!()
@@ -735,7 +735,7 @@ pub fn infinite_perspective_rh_zo<N: RealField>(aspect: N, fovy: N, near: N) -> 
     mat
 }
 
-/// Creates a matrix for a right hand perspective-view frustum with a depth range of -1 to 1.
+/// Creates a matrix for a right hand perspective-view frustum with a reversed depth range of -1 to 1.
 ///
 /// Note that when using reversed perspective, it is best to use a depth buffer based on floating points.
 ///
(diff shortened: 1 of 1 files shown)


Ralith:
Reviewing [the article](http://dev.theomader.com/depth-precision/), it looks like I misremembered: reversed Z actually works fine on integer depth buffers, it's just that *forward* Z is terrible on *floating-point* depth buffers.

ishitatsuyuki:
You should *not* use reversed Z with OpenGL ([-1, 1]) range.

> This doesn't make a difference for integer formats, but with floating-point, all the precision is stuck uselessly in the middle. (The value gets mapped into [0, 1] for storage in the depth buffer later, but that doesn't help, since the initial mapping to [-1, 1] has already destroyed all the precision in the far half of the range.) And by symmetry, the reversed-Z trick will not do anything here. 
> *[Depth Precision Visualized - NVIDIA Developer](https://developer.nvidia.com/content/depth-precision-visualized)*

Also, do you mind adding functions for the `nalgebra` types too? It doesn't have to be the same PR, but in that case I think you should leave the issue open.

nalgebra-glm/src/ext/matrix_clip_space.rs
```diff
@@ -691,10 +687,110 @@ pub fn perspective_zo<N: RealField>(aspect: N, fovy: N, near: N, far: N) -> TMat
     perspective_rh_zo(aspect, fovy, near, far)
 }
 
+/// Build infinite right-handed perspective projection matrix with [-1,1] depth range.
+///
+/// # Parameters
+///
+/// * `fovy` - Field of view, in radians
+/// * `aspect` - Ratio of viewport width to height (width/height)
+/// * `near` - Distance from the viewer to the near clipping plane.
```
Ralith: Could these (and similar below) be listed in the order they're passed to the function?

Ralith:
Would it make sense for the documentation to specifically encourage people to use `reversed_infinite_perspective_rh_zo`? I think it's pretty clearly the sane default option, so long as you're able to set up your depth range and depth test appropriately.

sebcrozet:
Thank you @ishitatsuyuki and @Ralith for your reviews!
@Ralith I've not modified the documentation to encourage the use of reversed perspective as this has implications on the depth-range and depth test settings that may not be what you find on common tutorials. We could add this kind of advice to the user guide though.

PR #576, opened 2019-04-06
Change R (row), with C (column) in alias

lans98:


sebcrozet:
Hi!

Could you explain why this is needed? By default, the elements `m[(i, j)]` and `m[(i, j + 1)]` should be separated by `R` (the number of rows) components for a column-major matrix.

PR #578, opened 2019-04-08
Allow the removal of multiple rows/columns given an array of indices.…

### LINKED ISSUE #530: Allow the removal of multiple rows/columns given an array of indices.
See [that discussion](https://discourse.nphysics.org/t/how-to-dynamically-remove-rows-with-a-vec-of-idexes/314) on the user forum.

The goal is to add to the `base/edition.rs` file the methods `remove_rows_at(self, indices: &[usize])` and `remove_columns_at(self, indices: &[usize])`. They will remove all rows or columns at the indices provided in the slice.

KappaDistributive:
fixes #530

The initial plan was to iteratively remove rows/columns one by one. That didn't work out due to the way nalgebra handles types. 
I therefore opted for manipulating the data directly which has the downside of introducing even more unsafe (and less readable) code but the upside of a better runtime (O(m x n) for a m x n matrix).

### DIFF
--- a/src/base/edition.rs
+++ b/src/base/edition.rs
@@ -308,7 +308,7 @@ impl<N: Scalar, R: Dim, C: Dim, S: Storage<N, R, C>> Matrix<N, R, C, S> {
     }
 
     /// Removes all columns in `indices`   
-    #[inline]
+    #[cfg(any(feature = "std", feature = "alloc"))]
     pub fn remove_columns_at(self, indices: &[usize]) -> MatrixMN<N, R, Dynamic>
     where
         C: DimSub<Dynamic, Output = Dynamic>,
@@ -346,7 +346,7 @@ impl<N: Scalar, R: Dim, C: Dim, S: Storage<N, R, C>> Matrix<N, R, C, S> {
     }
 
     /// Removes all columns in `indices`   
-    #[inline]
+    #[cfg(any(feature = "std", feature = "alloc"))]
     pub fn remove_rows_at(self, indices: &[usize]) -> MatrixMN<N, Dynamic, C>
     where
         R: DimSub<Dynamic, Output = Dynamic>,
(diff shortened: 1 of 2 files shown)


KappaDistributive:
I'd like some help with the errors Trevor CI was able to detect: Why exactly is the Dynamic type not declared in that setting?

src/base/edition.rs
```diff
@@ -303,6 +307,79 @@ impl<N: Scalar, R: Dim, C: Dim, S: Storage<N, R, C>> Matrix<N, R, C, S> {
         self.remove_fixed_columns::<U1>(i)
     }
 
+    /// Removes all columns in `indices`   
+    #[inline]
+    pub fn remove_columns_at(self, indices: &[usize]) -> MatrixMN<N, R, Dynamic>
```
sebcrozet:
```suggestion
    #[cfg(any(feature = "std", feature = "alloc"))]
    pub fn remove_columns_at(self, indices: &[usize]) -> MatrixMN<N, R, Dynamic>
```

src/base/edition.rs
```diff
@@ -303,6 +307,79 @@ impl<N: Scalar, R: Dim, C: Dim, S: Storage<N, R, C>> Matrix<N, R, C, S> {
...
+                nrows,
+                ncols.sub(Dynamic::from_usize(v.len())),
+                m.data,
+            ))
+        }
+    }
+
+    /// Removes all columns in `indices`   
+    #[inline]
+    pub fn remove_rows_at(self, indices: &[usize]) -> MatrixMN<N, Dynamic, C>
```
sebcrozet:
```suggestion
    #[cfg(any(feature = "std", feature = "alloc"))]
    pub fn remove_rows_at(self, indices: &[usize]) -> MatrixMN<N, Dynamic, C>
```

sebcrozet:
Thanks!

PR #579, opened 2019-04-09
MatrixSliceMutMN: prevent aliased indices.

### LINKED ISSUE #486: Strides may introduce multiple, aliased mutable pointers.
For instance:
```rust
use std::ptr;
use std::ops::IndexMut;

let data = &mut [1, 2];

let mut matrix =
  MatrixSliceMut1x2::from_slice_with_strides_mut(data, 0, 0);

assert_eq!(matrix[(0, 0)], matrix[(0, 1)]);
```
The code above is not, on its own, unsafe, but nalgebra relies on many functions with the invariant that strides will not result in aliased elements (e.g., [`matrixmultiply::dgemm`](https://docs.rs/matrixmultiply/0.2.0/matrixmultiply/fn.dgemm.html)).

This problem was noted in #473, but I'm creating a separate issue for visibility since this is a memory safety issue.

jswrenn:
Fixes #486.

The assertion is structured in a way that rules out false-positives. Each branch of the match arm ends in an assertion in the form:
```rust
index(i₀, j₀) != index(i₁, j₁)
```
If this expression evaluates to false, then `(i₀, j₀)` and `(i₁, j₁)` are, necessarily, examples of indices that will alias to the same linear index.

At a glance, it's a little less obvious that this assertion won't produce false-negatives. (This isn't as alarming as the possibility of a false-positive, since the current false-negative rate is 100%.) [The result of the assertion exactly matches the results of an exhaustive search for all combinations of rows, columns, row strides, and column strides ranging from 0 to 10.](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=7456bc83f93ed6e9bbf1daf27f27e672)

<small>Thanks to [@hwayne](https://github.com/hwayne) for the insight that in the non-trivial case, the aliased indices relate to the gcd of row and column stride!</small>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ generic-array  = "0.12"
 rand           = { version = "0.6", default-features = false }
 num-traits     = { version = "0.2", default-features = false }
 num-complex    = { version = "0.2", default-features = false }
+num-rational   = { version = "0.2", default-features = false }
 approx         = { version = "0.3", default-features = false }
 alga           = { version = "0.9", default-features = false }
 matrixmultiply = { version = "0.2", optional = true }
--- a/src/base/construction_slice.rs
+++ b/src/base/construction_slice.rs
@@ -114,26 +114,14 @@ impl<'a, N: Scalar, R: Dim, C: Dim, RStride: Dim, CStride: Dim>
             let rstride = rstride.value();
             let cstride = cstride.value();
 
-            // `Storage::linear_index`
-            let index = |i, j| (i * rstride) + (j * cstride);
-
-            // The final condition of each arm is an expression in the form:
-            //    index(i₀, j₀) != index(i₁, j₁)
-            // If this expression is `false`, then the values (i₀, j₀)
-            // and (i₁, j₁) are concrete examples of indices that would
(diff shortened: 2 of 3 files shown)


sebcrozet:
Thank you for addressing this!

It seems the `index(i₀, j₀) != index(i₁, j₁)` conditions will always evaluate to `false` here. So could those be commented out (to keep them as examples of indices that would alias) to avoid needless operations at runtime?

```rust
match (rstride, cstride) {
    (0, 0) => false, // otherwise the index (0, 0) aliases with (nrows - 1, ncols - 1)
    (0, _) => nrows <= 1, // otherwise the index (0, 0) aliases with (nrows - 1, 0)
    (_, 0) => ncols <= 1, // otherwise the index (0, 0) aliases with (0, ncols - 1)
    (_, _) => {
        let ratio = Ratio::new(rstride, cstride);
        let numer = *ratio.numer();
        let denom = *ratio.denom();

        nrows <= denom || ncols <= numer // otherwise  the index (0, numer) aliases with (denom, 0)
}
```

sebcrozet:
Thanks!

PR #582, opened 2019-04-12
import refactoring

adamnemecek:


sebcrozet:
Thank you for this PR but I will close it.

First I think we should let rustfmt do this kind of things. We already have a `rustfmt.toml` with some extra options enabled. We might want to add `merge_imports = true` too.

Also the `use {` is something I've rarely see on other rust codebases so I'd prefer to avoid it. Merged imports look interesting though.

PR #584, opened 2019-04-13
Derive Copy, Eq, and Hash for Quaternion.

### LINKED ISSUE #583: Quaternion's Hash and Eq are incompatible
`Quaternion` implements `Eq`, but it's `eq` implementation is incompatible with it's `Hash::hash` implementation.

The `eq` implementation considers `w + xi + yj + zk` and `-w - xi - yj - zk` equal because they represent the same rotation, whereas the `hash` implementation does not.

I think the mistake is the `eq` implementation. `-1` is not equal to `1`. These are two *different* quaternions, that just happen to *represent the same rotation*, if we use them for that. I think `eq` should be "equal quaternion" instead of "equal rotation". (Just like `Real::two_pi()` should not compare equal to `0.0`, even though they represent the same rotation.)

m-ou-se:
They were implemented manually.

Eq was implemented in such a way that -q was considered equal to q,
probably because they represent the same rotation. They are, however,
not equal quaternions. Because of this error, the implementation of Hash
was not compatible with the implementation of Eq. This fixes that.

Fixes #583.

### DIFF
--- a/src/geometry/quaternion.rs
+++ b/src/geometry/quaternion.rs
@@ -968,6 +968,28 @@ impl<N: RealField> UnitQuaternion<N> {
         self.conjugate()
     }
 
+    /// Check if the quaternion represents the same rotation as another.
+    ///
+    /// Returns true when `q1 == q2`, or when `q1 == -q2`, to account for the
+    /// double-covering of S², i.e. q = -q.
+    ///
+    /// # Example
+    /// ```
+    /// # use nalgebra::{Unit, UnitQuaternion, Quaternion};
+    /// let q1 = Unit::new_normalize(Quaternion::new(0.6, 0.8, 0.0, 0.0));
+    /// let q2 = Unit::new_normalize(Quaternion::new(-0.6, -0.8, 0.0, 0.0));
+    /// assert!(q1.eq_rotation(&q1));
+    /// assert!(q1.eq_rotation(&q2));
+    /// assert!(q2.eq_rotation(&q1));
+    /// assert!(!q1.eq_rotation(&UnitQuaternion::identity()));
+    /// ```
+    #[inline]
+    pub fn eq_rotation(&self, other: &Self) -> bool {
+        self.coords == other.coords ||
+        self.coords.iter().zip(&other.coords).all(|(a, b)| *a == -*b)
+    }
+
+
     /// The rotation angle needed to make `self` and `other` coincide.
     ///
     /// # Example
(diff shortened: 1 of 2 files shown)


sebcrozet:
Thank you for this PR! I will review it in depth next Saturday.

sebcrozet:
The CI fails because we actually have a test for `q = -q` [there](https://github.com/rustsim/nalgebra/blob/dev/tests/geometry/quaternion.rs#L111-L113).

Would you mind:
1. Adding the method `.is_same_rotation_as(other)` I mentioned in the issue #583 (the method name can be something else if you think of something better).
2. Replacing the `mq == q` check by `q.is_same_rotation_as(&mq)` in the test that currently fails?

m-ou-se:
Thanks for reviewing so quickly!

No problem, I'll make those changes tomorrow.

m-ou-se:
I also noticed that `AbsDiffEq`, `RelativeEq`, and `UlpsEq` were implemented to consider `q = -q`. I have changed those as well to no longer do that. Should that functionality be restored through some other function? Or another type?

m-ou-se:
Ah, looks like I should run `cargo test --all-features`, not just `cargo test`. */me goes back to fixing more tests.*

sebcrozet:
I'm starting to feel like our mistake in the first place was to not create a wrapper type like `struct RotationQuaternion<N>(Unit<Quaternion<N>>)` that would indeed assume that its internal `q` is the same as its `-q`.

There are at least two other benefits of doing that:
1.  It will be more obvious for newcomers to understand that `RotationQuaternion` is a rotation represented as a quaternion. (We could also rename the current `Rotation` type to `RotationMatrix` to be explicit.)
2. It will prevent a tricky mistake. Currently if you write `-q * v` where `q` is a unit-quaternion and `v` a vector, it will compile and be equal to `q * v` which is misleading considering one could expect the result to be `-(q * v)`. With a wrapper type that does not implement `Neg`, the `-q * v` expression will simply not compile.

 One downside is that this will be a breaking change.

Would you like to make this change @m-ou-se ? Here is a list:

1. Replace the `type UnitQuaternion<N> = Unit<Quaternion<N>>` by `struct RotationQuaternion<N>(pub Unit<Quaternion<N>>)`.
2. Replace `UnitQuaternion` by `RotationQuaternion` everywhere.
3. On all methods of the new `RotationQuaternion` the `self.` will have to be replaced by `self.0` to account for the struct wrapping.
4. Add a custom implementation of `Hash` that returns the same hash for both `self.0` and `-self.0`.
5. Add a `impl From<Unit<Quaternion>> for RotationQuaternion<N>`.

We should do something similar for `Unit<Complex>` but that will be the purpose of another issue.

m-ou-se:
Yes, that does feel like the best option to avoid a bunch of issues.

I'll work on that change. I'll close this PR and open a new issue to track it.

PR #585, opened 2019-04-16
Implement Default for Matrices and Quaternions

m-ou-se:


sebcrozet:
Thank you for this PR. However I'll close it because it is not clear what the `Default` trait should return. For examples, for matrices, should it return the zero matrix or the identity matrix? For quaternions should it return the identity quaternion or the zero quaternion? Both choices can make sense in different contexts so it's better to let the user choose explicitly with `::one(), ::identity()`, or `::zero(), ::zeros()`.

This is related to https://github.com/rustsim/nalgebra/issues/201.

m-ou-se:
Thanks for your quick reply.

I was on the edge about this, since it's not directly clear if `Default` should give the additive identity (`::zeros()`) or the multiplicative identity (`::identity()`). But then I noticed that `f64` has a `Default` implementation that gives `0.0` (the additive identity), and not `1.0` (the multiplicative identity). So I figured following the standard library would make most sense here.

m-ou-se:
(The main reason I want this, is so I can easily `#[derive(Default)]` for structs containing vectors, just like I can for structs containing f64s. Otherwise I'll have to make a custom proc macro for `#[derive(Zero)]` or something.)

sebcrozet:
Thank you for clarifying the use-case here! I'm surprised the standard library has this implementation for `f64` though it kind of makes sense for ergonomic reasons. I'll reopen this PR for further consideration.

Andlon:
Please consider closing this PR again. I am also surprised that the standard library has a `Default` implementation for `f64`, but in my eyes this is a mistake, and I would rather avoid `nalgebra` making the same mistake. I completely agree with @sebcrozet's original stance, that there is *no* obvious `Default` for matrices/vectors.

mgeier:
Please correct me if I'm wrong, I'm a Rust beginner!

As far as I understand, the whole `Default` trait is less about defining a "special" or "ideal" value of the given type, but rather about having some arbitrary (but deterministic and unique and known) value as a pragmatic solution.

The [docs](https://doc.rust-lang.org/std/default/trait.Default.html) tell us:

> Sometimes, you want to fall back to some kind of default value, and don't particularly care what it is.

This doesn't imply that the default value should have some special properties like being the additive or multiplicative identity or something like that.

I guess this is a case of "practicality beats purity".

Given the implementors of `Default` in the standard library, it looks like preferably some kind of "zero" or "null" or "empty" value is used as default.

I guess for matrices (with a given size) and quaternions this means a value of all zeros.

For `UnitQuaternion` all zeros is not possible, so I guess `identity()` is the next best choice (as @m-ou-se has used in this PR).

@Andlon 

> I am also surprised that the standard library has a `Default` implementation for `f64`, but in my eyes this is a mistake, [...]

If you argue like that, the outcome would be that none of the standard library types has an unquestionable default value, therefore the whole `Default` trait should be removed from Rust.

But if you accept that Rust has the `Default` trait, you should IMHO also accept that `nalgebra` should implement it.

wdanilo:
Could we at least add `Default` impl for Vectors? It is rather unambiguous that vectors should be zero as default, right?

sebcrozet:
Thank you for your additional feedbacks @Andlon, @mgeier, and @wdanilo. It is true that the standard library explicitly states that `Default` is for cases where you don't particularly care what the default value is, so we are free to use whatever value we want. If the user needs additional semantics, it should use functions like `::zero()`, `::one()`, `::identity()` insted. Moreover, since the stdlib prefers the use of zero as the default, we should do the same for every type that can be zero, including matrices.

So I will merge this PR after all since it is in line with what the stdlib recommends, and it is often used for some plumbing in crates that need to initialize some values to something not undefined. Thanks @m-ou-se!

wdanilo:
@sebcrozet thank you thank you thank you!

PR #586, opened 2019-04-17
Fix incorrect documentation from `s/Real/&Field`

Laaas:
I assume that you did a sed over the entire project, forgetting about the fact that "Real" is also used within the documentation. I haven't looked for other uses of "Real" other than this in the documentation.


sebcrozet:
Thanks!

PR #589, opened 2019-04-22
Fix for nostd builds

lassade:
1. Downgraded rand for version 0.5 to avoid compilations errors with rand_jitter crate;
2. Changed std::mem to core::mem, to allow glm to compile with nostd. (Maybe it will need a cfg to switch between the two)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ io = [ "pest", "pest_derive" ]
 [dependencies]
 typenum        = "1.10"
 generic-array  = "0.12"
-rand           = { version = "0.6", default-features = false }
+rand           = { version = "0.5", default-features = false }
 num-traits     = { version = "0.2", default-features = false }
 num-complex    = { version = "0.2", default-features = false }
 num-rational   = { version = "0.2", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ io = [ "pest", "pest_derive" ]
 [dependencies]
 typenum        = "1.10"
 generic-array  = "0.12"
-rand           = { version = "0.5", default-features = false }
+rand           = { version = "0.6", default-features = false }
 num-traits     = { version = "0.2", default-features = false }
 num-complex    = { version = "0.2", default-features = false }
 num-rational   = { version = "0.2", default-features = false }
--- a/nalgebra-glm/src/common.rs
+++ b/nalgebra-glm/src/common.rs
@@ -1,6 +1,6 @@
 use na::{self, DefaultAllocator, RealField};
 use num::FromPrimitive;
-use std::mem;
+use core::mem;
 
 use crate::aliases::{TMat, TVec};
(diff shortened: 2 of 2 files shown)


sebcrozet:
Hi! Thank you for this PR. Could you provide details on how I could reproduce the compilation failure when targeting WASM?

lassade:
My .toml settings are:
```
...
# I was using version 0.4 if I am not mistaken
nalgebra-glm = { path = "./lib/nalgebra/nalgebra-glm", default-features = false }
```

With rand 0.6 I got this error:
```
...
   Compiling rand_jitter v0.1.3
error[E0277]: the trait bound `error::TimerError: std::error::Error` is not satisfied
  --> /home/lassade/.cargo/registry/src/github.com-1ecc6299db9ec823/rand_jitter-0.1.3/src/error.rs:62:9
   |
62 |         Error::with_cause(ErrorKind::Unavailable,
   |         ^^^^^^^^^^^^^^^^^ the trait `std::error::Error` is not implemented for `error::TimerError`
   |
   = note: required because of the requirements on the impl of `core::convert::From<error::TimerError>` for `alloc::boxed::Box<(dyn std::error::Error + core::marker::Send + core::marker::Sync + 'static)>`
   = note: required because of the requirements on the impl of `core::convert::Into<alloc::boxed::Box<(dyn std::error::Error + core::marker::Send + core::marker::Sync + 'static)>>` for `error::TimerError`
   = note: required by `rand_core::Error::with_cause`
```

After fixing it by downgrading to rand 0.5, this error will show up:
```
   Compiling nalgebra-glm v0.4.0 (/home/lassade/wasm/start/crate/lib/nalgebra/nalgebra-glm)
error[E0432]: unresolved import `std`
 --> lib/nalgebra/nalgebra-glm/src/common.rs:3:5
  |
3 | use std::mem;
  |     ^^^ did you mean `nalgebra::std`?

error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: Could not compile `nalgebra-glm`.
```

sebcrozet:
Thank you. I can reproduce the error related to `std::mem` when compiling nalgebra-glm 0.4 for `no-std` with:

```sh
xargo build --no-default-features --target=x86_64-unknown-linux-gnu
```

However I still can't reproduce the compilation error when targeting WASM with:

```sh
cargo build --target wasm32-unknown-unknown
```

So my guess is that the problem may be due to something else in your `Cargo.toml` (perhaps another indirect dependency to `rand` that enables some its features?) Could you please provide:

1. The complete `Cargo.toml`.
2. The exact command-line you used to compile targeting WASM.

lassade:
Sorry, there is no problem building for wasm, just nostd give the errors.
I was trying to my create nostd to reduce the wasm size (that was for no avail).

sebcrozet:
@lassade Could you revert the downgrade of rand please? That way I could merge you fix related to  `std::mem`.

sebcrozet:
Thanks!

PR #599, opened 2019-06-06
Update lapack-src dependency.

waywardmonkeys:
This was just published and updates its own dependencies which
removes some dependencies on old versions of rand, rust-crypto,
and other things.


sebcrozet:
Thanks! Closing this as the update has been made since then.

PR #601, opened 2019-06-06
Update generic-array dep.

waywardmonkeys:
This update requires Rust 1.30.1 or later.


sebcrozet:
Thanks! Closing this as the update have been made since now.

PR #608, opened 2019-06-11
HTTPS All The Things

Jake-Shadle:
When deploying some locally generated documentation to netlify, netlify was "helpfully" spewing out warnings such as this

```
6:05:35 PM: Mixed content detected in: /nalgebra/base/allocator/type.sameshapec.html
6:05:35 PM: --> insecure link urls:
6:05:35 PM:   - http://nalgebra.org/img/favicon.ico
```

And indeed, the [nalegbra site](https://www.nalgebra.org/rustdoc/nalgebra/index.html) itself has a similar warning about insecure content, at least in Firefox.

image

image

So this change just basically turns all of the http:// links to https:// including for some links to external sites in some of the comments, I think there was actually only 1 site that didn't have valid HTTPS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 All notable changes to `nalgebra`, starting with the version 0.6.0 will be
 documented here.
 
-This project adheres to [Semantic Versioning](http://semver.org/).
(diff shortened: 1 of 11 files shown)


sebcrozet:
Thanks!

PR #612, opened 2019-06-24
QR factorization with column pivoting

russellb23:
Added QR factorization with column pivoting as a separate module.


russellb23:
Local build has no issues with cholesky decomposition which fails travis build

russellb23:
unpack fails. Revoking the PR to fix that and for adding tests

PR #615, opened 2019-07-04
Add methods to get transposed views of matrices

mad-s:
It works by flipping the shapes and strides. This should be much cheaper than copying.

This enables code like `m.row(0).transposed_slice().argmax()` to get argmax along the rows, and in general might be very useful for performance.


Andlon:
This PR has gone stale, but it seems like a useful addition. We might want to update it, and rename `slice` to `view` in the process, i.e. `transposed_view()`.

PR #629, opened 2019-08-06
Add implementation for Point + Point

kgreenek:


sebcrozet:
Hi!

The absence of implementation of addition between points is actually a design choice because it is not very meaningful from a mathematical/geometric point of view.

Adding a vector to a point on the other hand exists, so if you really need to add two points together you can do it with the vector representation of the second point: `let point_sum = point1 + point2.coords`.

Issue #620, opened 2019-07-23
Wrong result of min() and max() of nalgebra::base::Matrix


w1th0utnam3:
The methods `nalgebra::base::Matrix::min()` and `nalgebra::base::Matrix::max()` are documented as

> Returns the component with the smallest value.

and

> Returns the component with the largest value.

However, `A.min()` actually returns `N::zero().min(A[..])` (if this was a valid expression) and `A.max()` returns something like `N::zero().max(A[..])`. E.g.

- `na::Vector3::new(-1.0, -2.0, -3.0).max()` returns 0.0
- `na::Vector3::new(1.0, 2.0, 3.0).min()` returns 0.0

This is due to the fact that both rely on [`xcmp`](https://github.com/rustsim/nalgebra/blob/0384268bd4c757b5b0125d63bb0599695b961c48/src/base/ops.rs#L871) which always initializes the comparison value with `N::zero()` instead of e.g. the first value of the matrix. This is either a bug in the code or in the documentation.

Currently I have to use `A[A.imin()]` instead of `A.min()`.

sebcrozet:
Thank you for reporting this! This is definitely a bug in the code.

Andlon:
@sebcrozet: Note that this would give a `min`/`max` of `T::zero()` for 0x0 sized matrices (which might frequently happen for edge cases of algorithms that need to dynamically allocate matrices). We might want to think about the desired behavior for this edge case.

I've had exactly this issue with Eigen (C++) before: they didn't properly/consistently handle edge cases for 0x0 matrices, which led to some very painful workarounds for edge cases in various algorithms that I was writing.

Another, possibly more important point, is the handling of NaNs. It seems to me that it if you have a vector `[1, NaN, 3]`, it would be desirable to have `min`/`max`/ both return `NaN`, otherwise you might easily introduce silent errors in code (as you would never observe the `NaN` if all you ever do is call `min`/`max`).

Of course, this is part of the reason for why e.g. `Iterator::min()` and `Iterator::max()` require `Ord` in the first place... (I realize the reason that `nalgebra` only requires `PartialOrd` is pragmatism, but still)

Just some things to think about!

w1th0utnam3:
Wrt to the `NaN` problem I would simply propose

```
    fn xcmp<N2>(&self, abs: impl Fn(N) -> N2, cmp: impl Fn(N2, N2) -> bool) -> N2
        where N2: Scalar + PartialOrd + Zero {
[...]

PR #630, opened 2019-08-10
Fixes for [a,ca]min/max methods

w1th0utnam3:
Fixes for problems discussed in https://github.com/rustsim/nalgebra/issues/620
- Fixed functions returning wrong results
- ~~Added assert on empty matrices~~
- ~~Propagate `NaN` (or rather any value for which `PartialOrd::partial_cmp` returns `None`)~~
- Added doctests and documentation on the methods' behavior

### DIFF
--- a/src/base/ops.rs
+++ b/src/base/ops.rs
@@ -870,9 +870,11 @@ impl<N: Scalar, R: Dim, C: Dim, S: Storage<N, R, C>> Matrix<N, R, C, S> {
     #[inline(always)]
     fn xcmp<N2>(&self, abs: impl Fn(N) -> N2, ordering: Ordering) -> N2
         where N2: Scalar + PartialOrd + Zero {
-        assert!(self.len() > 0, "Empty matrix supplied to min/max function.");
         let mut iter = self.iter();
-        let mut max = abs(iter.next().cloned().unwrap());
+        let mut max = match iter.next() {
+            Some(first) => abs(*first),
+            None => { return N2::zero(); }
+        };
 
         for e in iter {
             let ae = abs(*e);
@@ -881,8 +883,6 @@ impl<N: Scalar, R: Dim, C: Dim, S: Storage<N, R, C>> Matrix<N, R, C, S> {
                 if ae_ordering == ordering {
                     max = ae;
                 }
-            } else {
-                return ae;
             }
         }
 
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thanks! Passing an `Ordering` instead of a closure as parameter to `xcmp` is a great idea.

The CI failed for some unexpected reasons so I restarted it before merging this PR.

sebcrozet:
Looks like the CI continues to fail, because of some packages no longer available (libblas, and liblapack). I'll investigate this.

Andlon:
@sebcrozet: I don't know if you've read the discussion in #620 yet, but I'm a little concerned with the new behavior when encountering `None` results from `partial_cmp`. Just wanted to chime in before it gets merged. In any case it's a very important PR, so that we get a fix for the `max`/`min` methods in the common use case.

w1th0utnam3:
For now, to just fix the wrong result without breaking changes, we could just remove the `else` branch in
```rust
            if let Some(ae_ordering) = ae.partial_cmp(&max) {
                if ae_ordering == ordering {
                    max = ae;
                }
            } else {
                return ae;
            }
```
to keep the old behavior of ignoring `NaN`'s etc. and instead of the `assert!` in
```rust
        assert!(self.len() > 0, "Empty matrix supplied to min/max function.");
        let mut iter = self.iter();
        let mut max = abs(iter.next().cloned().unwrap());
```
replace it with the initially suggested `let mut max = abs(it.next().cloned().unwrap_or(N::zero()));`?

w1th0utnam3:
Ok, just updated the branch. `assert!` is gone and old behavior on `partial_cmp() => None` is restored.

sebcrozet:
Thank you to chime in @Andlon. I did not read you comment about [rational numbers](https://github.com/rustsim/nalgebra/issues/620#issuecomment-520231273) carefully, and it make sense. I agree we should start by keeping this PR simple an fix the original bug.

Thank you for updating your PR @w1th0utnam3! One remark, the following:

```rust
   if let Some(ae_ordering) = ae.partial_cmp(&max) {
                if ae_ordering == ordering {
```

could be simply written as: `if ae.partial_cmp(&max) == Some(ordering) {`

w1th0utnam3:
Ah right, I still have to learn a bit wrt Rust patterns :smile:
Fixed it.

sebcrozet:
@w1th0utnam3 The CI is fixed. For the fix to take effect here, could you rebase your PR to the head of the `dev` branch please?

sebcrozet:
All good, thanks!

PR #631, opened 2019-08-11
Basic matrixcompare functionality

Andlon:
~~**Note: this is not ready for merging. I'm posting it here for some initial feedback.**~~
*Update: This is in principle finalized and ready for merging (barring minor issues, see below)*.

This PR provides `matrixcompare` functionality for `nalgebra`. `matrixcompare` is a crate which aims to provide generic functionality for debug comparisons of matrices. A consumer crate like `nalgebra` only needs to define how many rows/columns a matrix has and how to fetch coefficients. Macros like `assert_matrix_eq!` then makes it easy to compare matrices with different kinds of comparators (absolute tolerance, relative tolerance etc.), which significantly simplifies debugging when writing unit tests involving linear algebra code.

`matrixcompare` is the result of extracting, rewriting and generalizing similar functionality found in the `rulinalg` crate (and also originally contributed by me).

Note that `matrixcompare` is currently not published, and is also not complete. Obviously, this will have to be remedied before this PR can be merged. I open it now so that I can perhaps have some initial feedback (including CI runs).

Below is an excerpt from one of the new tests.

```rust
let a = MatrixMN::<_, U4, U5>::from_row_slice(&[
        1210, 1320, 1430, 1540, 1650,
        2310, 2420, 2530, 2640, 2750,
        3410, 3520, 3630, 3740, 3850,
        4510, 4620, -4730, 4840, 4950,
]);

let b = MatrixMN::<_, U4, U5>::from_row_slice(&[
        1210, 1320, 1430, 1540, 1650,
        2310, 2420, 2530, 2640, 2750,
        3410, 3520, 3630, 3740, 3850,
        4510, 4620, 4730, 4840, 4950,
]);

assert_matrix_eq!(a, b);
```

And this is the output (format subject to change) at the time of writing:

```
Matrices X and Y have 1 mismatched element pairs.
The mismatched elements are listed below, in the format
(row, col): x = X[[row, col]], y = Y[[row, col]].

 (3, 2): x = -4730, y = 4730.

Comparison criterion: exact equality x == y.
Please see the documentation for ways to compare matrices approximately.
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ rand_isaac = "0.2"
 #criterion = "0.2.10"
 
 # For matrix comparison macro
-matrixcompare = "0.1"
+matrixcompare = "0.1.3"
 
 [workspace]
 members = [ "nalgebra-lapack", "nalgebra-glm" ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ sparse = [ ]
 debug = [ "approx/num-complex", "rand/std" ]
 alloc = [ ]
 io = [ "pest", "pest_derive" ]
+compare = [ "matrixcompare-core" ]
 
 [dependencies]
 typenum        = "1.11"
@@ -50,6 +51,7 @@ mint           = { version = "0.5", optional = true }
 quickcheck     = { version = "0.9", optional = true }
 pest           = { version = "2.0", optional = true }
 pest_derive    = { version = "2.0", optional = true }
+matrixcompare-core = { version = "0.1", optional = true }
 
 [dev-dependencies]
 serde_json = "1.0"
@@ -60,6 +62,9 @@ rand_isaac = "0.2"
 ### compilation for #[no-std] because of the terrible Cargo bug
 ### https://github.com/rust-lang/cargo/issues/4866
 #criterion = "0.2.10"
+
+# For matrix comparison macro
+matrixcompare = "0.1"
 
 [workspace]
 members = [ "nalgebra-lapack", "nalgebra-glm" ]
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 all:
-	cargo test --features "debug arbitrary serde-serialize abomonation-serialize"
+	cargo test --features "debug arbitrary serde-serialize abomonation-serialize compare"
 	# cargo check --features "debug arbitrary serde-serialize"
 
 doc:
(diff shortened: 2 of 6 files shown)


Andlon:
Thanks for having a look! You're right, `panic_message` is a terrible name. It's somewhat of an old relic, in which it was used exclusively to panic inside of a macro. But I somehow didn't change the name as I was redesigning the functionality. I like your suggestions, I will try to apply them next chance I get!

Andlon:
@sebcrozet: I finally managed to revisit this PR. I made the changes you suggested earlier, and [matrixcompare](https://crates.io/crates/matrixcompare) has now been released on `crates.io`. I've updated this PR to use the updated version.

`matrixcompare` also supports sparse matrices (and dense-sparse comparisons of course), which is why I'm trying to get this across the finishing line now. It makes it much easier to write tests for my upcoming sparse contribution. This PR only includes support for `nalgebra` dense matrices, however. I will include sparse support in the upcoming CSR/CSC redesign.

The `matrixcompare` crate is organized in terms of the core traits, that reside in `matrixcompare-core`, and further re-exported in the `matrixcompare` crate. The idea here is that libraries like `nalgebra` need only rely on `matrixcompare-core` (which has no dependencies and is expected to need updates very rarely, reducing dependency breakage), whereas consumers (e.g. nalgebra's own test suite) rely on whatever version of `matrixcompare` that they want to in their tests.

I'm not sure how to really do it with features etc. I added a "compare" feature which would toggle the `matrixcompare-core` dependency. I cannot call it "matrixcompare", because `matrixcompare` is a dev-dependency already, which causes a conflict. Also, unfortunately I cannot toggle this feature for our tests by default, but I'm sure you're aware of this particular problem.

I see that the CI fails for `no-std` tests. Not exactly sure how things fit together here, I have basically almost zero experience with `no-std`.

sebcrozet:
Thank you for this PR! The CI failure when targeting `no-std` is caused by the `matrixcompare` dev-dependency which itself dependends on `num-traits` which has its `std` dependency enabled. Unfortunately, cargo will unify enabled features for all dependencies, including dev-dependencies, when building the crate. So if `num-traits/std` is enabled on one dev-dependency, it will be enabled for all builds of `nalgebra` itself. That pr should fix this: https://github.com/Andlon/matrixcompare/pull/2

sebcrozet:
Thanks!

PR #635, opened 2019-08-19
Implemented LowerExp for Matrix

p-avital:
It's a simple one, but working with usually either huge or very small floats, I've really been missing this format.
It's basically just a copy-paste of the Display implementation, but using the LowerExp formater for values.

### DIFF
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -1430,6 +1430,89 @@ where
     }
 }
 
+impl<N, R: Dim, C: Dim, S> fmt::LowerExp for Matrix<N, R, C, S>
+    where
+        N: Scalar + fmt::LowerExp,
+        S: Storage<N, R, C>,
+        DefaultAllocator: Allocator<usize, R, C>,
+{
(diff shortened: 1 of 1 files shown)


sebcrozet:
Hi! Thank you for this contribution.
Could we avoid the code duplication using a macro? We could define a `impl_fmt` macro that we could call like that to generate the implementations:

```rust
impl_fmt!(Display, "{}", "{:.1$}");
impl_fmt!(LowerExp, "{:e}", "{:.1$e}");
```

p-avital:
> Hi! Thank you for this contribution.
> Could we avoid the code duplication by using a macro? We could define a `impl_fmt` macro that we could call like that to generate the implementations:
> 
> ```rust
> impl_fmt!(Display, "{}", "{:.1$}");
> impl_fmt!(LowerExp, "{:e}", "{:.1$e}");
> ```

Done, took the occasion to add implementations for all the format traits in std (except Debug which should probably stay derived).

sebcrozet:
> Done, took the occasion to add implementations for all the format traits in std (except Debug which should probably stay derived).

Great, thanks!

PR #638, opened 2019-08-24
Don't export modules whose contents are reexported

Ralith:
Reduces confusion over the correct way to reference a type, particularly due to documentation and error messages. A downside to this is that all the typenums show up on the root doc page, but I think it's still a net win.


Ralith:
On second thought maybe the dimension stuff deserves a special case. TODO.

Ralith:
This seems like more of a happy medium. For some reason I had to move dimension.rs from base/ to the root to get rustdoc to not inline its contents into the root module documentation.

Andlon:
Given the age of this PR and its sweeping changes (and therefore conflicts with current `dev`), I think it might be hard to update. It's unclear to me if the issue that this PR fixed still persists or not, but in any case, we are probably better off starting off with a new PR if we wanted to attack the issue again. I will therefore close the PR. Please feel free to re-open if you think this is the wrong decision, @Ralith!

PR #643, opened 2019-08-30
Allow getting .min() and .max() of matrices of unsigned integers

koendeschacht:
I have also added a test to the documentation, although I'm not 100% sure that's the right place to put them.

Let me know if I can improve something for this PR.


koendeschacht:
Sure! Updated

sebcrozet:
Thanks!

PR #647, opened 2019-09-02
Added the mean computation of a vector of unit quaternions

thibaultbarbie:
When we want to compute the mean of multiple 3D rotations the documentation is scarce and the methods difficult. However, the paper of Markley et al.[1] describes an algorithm easy to implement to compute it. 

I implemented it for me but thought that it may be useful for others. I am still a rust beginner so my code may not be perfect so please feel free to correct it.
 
[1] Markley, F. L., Cheng, Y., Crassidis, J. L., & Oshman, Y. (2007). Averaging quaternions. Journal of Guidance, Control, and Dynamics, 30(4), 1193-1197.

### DIFF
--- a/src/geometry/quaternion_construction.rs
+++ b/src/geometry/quaternion_construction.rs
@@ -13,7 +13,7 @@ use alga::general::RealField;
 
 use crate::base::dimension::U3;
 use crate::base::storage::Storage;
-use crate::base::{Unit, Vector, Vector3, Vector4, Matrix3};
+use crate::base::{Unit, Vector, Vector3, Vector4, Matrix3, Matrix4};
 
 use crate::geometry::{Quaternion, Rotation3, UnitQuaternion};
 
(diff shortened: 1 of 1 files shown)


src/geometry/quaternion_construction.rs
```diff
@@ -676,6 +676,54 @@ impl<N: RealField> UnitQuaternion<N> {
...
+    /// let q3 = UnitQuaternion::from_euler_angles(0.1, 0.0, 0.0);
+    ///
+    /// let quat_vec = vec![q1, q2, q3];
+    /// let q_mean = UnitQuaternion::quaternions_mean(&quat_vec);
+    ///
+    /// let euler_angles_mean = q_mean.euler_angles();
+    /// assert_relative_eq!(euler_angles_mean.0, 0.0, epsilon = 1.0e-7)
+    /// ```
+    #[inline]
+    pub fn quaternions_mean(unit_quaternions: &Vec<Self>) -> Self {
```
Ralith:
```suggestion
    pub fn mean_of(unit_quaternions: &[Self]) -> Self {
```

Or maybe even

```suggestion
    pub fn mean_of(unit_quaternions: impl IntoIterator<Item=Self>) -> Self {
```

src/geometry/quaternion_construction.rs
```diff
@@ -676,6 +676,54 @@ impl<N: RealField> UnitQuaternion<N> {
...
+        assert!(unit_quaternions.len() > 0);
+        let quaternions_matrix: Matrix4<N> = unit_quaternions
+            .iter()
+            .map(|q| q.as_vector() * q.as_vector().transpose())
+            .sum();
+
+        let eigen_matrix = quaternions_matrix
+            .try_symmetric_eigen(N::RealField::default_epsilon(), 10)
+            .expect("Could not perform eigen decomposition when averaging quaternions.");
```
Ralith: Are there any cases where this can actually happen? If so, it'd be helpful to describe them in the message here, and document them in the function's documentation. Otherwise, the message should probably express that the case is supposed to be impossible.

src/geometry/quaternion_construction.rs
```diff
@@ -676,6 +676,54 @@ impl<N: RealField> UnitQuaternion<N> {
...
+    ///
+    /// let quat_vec = vec![q1, q2, q3];
+    /// let q_mean = UnitQuaternion::quaternions_mean(&quat_vec);
+    ///
+    /// let euler_angles_mean = q_mean.euler_angles();
+    /// assert_relative_eq!(euler_angles_mean.0, 0.0, epsilon = 1.0e-7)
+    /// ```
+    #[inline]
+    pub fn quaternions_mean(unit_quaternions: &Vec<Self>) -> Self {
+        assert!(unit_quaternions.len() > 0);
```
Ralith: The documentation should state that calling with an empty set will panic.

src/geometry/quaternion_construction.rs
thibaultbarbie: It theory it should not happen, as the quaternion matrix should be real symmetric and not zero.

"Could not perform eigen decomposition on the quaternions matrix. This case should not be possible." Does it look right ?

src/geometry/quaternion_construction.rs
thibaultbarbie: In your first suggestion you changed the vector argument into a slice. However, both of them are used as a reference. What would be the advantage of using a slice?

For your second suggestion, would it allow any structure that implement IntoIterator? If yes it is pretty neat! But apparently the compiler is complaining about the `map` function.

>  = note: the method `map` exists but the following trait bounds were not satisfied:
            `&mut impl IntoIterator<Item = Self> : std::iter::Iterator`
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following traits define an item `map`, perhaps you need to implement one of them:
            candidate #1: `std::iter::Iterator`
            candidate #2: `generic_array::functional::FunctionalSequence`

 I suppose I need to specify that the elements we are iterating on are mutable. Could you please guide me on how to write it (I almost never used traits in Rust yet) ?

Ralith: > What would be the advantage of using a slice?

Things other than vectors can produce a slice; for example, statically sized stack-allocated arrays.

> But apparently the compiler is complaining about the map function.

`map` is a method on `Iterator`, not `IntoIterator`. You need to call `into_iter` to get an `Iterator` from an `IntoIterator`.

src/geometry/quaternion_construction.rs
Ralith: It's best if you can concisely say *why* it's not possible, or failing that, briefly identify the violated assumption and elaborate in a nearby comment.

thibaultbarbie:
Thank you Ralith for your suggestions! I changed the code accordingly.

I am still unsure about the matrix message as it should really never occur (because it should panic if the matrix is zero).

sebcrozet:
Thank you @thibaultbarbie! I think the current error message is OK, considering this is more an internal error than something that the user is responsible for.

Issue #616, opened 2019-07-05
1.36 borrow checker 


nicodemus26:
Rust 1.36 (stable) raises a warning from the new borrow checker that will be come an error in the future, and represents potentially undefined behavior. Maybe related to this issue: https://github.com/rustsim/nalgebra/issues/561

See also: https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html#nll-for-rust-2015

```warning[E0502]: cannot borrow `*self` as immutable because it is also borrowed as mutable
   --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/nalgebra-0.16.13/src/base/cg.rs:291:44
    |
291 |                 self[(j, i)] += shift[j] * self[(D::dim() - 1, i)];
    |                 ---------------------------^^^^-------------------
    |                 |                          |
    |                 |                          immutable borrow occurs here
    |                 mutable borrow occurs here
    |                 mutable borrow later used here
    |
    = warning: this error has been downgraded to a warning for backwards compatibility with previous releases
    = warning: this represents potential undefined behavior in your code and this warning will become a hard error in the future```

est31:
Yes, this is a dupe of #561 .

You are using nalgebra 0.16.13 which is ancient. This bug is fixed since the 0.18.0 release in March 31, 2019.

PR #562, opened 2019-03-12
Fix two compiler warnings

est31:
Fixes #561

Issue #561, opened 2019-03-12
cannot borrow `*self` as immutable


est31:
Hi, as I'm seeing Rust developers getting serious about enabling the new borrow checker on the 2015 edition (https://github.com/rust-lang/rust/issues/57804 / https://github.com/rust-lang/rust/pull/59114), I compiled my local project with `RUSTFLAGS="-Zborrowck=migrate -Ztwo-phase-borrows" cargo +nightly check` to test for any possible future problems.

When it compiled `nalgebra 0.16.13` for me, I got multiple newly introduced compiler warnings. Even on nalgebra master, I'm getting one compiler warning:

```
warning[E0502]: cannot borrow `*self` as immutable because it is also borrowed as mutable
   --> src/base/cg.rs:297:44
    |
297 |                 self[(j, i)] += shift[j] * self[(D::dim() - 1, i)];
    |                 ---------------------------^^^^-------------------
    |                 |                          |
    |                 |                          immutable borrow occurs here
    |                 mutable borrow occurs here
    |                 mutable borrow later used here
    |
    = warning: this error has been downgraded to a warning for backwards compatibility with previous releases
    = warning: this represents potential undefined behavior in your code and this warning will become a hard error in the future
```

I'd love a fix of the issue and a new crates.io release soon after that. Thanks!

est31:
@matthewjasper is this behaviour intended? Or is it a legitimate error of the nalgebra crate?

sebcrozet:
Hi! Thank you for reporting this and for the fix. I find this warning strange, so I prefer to wait for what @matthewjasper has to say about it before merging anything.

Also keep in mind we don't intend to support rust 2015 indefinitely so at some point in the future rust 2018 will likely be necessary for using nalgebra (e.g. whenever specialization and const-generics land).

matthewjasper:
Yes, this is intended. It's the result of fixing rust-lang/rust#27868 .

est31:
Yeah, so I guess I'd ask my PR to get merged then.

est31:
This might already arrive in nightlies in next week. The sooner a fixed version is available on crates.io, the better.

PR #649, opened 2019-09-05
[v0.16] Fix NLL compatibility

Mark-Simulacrum:
Uncovered by a crater run in https://github.com/rust-lang/rust/pull/63565, the most recent v0.16.x version breaks when NLL is fully enabled. This PR is targeted to be based atop tag v0.16.13 (it's probably easiest to merge it manually, unless you want to create a branch for this PR to target).

### DIFF
--- a/src/base/cg.rs
+++ b/src/base/cg.rs
@@ -288,7 +288,8 @@ impl<N: Scalar + Ring, D: DimName, S: StorageMut<N, D, D>> SquareMatrix<N, D, S>
     {
         for i in 0..D::dim() {
             for j in 0..D::dim() - 1 {
-                self[(j, i)] += shift[j] * self[(D::dim() - 1, i)];
+                let add = shift[j] * self[(D::dim() - 1, i)];
+                self[(j, i)] += add;
             }
         }
     }


est31:
See also #561 , #562 and #616 . Interesting that 0.16.x is still used so much in the wild.

sebcrozet:
@Mark-Simulacrum Thank you for this PR! I've create a `v0.16` branch which is the new target of this PR. Why is the `#![allow(deprecated)]` necessary!

Mark-Simulacrum:
Ah, I think I forgot to drop that, it's not strictly necessary, I can drop it if you'd like.

sebcrozet:
> I can drop it if you'd like.

That would be great!

Mark-Simulacrum:
Okay, I've dropped the deprecation commit, so this should be good to go. (Closed and reopened because CI failed due to seemingly unrelated -- perhaps network -- reasons).

sebcrozet:
Thanks! Don't mind the CI. It will fail because of a changes on travis ubuntu version. We fixed our `.travis.yml` after the 16.x releases.

PR #650, opened 2019-09-05
Fixes #637: removes not used parameter from cross

### LINKED ISSUE #637: nalgebra_glm::cross doesn't use D
```rust
pub fn cross<N: Number, D: Dimension>(x: &TVec3<N>, y: &TVec3<N>) -> TVec3<N> {
    x.cross(y)
}
```
It is parameterized on `D`, but doesn't use it at all, so you have to manually specify a Dimension parameter that isn't used.

adevaykin:


sebcrozet:
Thanks!

PR #660, opened 2019-10-08
Rand 0.6 -> 0.7

dodomorandi:
With the new version of rand it is possible to use `Distribution<f32>` with `Matrix::from_distribution`, which is pretty handy.

* Bumped rand version to 0.7
* Added dependency to rand_distr
* Bumped quickcheck version to 0.9 (because of rand)
* Bumped rand_xorshift version to 0.2

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ path = "src/lib.rs"
 
 [features]
 default         = [ "std" ]
-std             = [ "matrixmultiply", "rand/std", "alga/std" ]
+std             = [ "matrixmultiply", "rand/std", "rand_distr", "alga/std" ]
 stdweb          = [ "rand/stdweb" ]
 arbitrary       = [ "quickcheck" ]
 serde-serialize = [ "serde", "serde_derive", "num-complex/serde" ]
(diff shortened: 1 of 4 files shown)


dodomorandi:
I think that this is pretty redudant with https://github.com/rustsim/nalgebra/pull/622, but at the same time it includes only the dependency update, without any particular tweaks.

dodomorandi:
It looks like there is some sort of numeric instability of worse, because I was not able to reproduce the failing tests shown by travis (different between nightly and beta), and after some tests I decided to re-run the CI on my repo... and all tests are passing.

The issue does not seem to be related to my PR, but I think that it is interesting and it should be taken into account as a possible problem.

sebcrozet:
Thank you for this PR @dodomorandi! Yes, I think it's a good idea to have a PR dedicated to the rand update. The tweaks are actually why #622 has not been merged yet.

You are right, it can happen that the test fail sometimes if the randomly generated input matrices were badly conditioned. We've not taken the time to address this, so I usually just restart the CI and see if it keeps failing (in which case the problem would likely be caused by the PR).

sebcrozet:
Thanks!

PR #665, opened 2019-10-21
Add component impls

### LINKED ISSUE #619: component_min and component_max
I've had to write my own min and max functions that take in two &Vector3 and computes a new vector containing the minimum components of the two vectors (and the max for the max function). Is there something like this already in nalgebra? max/min and partial_max/partial_min did not seem to do what I want. If this is something that would be beneficial to the library I would be happy to add it.

mandeep:
fixes #619


mandeep:
Seems to be an issue with the rebase. Let me fix this.

src/base/componentwise.rs
```diff
@@ -40,6 +41,44 @@ impl<N: Scalar, R: Dim, C: Dim, S: Storage<N, R, C>> Matrix<N, R, C, S> {
...
+    /// ```
+    /// use nalgebra::Vector3;
+    ///
+    /// let u = Vector3::new(-1.0, 3.0, 2.0);
+    /// let v = Vector3::new(1.0, 2.0, 3.0);
+    /// assert_eq!(u.component_min(&v), Vector3::new(-1.0, 2.0, 3.0))
+    /// ```
+    pub fn component_min(&self, rhs: &Matrix<N, R, C, S>) -> MatrixMN<N, R, C>
+    where
+        N: Ord,
```
Caellian: I think `Ord` is a too strong of a bound. `PartialOrd` might be a better choice due to `f32` and `f64` not being `Ord`. For minimum, `if a > b { b } else {a}` would be an ok default (prefer first operand if not comparable).

Also, bump :smile: 

I found `inf` and `sup` though the referenced issue, but always expect "component_min" and "component_max" instead.

mandeep: Thanks. `PartialOrd` does work better here.

src/base/componentwise.rs
```diff
@@ -43,7 +44,45 @@ impl<T: Scalar, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
...
+    /// ```
+    /// use nalgebra::Vector3;
+    ///
+    /// let u = Vector3::new(-1.0, 3.0, 2.0);
+    /// let v = Vector3::new(1.0, 2.0, 3.0);
+    /// assert_eq!(u.component_min(&v), Vector3::new(-1.0, 2.0, 2.0))
+    /// ```
+    pub fn component_min(&self, rhs: &Matrix<T, R, C, S>) -> OMatrix<T, R, C>
+    where
+        T: PartialOrd + Copy,
```
Caellian: `T` doesn't have to be `Copy`, `Clone` is enough and will be optimized away by the compiler. `Clone` is already provided by `Scalar`, so `T` just needs `PartialOrd` bound.

Then, there's `SimdPartialOrd::simd_min/max` which will use simd instead of a loop. So finally, `T: SimdPartialOrd`

src/base/componentwise.rs
```diff
@@ -43,7 +44,45 @@ impl<T: Scalar, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
...
+    /// # Example
+    ///
+    /// ```
+    /// use nalgebra::Vector3;
+    ///
+    /// let u = Vector3::new(-1.0, 3.0, 2.0);
+    /// let v = Vector3::new(1.0, 2.0, 3.0);
+    /// assert_eq!(u.component_max(&v), Vector3::new(1.0, 3.0, 3.0))
+    /// ```
+    pub fn component_max(&self, rhs: &Matrix<T, R, C, S>) -> OMatrix<T, R, C>
```
Caellian: rhs Matrix should have it's own storage generic to allow using the functions with matrices with different backing storages.

The `inf` and `sup` functions in the same file do the same. So these should just be `#[inline]` aliases to those. However, those two don't allow different storage for argument vector and take `rhs: &Self`, which in worst case will require 2 additional clones.

Tradeoff is flexibility or performance. Given that inf and sup already do what they do, maybe:
```rust
pub fn component_max<SB>(&self, rhs: &Matrix<T, R, C, SB>) -> OMatrix<T, R, C>
where
    T: PartialOrd,
    SB: Storage<T, R, C>,
    DefaultAllocator: Allocator<R, C>,
{ 
    self.zip_map(rhs, |a, b| if a > b { a } else { b })
}
```

Caellian: @sebcrozet can you give input on this? Should these just be inlined aliases?

mandeep: Good point. I personally don't think they should do anything different than `inf` and `sup` as that could cause confusion. Inline alias works fine for me and I'm happy to hear what everyone thinks.

Caellian: I believe Sebastian likely expected it to be an alias just as well.

Issue #655, opened 2019-09-20
Issue with Schur decomposition?


lesurp:
Hi,

# Code

```rust
    let p = nalgebra::Matrix2::new(2.5000002, -1.4999999, -1.4999999, 2.4999998);
    let schur = p.schur();
    println!("{}", schur.unpack().0);
    println!("{}", schur.complex_eigenvalues());
    match schur.eigenvalues() {
        Some(d) => println!("{}", d),
        None => println!("None"),
    }
```

# Result
```
  ┌                                         ┐
  │  0.7071068283270011  0.7071067340460907 │
  │ -0.7071067340460907  0.7071068283270011 │
  └                                         ┘

  ┌                         ┐
  │ 2.4999999999999996+NaNi │
  │ 2.4999999999999996+NaNi │
  └                         ┘

None
```

# Expected

An equivalent cpp code:

```cpp
#include <Eigen/Eigen>
#include <Eigen/src/Eigenvalues/RealSchur.h>
#include <iostream>
#include <iomanip>

int main(int argc, char *argv[])
{
    Eigen::Matrix2d p;
    p << 2.5000002, -1.4999999, -1.4999999, 2.4999998;

    auto schur = Eigen::RealSchur<Eigen::Matrix2d>(p);
    std::cout << std::setprecision(17);
    std::cout << schur.matrixU() << std::endl;
    std::cout << schur.matrixT() << std::endl;
}
```

produces:

```
 0.70710682832700111  0.70710673404609059
-0.70710673404609059  0.70710682832700111

3.9999999000000122                  0
                 0 1.0000000999999865
```

If I replace the `2.*` with an exact `2.5` the schur decomposition is correct. Note that the computed eigenvectors are correct - only the computed eigenvalues are wrong.

Am I misusing this decomposition? Of course for a 2*2 I don't *need* to use schur so it's not exactly a problem but still...

daingun:
Hi,

I'm having a similar problem calculating the eigenvalues of a 2x2 matrix. I think that both issues are the same.
```rust
let a = DMatrix::from_row_slice(2, 2, &[-2., 0., 3., -7.]);
let schur = Schur::new(a);
dbg!(&schur);
let poles = schur.complex_eigenvalues();
dbg!(poles);
```
Here the eigenvalues must be equal to `-2.` and `-7.`, however the result is `Complex {re: -4.499999999999999, im: NaN}` for both eigenvalues.

[...]

PR #668, opened 2019-11-01
Correct Schur decomposition for 2x2 matrices

daingun:
Due to rounding and possible loss of precision the lower left element of the 2x2 matrix
may be different from zero.

This optimization should fix issue #655. 

An example for this optimization can be found in Eigen code at the following [link](https://bitbucket.org/eigen/eigen/src/b58ba8bfb096e616cfe2fc39d1d94780174df809/Eigen/src/Eigenvalues/RealSchur.h#lines-419).

### DIFF
--- a/src/linalg/schur.rs
+++ b/src/linalg/schur.rs
@@ -413,6 +413,7 @@ where
             let inv_rot = rot.inverse();
             inv_rot.rotate(&mut m);
             rot.rotate_rows(&mut m);
+            m[(1, 0)] = N::zero();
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thanks!

daingun:
Hi,
please note that I used mutable index to change the value, I have not used get_unchecked_mut() to  avoid the introduction of unsafe code without any test.
If you may change it if you consider it appropriate for efficiency.

PR #669, opened 2019-11-01
Use same algorithm to solve 2x2 eigenvalue problem

daingun:
The eigenvalue problem is solved in two different method that use different methods
to calculate the discriminant of the solution to the quadratic equation.
Use the method whose computation is considered more stable.

Reference issue #655 in [comment](https://github.com/rustsim/nalgebra/issues/655#issuecomment-546641755)

### DIFF
--- a/src/linalg/schur.rs
+++ b/src/linalg/schur.rs
@@ -309,16 +309,17 @@ where
                 let hmn = t[(m, n)];
                 let hnn = t[(n, n)];
 
-                let tra = hnn + hmm;
-                let det = hnn * hmm - hnm * hmn;
-                let discr = tra * tra * crate::convert(0.25) - det;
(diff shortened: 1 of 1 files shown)


src/linalg/schur.rs
```diff
@@ -413,7 +414,6 @@ where
             let inv_rot = rot.inverse();
             inv_rot.rotate(&mut m);
             rot.rotate_rows(&mut m);
-            m[(1, 0)] = N::zero();
```
sebcrozet: Is this intentional? This reverts your change introduced by #668.

daingun: Sorry, my bad. The creation of a new pull request used the nalgebra/dev branch not mine.
I should have fixed the issue.

sebcrozet:
Thanks!

PR #671, opened 2019-11-02
insured Storage trait is used in solve methods

nestordemeure:
see issue 667 : https://github.com/rustsim/nalgebra/issues/667


sebcrozet:
Thanks!

PR #673, opened 2019-11-03
Cholupdate

nestordemeure:
Added code and test for cholesky decomposition update (see [issue 652](https://github.com/rustsim/nalgebra/issues/652)).

Also relaxed some traits from `StorageMut` to `Storage` in `solve.rs` methods (following [issue 667](https://github.com/rustsim/nalgebra/issues/667)).

My code to create a new matrix with one more/less row+column should be reviewed : it is correct but there is probably a more efficient way to do that using nalgebra internals.

The `insert_column` and `insert_row` methods take `&self` (where methods with the same name elsewhere in nalgebra take `self`) as it is the minimum required by the current implementation : it could be changed for homogenization purposes.

### DIFF
--- a/src/linalg/cholesky.rs
+++ b/src/linalg/cholesky.rs
@@ -6,8 +6,9 @@ use alga::general::ComplexField;
 use crate::allocator::Allocator;
 use crate::base::{DefaultAllocator, Matrix, MatrixMN, MatrixN, SquareMatrix};
 use crate::constraint::{SameNumberOfRows, ShapeConstraint};
-use crate::dimension::{Dim, DimSub, Dynamic, U1};
+use crate::dimension::{Dim, DimAdd, DimSum, DimSub, Dynamic, U1};
 use crate::storage::{Storage, StorageMut};
+use crate::base::allocator::Reallocator;
 
 /// The Cholesky decomposition of a symmetric-definite-positive matrix.
 #[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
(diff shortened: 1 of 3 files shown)


sebcrozet:
Thank you for this PR! The CI currently fails because the call to [`solve_lower_triangular`](https://github.com/rustsim/nalgebra/pull/673/files#diff-563510a5abdf0e0c919b3b00d7e4d50bR190] performs an allocation, which is not available when compiling to `#[no-std]` platforms. The allocation happens because the `solve_lower_triangular` involves dynamically-sized slices:

```rust
        let top_left_corner = self.chol.slice_range(..j, ..j);
        let col_jminus = col.rows_range(..j);
```

I will review this in more details at the end of this week or next week to see if this allocation could be avoided somehow.

nestordemeure:
Don't hesitate to ask me if I can be of any help.

sebcrozet:
@nestordemeure Hi! I've pushed a commit on your branch containing changes that should make the compilation work for `#[no-std]` environments too. Feel free to make further modifications if you think there is a better way of doing this.

My changes have two consequences:

- For `insert_column`, the column to insert is now taken by-move instead of by-ref. The idea is that we will use this column as a workspace and thus avoid allocations that were initially done by `lower_triangular_solve` and `clone_owned`.
- In `remove_column`, I create a workspace vector with a length equal to the number of rows of the already decomposed matrix. This is kind of unfortunate because we may end up with a workspace vector that is unnecessarily large (we will only actually use `D - j` elements). However this will avoid allocations when `D` is an integer known as compile-time because the workspace now has the dimension `D` too.

Please let me know if you are OK with those changes.

Also, I've noticed you don't put spaces before an after math operators. Could you add some? For example, `j+1` should be `j + 1`.

nestordemeure:
I am reading it at the moment.

Is the code used to create the new matrix (with one more/less column) satisfying as it is currently written ?

Regarding the missing spaces, I let rustfmt do the formating for me so did not bother while writing the code but I just checked the [doc](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md) and it seems that there is no option to get spaces around operators.

sebcrozet:
> Is the code used to create the new matrix (with one more/less column) satisfying as it is currently written ?

Yeah, it's the right of doing this until when we will start to remove the `new_uninitialized` constructors.

nestordemeure:
Nice to discover `rows_range_pair_mut`, I searched precisely for something like that while writing the code.

I just corrected some typo in a comment. Appart from that, everything in the Cholesky code seems ok to me.

sebcrozet:
Thanks!

PR #675, opened 2019-11-04
Implement From<&Matrix> for MatrixSlice

Andlon:
This PR adds some `From` implementations that allows us to construct matrix slices from matrix references. The main motivation for this is to allow for a simplified generic handling of matrix input when we don't care about whether the matrix sizes are static or dynamic:

```rust
fn solve(x: impl Into<DMatrixSliceMut<T>>, rhs: impl Into<DMatrixSlice<T>>) {
     // Solve something and store result in `x`
}
```

Thanks to @sebcrozet for proposing the correct trait bounds to use to avoid problems with overlapping implementations.

One remark: I put the tests in `tests/core/conversion.rs`, but this module requires the `arbitrary` feature, which is unfortunate given that the test itself does not use `quickcheck`.

### DIFF
--- a/src/base/conversion.rs
+++ b/src/base/conversion.rs
@@ -21,6 +21,7 @@ use crate::base::storage::{ContiguousStorage, ContiguousStorageMut, Storage, Sto
 #[cfg(any(feature = "std", feature = "alloc"))]
 use crate::base::VecStorage;
 use crate::base::{DefaultAllocator, Matrix, ArrayStorage, MatrixMN, MatrixSlice, MatrixSliceMut, Scalar};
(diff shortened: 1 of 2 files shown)


Andlon:
In the latest commit, I generalized the `From` impl to allow constructing slices which have different (but compatible through `DimEq`) stride types. However, I couldn't use `generic_slice` anymore, so I had to construct slices directly from raw pointers. There might be a better way to do this..

sebcrozet:
Thank you for this PR! Yes, the inability to use `generic_slice` is unfortunate. We should find a way to improve this in another PR (cf. #676).

PR #678, opened 2019-11-15
Import IsaacRng from rand_isaac instead of rand.

aweinstock314:
The rand crate removed IsaacRng in the 0.5 -> 0.6 transition, so the benchmarks that use it haven't compiled since 9c37c512039059281b210521cfd53e931f5fa0b5.


sebcrozet:
Thanks!

PR #682, opened 2019-11-29
Implement Extend<&N> for VecStorage.

m-ou-se:
`Extend<N>` was already implemented, but nalgebra vectors/matrices give
iterators that give `&N`, not `N`, so implementing `Extend<&N>` as well makes
it easier to use.

It seems common practice to do so: The standard library's `Vec` also
implments Extend for both `T` and `&T`.


sebcrozet:
Thanks!

Issue #679, opened 2019-11-15
Intermittent failures in `cholesky_solve` and `cholesky_solve_static`


aweinstock314:
I noticed this on the CI run for #678. The CI was running commit 8bce398653395c219edf85dcb4d95c3ff976a69f on `cargo 1.40.0-beta (5da4b4d47 2019-10-28)`. I was able to reproduce it locally on both `cargo 1.40.0-beta (5da4b4d47 2019-10-28)` and `cargo 1.38.0 (23ef9a4ef 2019-08-20)` on commmit 31ef5f0ab02c6ecf279867f07cd63e16cece8b75 (the previous HEAD) via the following commands:

```bash
for i in $(seq 100); do RUST_BACKTRACE=1 LAPACK=1 CARGO_FEATURE_SYSTEM_NETLIB=1 CARGO_FEATURE_EXCLUDE_LAPACKE=1 CARGO_FEATURE_EXCLUDE_CBLAS=1 sh -c "cd nalgebra-lapack && cargo +beta test cholesky_solve"; done
for i in $(seq 100); do RUST_BACKTRACE=1 LAPACK=1 CARGO_FEATURE_SYSTEM_NETLIB=1 CARGO_FEATURE_EXCLUDE_LAPACKE=1 CARGO_FEATURE_EXCLUDE_CBLAS=1 sh -c "cd nalgebra-lapack && cargo test cholesky_solve"; done
```

It's somewhat rare, there only seem to be 2-3 crashes per 100 runs.

Sample backtraces:
```
---- linalg::cholesky::cholesky_solve_static stdout ----
thread 'linalg::cholesky::cholesky_solve_static' panicked at '[quickcheck] TEST FAILED. Arguments: (Matrix { data: [69.69431563547087, 75.85874450854578, 80.51503043974236, 55.72021550747667, 50.60926688060886, -47.98551293031665, 35.8865183309388, -45.711636343254
995, 69.73960842665585, 29.100233100697324, 92.95087043025657, -82.45703076757903, 79.48975128880761, 27.7026291273621, 93.98360271057183, -51.463607224805855] })', /home/avi/.cargo/registry/src/github.com-1ecc6299db9ec823/quickcheck-0.9.0/src/tester.rs:178:28
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/libunwind.rs:88
[...]

PR #684, opened 2019-12-05
Add inlined_clone to Scalar, and relax bounds from `Scalar + Copy` to…

aweinstock314:
… `Scalar + Clone` nearly everywhere.

The various nalgebra-lapack FooScalars are still Copy because they make use of uninitialized memory.
nalgebgra-glm Number still uses Copy because upstream `approx` requires it.

### DIFF
--- a/examples/scalar_genericity.rs
+++ b/examples/scalar_genericity.rs
@@ -4,11 +4,11 @@ extern crate nalgebra as na;
 use alga::general::{RealField, RingCommutative};
 use na::{Scalar, Vector3};
 
-fn print_vector<N: Scalar + Copy>(m: &Vector3<N>) {
+fn print_vector<N: Scalar + Clone>(m: &Vector3<N>) {
     println!("{:?}", m)
 }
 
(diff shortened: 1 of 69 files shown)


sebcrozet:
Thanks! Why is `Scalar + Clone` still needed? Isn't just `Scalar` enough now that we rely on `.inlined_clone()` directly instead of `.clone()`.

aweinstock314:
A bunch of the `Allocator` machinery still requires `Clone`. I could add a `Clone` bound to `Scalar` if it's unlikely that we'll need non-Clone Scalars in the future, and replace `Scalar+Clone` with `Scalar` at all use sites; should I?

sebcrozet:
Yes, let's make `Scalar` require `Clone`. Doing so, we could provide in the trait a default implementation of `inline_clone` based on clone. This reduces further the work to be done by the end-user when implementing `Scalar` for its own non-`Copy` type.

We'll see in the future if some people come up with a use-case for non-clone types as scalars.

aweinstock314:
It looks like the failing test here is `cholesky_inverse`, which is probably the same intermittent conditioning issue as #679.

sebcrozet:
Thanks!

Issue #674, opened 2019-11-04
PartialEq implementation panics, plus generalization


Andlon:
The current implementation of `PartialEq` for `Matrix` only works between two instances of the exact same type. We should extend it to allow comparison of matrices that have different types (owned vs sliced, static vs dynamic shapes etc.), but represent the same matrix, in a mathematical sense.

Moreover, the current implementation of `PartialEq` panics when the sizes of the matrices don't match. Instead, it should simply return `false`.

Andlon:
I marked this as both a bug (due to the panic behavior) and enhancement (due to the desired generalization).

guissalustiano:
This issue can be close, no?

Andlon:
@guissalustiano: I believe this was fixed in #692, yes! Although it's slightly confusing to me, as the description of #692 suggests that there were some outstanding issues, but I think this was subsequently fixed in later commits/review, so I will close this issue. @SBrandeis: can you confirm that it's safe to close? (Otherwise I will re-open)

PR #692, opened 2020-01-19
Fix Matrix PartialEq bug and add some generalization

SBrandeis:
Addresses #674 (at least partly)

## What's new
- `matrix_a == matrix_b` will no longer panic when shapes mismatch.
- In addition, comparison can be done if the number of rows / columns or storage are of different types (static vs dynamic, owned vs slice).

## What's left to do
PartialEq trait for Matrices with `typenum`-typed number of rows and / or columns is not implemented symmetrically.
In other words, comparing two matrices with a `typenum`-typed Matrix on the left hand side of the `==` operator will work, but not if it is on the right hand side of the operator (compilation error).

This is because 
```
impl<
    T: Dim,
    U: ...,
    B: ...
> PartialEq<UInt<U, B>> for T { ...}
```
violates Rust orphan rules for Trait implementation (see error `E0210`), and I could not find a way to work around it. Using a local struct to cover the type seems not a good idea, as we lose all trait implementation done in `typenum` crate.

Any idea to solve this issue is welcome !

### DIFF
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -1356,10 +1356,7 @@ where
 {
     #[inline]
     fn eq(&self, right: &Matrix<N, R2, C2, S2>) -> bool {
-        if self.shape() == right.shape() {
-            return self.iter().zip(right.iter()).all(|(l, r)| l == r)
-        }
-        false
+        self.shape() == right.shape() && self.iter().zip(right.iter()).all(|(l, r)| l == r)
     }
 }
 
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -1137,7 +1137,7 @@ impl<N: Scalar + Zero + One, D: DimAdd<U1> + IsNotStaticOne, S: Storage<N, D, D>
     where DefaultAllocator: Allocator<N, DimSum<D, U1>, DimSum<D, U1>> {
         assert!(self.is_square(), "Only square matrices can currently be transformed to homogeneous coordinates.");
         let dim = DimSum::<D, U1>::from_usize(self.nrows() + 1);
-        let mut res = MatrixN::identity_generic(dim, dim); 
+        let mut res = MatrixN::identity_generic(dim, dim);
         res.generic_slice_mut::<D, D>((0, 0), self.data.shape()).copy_from(&self);
         res
     }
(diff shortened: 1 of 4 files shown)


SBrandeis:
CI fails with the following message:
```
error[E0283]: type annotations needed: cannot resolve `_: na::Dim`

  --> tests/core/serde.rs:26:5

   |

26 |     assert_eq!(v, serde_json::from_str(&serialized).unwrap());

   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

   |

   = note: required because of the requirements on the impl of `std::cmp::PartialEq<na::Matrix<f32, _, _, _>>` for `na::Matrix<f32, na::Dynamic, na::Dynamic, na::VecStorage<f32, na::Dynamic, na::Dynamic>>`

   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```
Apparently, compiler is not able to determine the type of `serde_json::from_str(&serialized).unwrap()` without type annotations.
Not sure how to handle this, any suggestion welcome.

src/base/matrix.rs
```diff
@@ -1344,18 +1344,22 @@ where
...
-        assert!(
-            self.shape() == right.shape(),
-            "Matrix equality test dimension mismatch."
-        );
-        self.iter().zip(right.iter()).all(|(l, r)| l == r)
+    fn eq(&self, right: &Matrix<N, R2, C2, S2>) -> bool {
+        if self.shape() == right.shape() {
+            return self.iter().zip(right.iter()).all(|(l, r)| l == r)
+        }
+        false
```
sebcrozet: This can be a one-liner:

```rust
self.shape() == right.shape() && self.iter().zip(right.iter()).all(|(l, r)| l == r)
```

SBrandeis:
Thank you for your review.
`PartialEq` trait bounds were indeed unnecessary, I removed them.
SerDe tests still raise a `E0283` error, what should be the strategy to solve this ?
Should I modify the test to add type annotations ?
Is this PR likely to introduce breaking changes ?

sebcrozet:
Thanks!

PR #695, opened 2020-02-05
GivensRotation::new() should default to (I, 0)

ilya-epifanov:
The bug was introduced in b5f452087e3367b60dfc097078c5862f17bd8dc9
SVD sometimes fails trying to do `None.unwrap()` in `GivensRotation::new()`


ilya-epifanov:
The fix applies to all releases since v0.18.0

sebcrozet:
Thanks!

PR #696, opened 2020-02-23
Small docs improvements

nnmm:
* The `Perspective3` docs mentioned an "orthographic" projection where it should say "perspective"
* Replaced "an homogeneous" by "a homogeneous"
* Removed a double colon in lib.rs:54


sebcrozet:
Thanks @nnmm!

PR #705, opened 2020-03-10
Fix warnings on latest rustc

ProfFan:
Should be very straightforward :)

### DIFF
--- a/src/base/conversion.rs
+++ b/src/base/conversion.rs
@@ -118,12 +118,13 @@ macro_rules! impl_from_into_asref_1D(
               S: ContiguousStorage<N, $NRows, $NCols> {
             #[inline]
             fn into(self) -> [N; $SZ] {
-                unsafe {
-                    let mut res: [N; $SZ] = mem::MaybeUninit::uninit().assume_init();
-                    ptr::copy_nonoverlapping(self.data.ptr(), &mut res[0], $SZ);
+                let mut res = mem::MaybeUninit::<[N; $SZ]>::uninit();
 
-                    res
-                }
+                unsafe { ptr::copy_nonoverlapping(self.data.ptr(), res.as_mut_ptr() as *mut N, $SZ); }
+
+                let res: [N; $SZ] = unsafe { res.assume_init() };
+                
+                res
             }
         }
 
@@ -185,12 +186,13 @@ macro_rules! impl_from_into_asref_2D(
         where S: ContiguousStorage<N, $NRows, $NCols> {
             #[inline]
             fn into(self) -> [[N; $SZRows]; $SZCols] {
-                unsafe {
-                    let mut res: [[N; $SZRows]; $SZCols] = mem::MaybeUninit::uninit().assume_init();
-                    ptr::copy_nonoverlapping(self.data.ptr(), &mut res[0][0], $SZRows * $SZCols);
+                let mut res = mem::MaybeUninit::<[[N; $SZRows]; $SZCols]>::uninit();
 
-                    res
-                }
+                unsafe { ptr::copy_nonoverlapping(self.data.ptr(), res.as_mut_ptr() as *mut N, $SZRows * $SZCols); }
+                
+                let res: [[N; $SZRows]; $SZCols] = unsafe { res.assume_init() };
+                
+                res
             }
         }
 
--- a/src/base/conversion.rs
+++ b/src/base/conversion.rs
@@ -120,11 +120,9 @@ macro_rules! impl_from_into_asref_1D(
             fn into(self) -> [N; $SZ] {
                 let mut res = mem::MaybeUninit::<[N; $SZ]>::uninit();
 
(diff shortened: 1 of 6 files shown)


src/base/conversion.rs
```diff
@@ -119,7 +119,7 @@ macro_rules! impl_from_into_asref_1D(
             #[inline]
             fn into(self) -> [N; $SZ] {
                 unsafe {
-                    let mut res: [N; $SZ] = mem::uninitialized();
+                    let mut res: [N; $SZ] = mem::MaybeUninit::uninit().assume_init();
```
josephlr: Per the [`MaybeUninit` docs](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initialization-invariant), this updated code does not avoid the undefined behavior that `MaybeUninit` tries to solve.

The correct code would be:
```rust
let mut res: MaybeUninit<[N; $SZ]> = MaybeUninit::uninit();
ptr::copy_nonoverlapping(self.data.ptr(), res.as_mut_ptr() as *mut $N, $SZ);
res.assume_init()
```

src/base/conversion.rs
```diff
@@ -186,7 +186,7 @@ macro_rules! impl_from_into_asref_2D(
             #[inline]
             fn into(self) -> [[N; $SZRows]; $SZCols] {
                 unsafe {
-                    let mut res: [[N; $SZRows]; $SZCols] = mem::uninitialized();
+                    let mut res: [[N; $SZRows]; $SZCols] = mem::MaybeUninit::uninit().assume_init();
```
josephlr: Similarly here:
```rust
let mut res: MaybeUninit<[[N; $SZRows]; $SZCols]> = MaybeUninit::uninit();
ptr::copy_nonoverlapping(self.data.ptr(), res.as_mut_ptr() as *mut $N, $SZRows * $SZCols);
res.assume_init()
```

src/base/default_allocator.rs
```diff
@@ -46,7 +46,7 @@ where
 
     #[inline]
     unsafe fn allocate_uninitialized(_: R, _: C) -> Self::Buffer {
-        mem::uninitialized()
+        mem::MaybeUninit::uninit().assume_init()
```
josephlr: This is still undefined behavior, but at least the function is `unsafe`. This is the best we can do with the existing public API. See https://github.com/rustsim/nalgebra/issues/556 for more info on how this can be fixed.

src/base/conversion.rs
ProfFan: Thanks for the suggestion! Fixed now :)

src/base/default_allocator.rs
ProfFan: Yea I agree... Unfortunately I no longer have much time on this now...

ProfFan:
@sebcrozet Ping :)

src/base/conversion.rs
```diff
@@ -185,12 +186,13 @@ macro_rules! impl_from_into_asref_2D(
...
-                    ptr::copy_nonoverlapping(self.data.ptr(), &mut res[0][0], $SZRows * $SZCols);
+                let mut res = mem::MaybeUninit::<[[N; $SZRows]; $SZCols]>::uninit();
 
-                    res
-                }
+                unsafe { ptr::copy_nonoverlapping(self.data.ptr(), res.as_mut_ptr() as *mut N, $SZRows * $SZCols); }
+                
+                let res: [[N; $SZRows]; $SZCols] = unsafe { res.assume_init() };
+                
+                res
```
josephlr, line 195: Same nits here

src/base/conversion.rs
```diff
@@ -118,12 +118,13 @@ macro_rules! impl_from_into_asref_1D(
...
             #[inline]
             fn into(self) -> [N; $SZ] {
-                unsafe {
-                    let mut res: [N; $SZ] = mem::uninitialized();
-                    ptr::copy_nonoverlapping(self.data.ptr(), &mut res[0], $SZ);
+                let mut res = mem::MaybeUninit::<[N; $SZ]>::uninit();
 
-                    res
-                }
+                unsafe { ptr::copy_nonoverlapping(self.data.ptr(), res.as_mut_ptr() as *mut N, $SZ); }
```
josephlr, line 123: Nit: Rustfmt usually does
```rust
unsafe { ptr::copy_nonoverlapping(self.data.ptr(), res.as_mut_ptr() as *mut N, $SZ) };
```
with the `}` before the `;`.

src/base/conversion.rs
```diff
@@ -118,12 +118,13 @@ macro_rules! impl_from_into_asref_1D(
...
+
+                let res: [N; $SZ] = unsafe { res.assume_init() };
+                
+                res
```
josephlr, line 127: Nit: this can just be `unsafe { res.assume_init() }`. No need for the assignment to `res`.

src/base/conversion.rs
ProfFan: Thanks :)

sebcrozet:
Thanks!

PR #706, opened 2020-03-13
Unnecessary Trait bound DimSub<Dynamic> in impl Cholesky

dayrover:
Looks like the bound is unnecessary as there is no use of DimDiff in the implementation.
Integration tests are OK with it replaced by a simple Dim


sebcrozet:
Thanks!

PR #708, opened 2020-04-01
Addition of matrix exponent for static size matrices.

fredrik-jansson-se:
I've implemented matrix exponent for static size matrices, happy to get feedback on improvements, e.g. for dynamic sized matrices.


sebcrozet:
Hi @fredrik-jansson-se!
Thank you for this PR. I'll review this next week-end.

src/linalg/exp.rs
```diff
@@ -0,0 +1,479 @@
...
+            d4_approx: None,
+            d6_approx: None,
+            d8_approx: None,
+            d10_approx: None,
+        }
+    }
+
+    fn a2(&self) -> &MatrixN<N, R> {
+        if self.a2.is_none() {
+            let ap = &self.a2 as *const Option<MatrixN<N, R>> as *mut Option<MatrixN<N, R>>;
```
sebcrozet, line 69: Casting a non-mutable reference to a to a mutable pointer is always UB. See https://doc.rust-lang.org/nomicon/transmutes.html.
You can either use `std::cell::Cell` or `std::cell::RefCell` instead of this unsafe cast.

What is the reason why you can't take `&mut self` as parameter here?

src/linalg/exp.rs
```diff
@@ -0,0 +1,479 @@
...
+
+        self.d10_approx.unwrap()
+    }
+
+    fn pade3(&mut self) -> (MatrixN<N, R>, MatrixN<N, R>) {
+        let b = [
+            N::from_f64(120.0).unwrap(),
+            N::from_f64(60.0).unwrap(),
+            N::from_f64(12.0).unwrap(),
+            N::from_f64(1.0).unwrap(),
```
sebcrozet, line 221: You should use `crate::convert(120.0)` instead of `N::from_f64` since this is the way we do this everywhere else in nalgebra.

src/linalg/exp.rs
```diff
@@ -0,0 +1,479 @@
...
+    base::{
+        allocator::Allocator,
+        dimension::{DimMin, DimMinimum, DimName},
+        DefaultAllocator,
+    },
+    try_convert, ComplexField, MatrixN, RealField,
+};
+
+// https://github.com/scipy/scipy/blob/c1372d8aa90a73d8a52f135529293ff4edb98fc8/scipy/sparse/linalg/matfuncs.py
+struct ExpmPadeHelper<N, R>
```
sebcrozet, line 13: Whenever we have a square matrix, we use `D` as the dimension type parameter in nalgebra. So every occurrence of `R` should be replaced by `D` in this file.

src/linalg/exp.rs
fredrik-jansson-se: Now fixed!

src/linalg/exp.rs
fredrik-jansson-se: Changed everywhere

src/linalg/exp.rs
fredrik-jansson-se: Done!

fredrik-jansson-se:
Hey, thanks for reviewing. I've now made all changes requested.

sebcrozet:
That's great! Thanks!

PR #711, opened 2020-04-03
Supplement the document of base/matrix

cschenio:
- Supplement the document of fn vector_to_matrix_index
- Supplement the document of fn as_ptr


sebcrozet:
Thanks!

Issue #719, opened 2020-04-15
Closed{Add,Sub} trait-bound is unnecessarily strict for Point::{add,sub}


z33ky:
As background for this issue, I've implemented two newtypes `Coordinate` and `Distance` to wrap an `f32` and use these as basis for nalgebra's `Point` and `Vector` types:
```
use derive_more::{Add, AddAssign, Display};

#[derive(Clone, Copy, Debug, Display, PartialEq, PartialOrd)]
pub struct Coordinate(pub f32);

#[derive(Add, AddAssign, Clone, Copy, Debug, Display, PartialEq, PartialOrd)]
pub struct Distance(pub f32);

impl std::ops::Mul<f32> for Distance { ... }
impl std::ops::MulAssign<f32> for Distance { ... }

pub use type Point = nalgebra::Point2<Coordinate>;
pub use type Vector = nalgebra::Vector2<Distance>;
```
I have then encountered that subtracting two `Point`s does not work.

I think requiring `N: ClosedSub` for `impl Sub<Point<N, D>> for Point<N, D>` is too strict.
`ClosedSub` requires that `N: Sub<N, Output=N> + SubAssign<N>`. I see no reason why `SubAssign` is required.
Further, the `Output=N` for `Sub` seems unnecessarily restrictive. The `Sub`-impl returns `VectorSum<N, D, D>`, which would require that bound, however if it returned `VectorSum<Sub<N>::Output, D, D> (where Sub<N>::Output: Scalar)` then `Output` would be allowed to differ from `N`.

I haven't gotten through the generics for `Point::add`, but adding my `Vector` to a `Point` also fails, so I'm assuming that `Point::add` has a similar situation and could be resolved similarly.

The same restriction seems to apply to other types too, e.g. `Vector`, though that is currently not an issue for me.

sebcrozet:
Hi!

Yes, nalgebra uses `Closed{Sub, Add, Mul, Div}` almost systematically instead of just `Sub/Add/Mul/Div`. The main reason is to simplify our code base. In particular, allowing an `Output` different than `N` itself would make some impls more verbose and difficult.

Relaxing this restriction whenever it is not too difficult still represent quite a lot of work though so any help is welcome, even if only a few methods are addressed.

z33ky:
I'm starting to remove the need for `{Add,Sub}Assign` for `impl {Add,Sub} for Matrix`.
[...]

PR #728, opened 2020-05-01
Relax constraints for some operations

z33ky:
Many operations expected the numeric type `N` to implement `ClosedAdd` and/or `ClosedSub`.
This commit replaces some of these with the less restrictive `Simple{Add,Sub}`, which does not require `{Add,Sub}Assign`.

This partially addresses #719.


src/base/ops.rs
```diff
@@ -368,11 +405,19 @@ macro_rules! componentwise_binop_impl(
     }
 );
 
-componentwise_binop_impl!(Add, add, ClosedAdd;
-                          AddAssign, add_assign, add_assign_statically_unchecked, add_assign_statically_unchecked_mut;
+/// Trait __alias__ for `Add` with result of type `Self`.
+pub trait SimpleAdd<Right = Self>: Sized + Add<Right, Output = Self> {}
```
z33ky, line 409: Should these traits be pushed to [simba](https://github.com/rustsim/simba)?

sebcrozet:
Hi! Thank you for this PR!

What this PR demonstrates is that we are missing a trait in `simba`, and this trait is what you called `Simple{Add,Sub}`. And I think I did a mistake in the naming of the `Closed{Add,Sub}` trait. Instead simba should have:
- A `Closed{Add,Sub}` trait which is defined the same way as your `Simple{Add,Sub}` traits.
- A `Closed{Add,Sub}Assign` trait which is what simba currently calls `Closed{Add,Sub}`.

These naming would be more in line with the spirit of the std and num traits.
This mistake is unfortunate because renaming the existing `ClosedAdd` to `ClosedAddAssign` is quite a significant breaking change. Though I'm OK with this kind of breaking change since simba did not reach 1.0 yet.

Would you like to open a PR in simba which:
1. Renames `Closed(.*)` to `Closed(.*)Assign`.
2. Adds new `Closed(.*)` traits which are what your `Simple(.*)` are in this PR.

Then to update this PR so it works with the modified simba, it would be possible to:
1. Replace all occurrence of `Closed(.*)` by `Closed(.*)Assign`.
2. Replace all occurrence of `Simple(.*)` by `Closed(.*)`.

z33ky:
I was thinking that these traits maybe should reside in `simba` myself. To test if what I'm doing works out at all I just put them in this code-base; moving them is simple enough :)

tests/geometry/point.rs
```diff
@@ -93,6 +93,94 @@ fn to_homogeneous() {
     assert_eq!(a.to_homogeneous(), expected);
 }
 
+#[test]
+fn point_ops_without_assign() {
+    // Ensure adding matrices works without implementing AddAssign
```
z33ky, line 98: This should say "points", not "matrices".

Issue #672, opened 2019-11-02
FullPivLU and QR refuse to solve for non square matrix


nestordemeure:
The [doc](https://www.nalgebra.org/decompositions_and_lapack/) says that LU will only solve for square matrix but that both FullPivLU and QR should work for non-square matrices.

In practice there is an assert in both [implementations of solve](https://docs.rs/nalgebra/0.19.0/src/nalgebra/linalg/full_piv_lu.rs.html#186-211) to insures the methods fails on non-square matrices.

Is it because the implementation of FullPivLU and QR is not able to deal with non-square matrix yet or is it an error (in which case I can submit a PR) ?

sebcrozet:
It appears the documentation is not accurate. While `LU`, `FullPivLU` and `QR` can all be computed for non-square matrices, all their methods related to inverse/solve are only implemented for square matrices.

nestordemeure:
This means that, if I am not mistaken, the only way to currenlty solve for a non-square matrix is to use a SVD decomposition.

It feels problematic to have only one costly algorithm for this use case.

Andlon:
@nestordemeure: Can you elaborate on what exactly you mean by solving a non-square system? Do you mean in a least squares sense?

I believe that if you want to use e.g. `QR` decomposition or `LU` decomposition to solve least squares problem, you need certain guarantees about the factors that you get back. For example, if I remember correctly, for the `QR` decomposition you would need a rank-revealing QR decomposition. The reason is that the `R` factor is also non-square, and so the usual back-substitution procedure used for solving linear systems after factorization is no longer directly applicable. Only if the `R` factor has certain structure (which a rank-revealing QR decomposition would give) would you be able to actually easily solve the (non-square) upper triangular problem.

To summarize, if you want to solve non-square systems a specialized algorithm for this purpose may be necessary.

nestordemeure:
Yes, I mean it in the least square sense.

Reading the documentation I believed that the method implemented deals with non-squares matrix (which might require using one of two underlying implementations as, indeed, the algorithm is different). As sebcrozet said, it is an inacuracy in the documentation.

On a larger scale, it appears that there is currently only one algorithm to solve non-square linear systems in nalgebra and, to the best of my knowledge, it is a costly one. That feels like a hole in the API.

We could probably take inspiration from Eigen and implement some of the algorithms from the [Solving linear least squares systems
[...]

PR #730, opened 2020-05-09
Change the QR implementation, inspired by LAPACK

jannschu:
I ported some ideas from LAPACK's QR implementation, more specifically from `?ORG2R`.

As a result I could increase all epsilons in the test by two order of magnitudes. Also the _full_ computation of the `Q` matrix is now supported, not only the first `min(n, m)` columns. I added a new method `q_columns` for this. The current `q()` method stayed the same.

The main difference in the implementation seems to be that LAPACK applies the scaling of the Householder vectors at a different point in the code.

# Breaking changes

- The internal representation has changed (is exposed by hidden method `qr_internal`)
- For some inputs some column signs flipped of `R` and `Q`.
- Some allocations were added, can be found when searching for `work`. This resulted in additional allocator bounds. See discussion below.

# Things to discuss

- Currently `q_tr_mul` borrows self mutably due to an implementation detail although the struct's state will only be changed temporarily inside the method. Ideas to resolve this:
  - Keep as is
  - Use interior mutability
  - Allocate
  - Change affected `gerc` and `gemv` call (possible performance penalty) 
- This would be a good occasion to implement #672  for QR
- The additional `work` allocations could be done only once by storing a suitable vector in the struct and sharing it. This will result in `&mut self` or interior mutability.


jannschu:
I rebased this and

- added CHANGELOG entry
- update svd for sign change

The TODOs can now be addressed, was not possible 6 years ago :D, will look into these.

PR #741, opened 2020-06-23
More verbose DMatrix dim asserts where possible.

jenanwise:
Previously, most dimension mismatch asserts used raw `assert!` and did not include the mismatching dimensions in the panic message. When using dynamic matrices, this led to somewhat-opaque panics such as:

```rust
let m1 = DMatrix::<f32>::zeros(2, 3);
let m2 = DMatrix::<f32>::zeros(5, 10);
m1 + m2 // panic: Matrix addition/subtraction dimensions mismatch.
```

This patch adds dimension information in the panic messages wherever doing so did not add additional bounds checks, mostly by simply changing `assert!(a == b, ...)` cases to `assert_eq!`. After:

```rust
// panic: assertion failed: `(left == right)`
//   left: `(2, 3)`,
//  right: `(5, 10)`: Matrix addition/subtraction dimensions mismatch.
```

Note that the `gemv` and `ger` were not updated, as they are called from within other functions on subset matricies -- e.g., `gemv` is called from `gemm` which is called from `mul_to` . Including dimension information in the `gemv` panic messages would be confusing to `mul` / `mul_to` users, because it would include dimensions of the column vectors that `gemm` passes to `gemv` rather than of the original `mul` arguments. A fix would be to add bounds checks to `mul_to`, but that may have performance and redundancy implications, so is left to another patch.

### DIFF
--- a/src/base/blas.rs
+++ b/src/base/blas.rs
@@ -284,7 +284,9 @@ where
     {
         assert!(
             self.nrows() == rhs.nrows(),
-            "Dot product dimensions mismatch."
+            "Dot product dimensions mismatch for shapes {:?} and {:?}: left rows != right rows.",
+            self.shape(),
+            rhs.shape(),
         );
 
         // So we do some special cases for common fixed-size vectors of dimension lower than 8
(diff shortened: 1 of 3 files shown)


sebcrozet:
Thanks!

PR #742, opened 2020-06-23
Empty rustfmt.toml.

jenanwise:
The CI rustfmt checker uses `stable` rust, not `nightly`, but the config options in `rustfmt.toml` only work on nightly -- on stable, there are ignored with warnings. This means that devs running stable get a lot of warnings and devs running nightly will accidentally format in a way that the CI checker forbids. (specifically, using `where_single_line`). E.g.:

```
$ cargo +stable fmt -- --check
Warning: can't set `indent_style = Block`, unstable features are only available in nightly channel.
Warning: can't set `where_single_line = true`, unstable features are only available in nightly channel.
...etc

$ cargo +nightly fmt -- --check
-where
-    N: RealField,
-{
+where N: RealField {
...etc
```

Just removing the toml fixes this problem.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,3 +1,0 @@
-unstable_features = true
-indent_style = "Block"
-where_single_line = true


sebcrozet:
Hi! Thank you for this PR.
I think that instead of completely removing `rustfmt.toml` we should keep it but remove its content. Having an empty `rustfmt.toml` will protect us from the case where some one has another `rustfmt.toml` in some parent directory (or `$HOME`). Cf. the doc of rustfmt:

>  You can create a TOML file called rustfmt.toml or .rustfmt.toml, place it in the project or any other parent directory and it will apply the options in that file.

jenanwise:
@sebcrozet 👍 Agreed! Added back the empty config file. Please take a look.

sebcrozet:
Thanks!

PR #745, opened 2020-07-02
added new_nonuniform_scaling_wrt_point to Matrix3 & Matrix4

adamnemecek:


src/base/cg.rs
```diff
@@ -70,6 +71,17 @@ impl<N: RealField> Matrix3<N> {
     pub fn new_rotation(angle: N) -> Self {
         Rotation2::new(angle).to_homogeneous()
     }
+
+    /// Creates a new homogeneous matrix that applies a scaling factor for each dimension with respect to point.
+    ///
+    /// Can be used to implement "zoom_to" functionality.
+    #[inline]
+    pub fn new_nonuniform_scaling_wrt_point(scaling: Vector2<N>, pt: Point2<N>) -> Self {
```
sebcrozet, line 79: Arguments should be provided by-ref to be coherent with the remaining of the cg api.

```suggestion
    pub fn new_nonuniform_scaling_wrt_point(scaling: &Vector2<N>, pt: &Point2<N>) -> Self {
```

src/base/cg.rs
```diff
@@ -90,6 +102,17 @@ impl<N: RealField> Matrix4<N> {
         Isometry::rotation_wrt_point(rot, pt).to_homogeneous()
     }
 
+    /// Creates a new homogeneous matrix that applies a scaling factor for each dimension with respect to point.
+    ///
+    /// Can be used to implement "zoom_to" functionality.
+    #[inline]
+    pub fn new_nonuniform_scaling_wrt_point(scaling: Vector3<N>, pt: Point3<N>) -> Self {
```
sebcrozet, line 109:
```suggestion
    pub fn new_nonuniform_scaling_wrt_point(scaling: &Vector3<N>, pt: &Point3<N>) -> Self {
```

adamnemecek:
Note that the formatting of the arguments passed to matrix::new is from cargo fmt.

src/base/cg.rs
adamnemecek: Why does matrix3::new_rotation (and others) take the argument by value?

sebcrozet: You are right. This looks inconsistent. We might need to revisit this at some point in the future, and settle in one conversion.

sebcrozet:
Thanks!

PR #746, opened 2020-07-06
Why Option<_>

CGMossa:
Tried to add a comment on why this is returns an Option.

### DIFF
--- a/src/geometry/quaternion_construction.rs
+++ b/src/geometry/quaternion_construction.rs
@@ -374,7 +374,8 @@ where
     }
 
     /// The unit quaternion needed to make `a` and `b` be collinear and point toward the same
-    /// direction.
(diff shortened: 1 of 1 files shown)


src/geometry/quaternion_construction.rs
```diff
@@ -374,7 +374,8 @@ where
     }
 
     /// The unit quaternion needed to make `a` and `b` be collinear and point toward the same
-    /// direction.
+    /// direction. Returns `None` if both `a` and `b points in opposite direction, as then the
```
sebcrozet, line 377:
```suggestion
    /// direction. Returns `None` if both `a` and `b` are collinear and point to opposite directions, as then the
```
Using `point to opposite directions` alone may be a bit ambiguous as some people may think this is the case as soon as their dot product is negative.

Issue #697, opened 2020-03-01
Create Quaternion from slice


dakom:
Currently there is no Quaternion<f64>::from_slice(). @sebcrozet suggested to open an issue/PR as a starting point to adding it :)

khusnetdinov:
@sebcrozet As I see this is not an actual issue, In https://github.com/dimforge/nalgebra/pull/752 you reference to https://github.com/dimforge/nalgebra/pull/780 PR that merged.

PR #752, opened 2020-07-22
Add Quaternion::from_slice

KaczuH:
It's my try to tackle #697 

I'm relatively inexperienced in Rust. All feedback will be appreciated.


sebcrozet:
Thank you for this PR @KaczuH! I will close it since I recently added something similar in #780 with the `impl From<[f32; 4]> for Quaternion`, which is more idiomatic.

By the way, keep in mind that `&[f32; 4]` is not a slice. It is a reference to an array. A slice would be a `&[f32]`.

KaczuH:
@sebcrozet Thanks a lot for the suggestions!

PR #754, opened 2020-07-31
Update version in docs header

TheZoq2:
Looks like the docs are pointing to an outdated version, this is a fix for that


sebcrozet:
Thanks!

PR #756, opened 2020-08-03
Remove unnecessary Trait bound DimSub<Dynamic> in fn cholesky(self)

dayrover:
Follows on from #706 
The bound is unnecessary as there is no use of DimDiff in the implementation.
Integration tests are OK with it replaced by a simple Dim


sebcrozet:
Thanks!

PR #759, opened 2020-08-19
Add methods for in-place reshaping of matrices

neachdainn:
Based on the conversation on Discord, this allows for the reshaping of matrices in a way that guarantees that the elements won't be moved or copied.

Additionally, there are a few fixes for typos that I noticed while working on this.

### DIFF
--- a/examples/matrix_construction.rs
+++ b/examples/matrix_construction.rs
@@ -1,6 +1,6 @@
 extern crate nalgebra as na;
 
-use na::{DMatrix, Matrix2x3, RowVector3, Vector2};
+use na::{DMatrix, Matrix2x3, Matrix3x2, RowVector3, Vector2};
 
 fn main() {
     // All the following matrices are equal but constructed in different ways.
--- a/examples/reshaping.rs
+++ b/examples/reshaping.rs
@@ -1,0 +1,33 @@
+extern crate nalgebra as na;
+
+use na::{DMatrix, Dynamic, Matrix2x3, Matrix3x2, U2, U3};
+
+fn main() {
+    // Matrices can be reshaped in-place without moving or copying values.
+    let m1 = Matrix2x3::new(1.1, 1.2, 1.3, 2.1, 2.2, 2.3);
+    let m2 = Matrix3x2::new(1.1, 2.2, 2.1, 1.3, 1.2, 2.3);
+
+    let m3 = m1.reshape_generic(U3, U2);
+    assert_eq!(m3, m2);
+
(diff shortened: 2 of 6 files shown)


neachdainn:
I also have no idea if `master` or `dev` is the correct branch to merge this.

Andlon:
That's certainly an addition I've been wanting for many times. Unfortunately I am very pressed on time right now and don't have time to do the most thorough review, but just a few comments:

- Maybe the docs could be a bit more specific about *how* the reshaping works (i.e. it reshapes in a column-major sense).
- I'm not sure if maybe there should be some actual unit tests as opposed to just an example...? I'll leave that up to @sebcrozet though.
- Bonus: You *could* add [compile-fail doctests](https://doc.rust-lang.org/edition-guide/rust-2018/rustdoc/documentation-tests-can-now-compile-fail.html) to make sure some of those commented out lines in the example actually fail to compile

As for `master`/`dev`, I believe new work is merged into `dev` first, and @sebcrozet syncs `master` before releases, so you should probably switch to merging into `dev` :)

neachdainn:
> Maybe the docs could be a bit more specific about how the reshaping works (i.e. it reshapes in a column-major sense).

Absolutely. I figured that anyone reshaping matrices probably has some idea as to how the matrices are represented, but extra documentation isn't harmful. If there is any indication that this is likely to be merged, I'll do some more documentation.

> I'm not sure if maybe there should be some actual unit tests as opposed to just an example...?

> You could add compile-fail doctests to make sure some of those commented out lines in the example actually fail to compile

I agree that these should be added. I didn't see that there was a "tests" directory, so I assumed that tests weren't a thing and didn't include them. I'll work on adding those with the additional documentation.

 >  so you should probably switch to merging into `dev`

There were conflicts, so I'll have to get around to resolving those.

neachdainn:
I might have messed something up. Did `rustsim` get changed to `dimforge` or did I switch the merge request to a random person's fork?

neachdainn:
It also might be worth adding a `reshape_row_major_generic` function. It could be that people will commonly expect row-major reshaping and supplying such a method would be significantly more efficient.

neachdainn:
Force push to correct the weird Git history.

sebcrozet:
Thank you for this PR @neachdainn!

Since this is a rather old PR, I took the liberty of:
- Rebasing it.
- Add more details on the documentation of `reshape_generic`.
- Adding a doc-test.

I thought that manually merging this would automatically close this PR, but it does seems to be the case. So I'll close it now.

PR #766, opened 2020-09-26
Add UDU factorization

### LINKED ISSUE #762: Implement a UDU factorization for square matrices
A UDU factorization is more stable than an LU decomposition when tackling covariance matrices, as is done in in [nyx-space](https://crates.io/crates/nyx-space), a high-fidelity astrodynamics toolkit. It is in fact [recommended](https://ntrs.nasa.gov/citations/20180003657) for spacecraft navigation by NASA.

## Implementation
I would be thrilled to implement this! @sebcrozet would the best way for me to implement this to start with the [LU decomposition code architecture](https://www.nalgebra.org/rustdoc/src/nalgebra/linalg/lu.rs.html#31-37) and would that source fit in the linalg module?

## Algorithm 
udu

ChristopherRabotin:
Closes #762 

Concerns:
1. The structure is defined as `UDU<N: ComplexField, D: DimName>` instead of `UDU<N: ComplexField, D: Dim>`. That's because I could not find a way to initialize a zero matrix with `D: Dim`. How can I fix that?
2. The diagonal matrix is currently stored as `Matrix<N, D>` but it could easily be a `Vector<N, D>`. Is that preferable? It's an easy change.
3. On line 51 and 58 of src/linalg/udu.rs, the compiler did not accept an AddAssign operation, so I had to resort to an Add (e.g. `d[(j, j)] = d[(j, j)] + ...`). Is there a way I can specify in the `N` constraint that it needs to allow for AddAssign, and is that desired?

### DIFF
--- a/src/linalg/udu.rs
+++ b/src/linalg/udu.rs
@@ -30,37 +30,40 @@ impl<N: ComplexField, D: DimName> UDU<N, D>
 where
     DefaultAllocator: Allocator<N, D, D>,
 {
-    /// Computes the UDU^T factorization as per NASA's "Navigation Filter Best Practices", NTRS document ID 20180003657
-    /// section 7.2 page 64.
-    /// NOTE: The provided matrix MUST be symmetric.
-    pub fn new(matrix: MatrixN<N, D>) -> Self {
+    /// Computes the UDU^T factorization
+    /// NOTE: The provided matrix MUST be symmetric, and no verification is done in this regard.
+    /// Ref.: "Optimal control and estimation-Dover Publications", Robert F. Stengel, (1994) page 360
+    pub fn new(p: MatrixN<N, D>) -> Self {
         let mut d = MatrixN::<N, D>::zeros();
         let mut u = MatrixN::<N, D>::zeros();
 
-        let n = matrix.ncols();
+        let n = p.ncols() - 1;
 
-        d[(n, n)] = matrix[(n, n)];
+        d[(n, n)] = p[(n, n)];
         u[(n, n)] = N::one();
 
-        for j in (0..n - 1).rev() {
-            u[(j, n)] = matrix[(j, n)] / d[(n, n)];
+        for j in (0..n).rev() {
+            u[(j, n)] = p[(j, n)] / d[(n, n)];
         }
 
-        for j in (1..n).rev() {
-            d[(j, j)] = matrix[(j, j)];
-            for k in (j + 1..n).rev() {
+        for j in (0..n).rev() {
+            for k in j + 1..=n {
                 d[(j, j)] = d[(j, j)] + d[(k, k)] * u[(j, k)].powi(2);
             }
 
-            u[(j, j)] = N::one();
+            d[(j, j)] = p[(j, j)] - d[(j, j)];
 
-            for i in (0..j - 1).rev() {
-                u[(i, j)] = matrix[(i, j)];
-                for k in j + 1..n {
-                    u[(i, j)] = u[(i, j)] + d[(k, k)] * u[(i, k)] * u[(j, k)];
+            for i in (0..=j).rev() {
+                for k in j + 1..=n {
+                    u[(i, j)] = u[(i, j)] + d[(k, k)] * u[(j, k)] * u[(i, k)];
                 }
+
+                u[(i, j)] = p[(i, j)] - u[(i, j)];
+
                 u[(i, j)] /= d[(j, j)];
             }
+
+            u[(j, j)] = N::one();
         }
 
         Self { u, d }
--- a/src/linalg/udu.rs
+++ b/src/linalg/udu.rs
@@ -3,13 +3,13 @@ use serde::{Deserialize, Serialize};
 
 use crate::allocator::Allocator;
 use crate::base::{DefaultAllocator, MatrixN};
-use crate::dimension::DimName;
+use crate::dimension::Dim;
 use simba::scalar::ComplexField;
 
 /// UDU factorization
 #[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
 #[derive(Clone, Debug)]
-pub struct UDU<N: ComplexField, D: DimName>
+pub struct UDU<N: ComplexField, D: Dim>
 where
     DefaultAllocator: Allocator<N, D, D>,
 {
@@ -19,14 +19,14 @@ where
     pub d: MatrixN<N, D>,
 }
 
-impl<N: ComplexField, D: DimName> Copy for UDU<N, D>
+impl<N: ComplexField, D: Dim> Copy for UDU<N, D>
 where
     DefaultAllocator: Allocator<N, D, D>,
     MatrixN<N, D>: Copy,
 {
 }
 
-impl<N: ComplexField, D: DimName> UDU<N, D>
+impl<N: ComplexField, D: Dim> UDU<N, D>
 where
     DefaultAllocator: Allocator<N, D, D>,
 {
@@ -34,10 +34,11 @@ where
     /// NOTE: The provided matrix MUST be symmetric, and no verification is done in this regard.
     /// Ref.: "Optimal control and estimation-Dover Publications", Robert F. Stengel, (1994) page 360
     pub fn new(p: MatrixN<N, D>) -> Self {
-        let mut d = MatrixN::<N, D>::zeros();
-        let mut u = MatrixN::<N, D>::zeros();
-
         let n = p.ncols();
+        let n_as_dim = D::from_usize(n);
+
+        let mut d = MatrixN::<N, D>::zeros_generic(n_as_dim, n_as_dim);
+        let mut u = MatrixN::<N, D>::zeros_generic(n_as_dim, n_as_dim);
 
         d[(n - 1, n - 1)] = p[(n - 1, n - 1)];
         u[(n - 1, n - 1)] = N::one();
--- a/src/linalg/udu.rs
+++ b/src/linalg/udu.rs
@@ -32,7 +32,7 @@ where
     DefaultAllocator: Allocator<N, D> + Allocator<N, D, D>,
 {
     /// Computes the UDU^T factorization
-    /// NOTE: The provided matrix MUST be symmetric, and no verification is done in this regard.
+    /// The input matrix `p` is assumed to be symmetric and this decomposition will only read the upper-triangular part of `p`.
     /// Ref.: "Optimal control and estimation-Dover Publications", Robert F. Stengel, (1994) page 360
     pub fn new(p: MatrixN<N, D>) -> Self {
         let n = p.ncols();
--- a/src/linalg/udu.rs
+++ b/src/linalg/udu.rs
@@ -2,7 +2,7 @@
 use serde::{Deserialize, Serialize};
 
 use crate::allocator::Allocator;
-use crate::base::{DefaultAllocator, MatrixN};
+use crate::base::{DefaultAllocator, MatrixN, VectorN, U1};
 use crate::dimension::Dim;
 use simba::scalar::ComplexField;
 
@@ -11,24 +11,25 @@ use simba::scalar::ComplexField;
 #[derive(Clone, Debug)]
 pub struct UDU<N: ComplexField, D: Dim>
 where
-    DefaultAllocator: Allocator<N, D, D>,
+    DefaultAllocator: Allocator<N, D> + Allocator<N, D, D>,
 {
     /// The upper triangular matrix resulting from the factorization
     pub u: MatrixN<N, D>,
     /// The diagonal matrix resulting from the factorization
-    pub d: MatrixN<N, D>,
+    pub d: VectorN<N, D>,
 }
 
 impl<N: ComplexField, D: Dim> Copy for UDU<N, D>
 where
-    DefaultAllocator: Allocator<N, D, D>,
+    DefaultAllocator: Allocator<N, D> + Allocator<N, D, D>,
+    VectorN<N, D>: Copy,
     MatrixN<N, D>: Copy,
 {
 }
 
 impl<N: ComplexField, D: Dim> UDU<N, D>
 where
-    DefaultAllocator: Allocator<N, D, D>,
+    DefaultAllocator: Allocator<N, D> + Allocator<N, D, D>,
 {
     /// Computes the UDU^T factorization
     /// NOTE: The provided matrix MUST be symmetric, and no verification is done in this regard.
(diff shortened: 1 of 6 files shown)


Andlon:
Great job! I only have the time to try to answer your questions, not to do a review at the moment.

> Closes #762
> 
> Concerns:
> 
>     1. The structure is defined as `UDU<N: ComplexField, D: DimName>` instead of `UDU<N: ComplexField, D: Dim>`. That's because I could not find a way to initialize a zero matrix with `D: Dim`. How can I fix that?

For  `Dim`, the dimension is not known at compile time, so `Matrix::zeros()` does not work as it does not know the dimensions. You need to use the concrete dimensions and pass it as arguments to `zeros_generic`. I think you can obtain these arguments from `matrix.data.shape()`, but maybe there's a better way. I am not sure.

>     2. The diagonal matrix is currently stored as `Matrix<N, D>` but it could easily be a `Vector<N, D>`. Is that preferable? It's an easy change.

I would recommend to go with `Vector<N, D>`, since this explicitly encodes the structure, is (much) more efficient and intuitive to work with.

>     3. On line 51 and 58 of src/linalg/udu.rs, the compiler did not accept an AddAssign operation, so I had to resort to an Add (e.g. `d[(j, j)] = d[(j, j)] + ...`). Is there a way I can specify in the `N` constraint that it needs to allow for AddAssign, and is that desired?

I know the problem. I don't think there's any way around this at present. I am not sure if the compiler is able to optimize this, because a fused add operation may not be exactly equal to the add operation, so the compiler might not be able to reorganize the operation since it would change the observable behavior of the program. Perhaps @sebcrozet (or someone else) has some thoughts on this though.

I'd like to be able to say that I would properly review your PR, but I will likely be incredibly busy in the next few weeks...

ChristopherRabotin:
Thanks for this review.

I tried switching `d` to a `Vector`, but that needed a storage (and I didn't figure out how to get that from the input matrix `p`). To start with, I switched the structure to use a generic `Matrix<N, D, D, S>` instead of a `Matrix<N, D: Dim>`. However, that makes `zero_generic` unavailable.

Here is some code in the demo implementation I use to test the algorithm. 

```rust
pub struct UDU<D: Dim, S: Storage<f64, D, D>>
where
    DefaultAllocator: Allocator<f64, D, D>,
    S: Storage<f64, D, D>,
{
    pub u: Matrix<f64, D, D, S>,
    pub d: Matrix<f64, D, D, S>,
}
```

The error is:
```
error[E0599]: no function or associated item named `zeros_generic` found for struct `na::Matrix<f64, D, D, S>` in the current scope
  --> tests/lib.rs:27:41
   |
27 |     let mut d = Matrix::<f64, D, D, S>::zeros_generic(n_as_dim, n_as_dim);
   |                                         ^^^^^^^^^^^^^
   |                                         |
   |                                         function or associated item not found in `na::Matrix<f64, D, D, S>`
   |                                         help: there is an associated function with a similar name: `rows_generic`
```

Any idea what the problem could be?

Note: the latest implementation I pushed works with all `D: Dim`, but it still stores the diagonal in a `MatrixN` instead of a `Vector<N, D>`.

Andlon:
To start with, I would recommend you to keep your original formulation with `MatrixN<N, D>`. This will ensure that you have a "correct" storage. The problem with expanding to a more general storage (besides the added API complexity) is that this also includes things like matrix views. However, you can not construct a "new" view with zero entries, this is impossible, which is why you cannot use `zeros_generic`. If you look at the [trait bounds for zeros_generic](https://docs.rs/nalgebra/0.22.0/nalgebra/base/struct.Matrix.html#impl-27), you see that it makes assumptions on the storage. By using `MatrixN` you automatically satisfy these constraints.

I think for the vector you should be able to use a `VectorN<N, D>`, but you'll need the additional bound `DefaultAllocator: Allocator<N, D>` (`U1` is implied for the column dimension). Do you run into any problems this way?

ChristopherRabotin:
That worked, thanks!

Let me know if there are other changes you'd like me to make. I added a function called `d_matrix` which returns `d` as a `MatrixN`. I imagine that this would be quite helpful for a number of applications.

I'm still not sure what is going on with the tests: any idea what the problem is and how I can fix it? I don't know enough about the  nalgebra to fix it myself right now.

lebensterben:
>That worked, thanks! ...

@ChristopherRabotin
There are other properties of the factorization we can use for test. (Surely, P = UDU' is the most important one.)
Notably this doesn't work for non-symmetric matrix. Though your implementation doesn't check for symmetry, the equality in general doesn't hold. So for non-symmtric input, the test should expect a panic (for non-equality).

ChristopherRabotin:
Hi there. Any update on this PR? I added a panic test [three weeks ago](https://github.com/dimforge/nalgebra/pull/766/commits/e24964cc813ee385cbbf649a13b798d52e782458) and was wondering if I could help in the review process, or even just getting the tests to build on Circle CI.

Thanks

sebcrozet:
@ChristopherRabotin Hi! Thank you for this PR. Sorry I did not have the time to review this so far. I will make a detailed review next week-end.

src/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
...
+        for j in (0..n - 1).rev() {
+            u[(j, n - 1)] = p[(j, n - 1)] / d[n - 1];
+        }
+
+        for j in (0..n - 1).rev() {
+            for k in j + 1..n {
+                d[j] = d[j] + d[k] * u[(j, k)].powi(2);
+            }
+
+            d[j] = p[(j, j)] - d[j];
```
sebcrozet, line 56: I suggest adding an accumulator here, to avoid the weird `d[j] = d[j] + ...`:

```suggestion
            let mut d_j = d[j];
            for k in j + 1..n {
                d_j += d[k] * u[(j, k)].powi(2);
            }

            d[j] = p[(j, j)] - d_j;
```

src/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
...
+
+            for i in (0..=j).rev() {
+                for k in j + 1..n {
+                    u[(i, j)] = u[(i, j)] + d[k] * u[(j, k)] * u[(i, k)];
+                }
+
+                u[(i, j)] = p[(i, j)] - u[(i, j)];
+
+                u[(i, j)] /= d[j];
```
sebcrozet, line 65: Same remark as for `d[j]` above.

```suggestion
                let mut u_ij = u[(i, j)];
                for k in j + 1..n {
                    u_ij += d[k] * u[(j, k)] * u[(i, k)];
                }

                u[(i, j)] = (p[(i, j)] - u_ij) / d[j];
```

src/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
...
+            for k in j + 1..n {
+                d[j] = d[j] + d[k] * u[(j, k)].powi(2);
+            }
+
+            d[j] = p[(j, j)] - d[j];
+
+            for i in (0..=j).rev() {
+                for k in j + 1..n {
+                    u[(i, j)] = u[(i, j)] + d[k] * u[(j, k)] * u[(i, k)];
+                }
```
sebcrozet, line 61: It is very unfortunate that we have to iterate on the columns here. This will be very inefficient for large matrices because our matrices are stored internally in column-major order.

src/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
...
+        d[n - 1] = p[(n - 1, n - 1)];
+        u[(n - 1, n - 1)] = N::one();
+
+        for j in (0..n - 1).rev() {
+            u[(j, n - 1)] = p[(j, n - 1)] / d[n - 1];
+        }
+
+        for j in (0..n - 1).rev() {
+            for k in j + 1..n {
+                d[j] = d[j] + d[k] * u[(j, k)].powi(2);
```
sebcrozet, line 53: Iterating on the columns of `u` is unfortunate because our matrices are stored in column-major order. Though avoiding this would probably require us to completely rethink the algorithm. I guess the LDL decomposition would not have this issue.

src/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
+#[cfg(feature = "serde-serialize")]
+use serde::{Deserialize, Serialize};
+
+use crate::allocator::Allocator;
+use crate::base::{DefaultAllocator, MatrixN, VectorN, U1};
+use crate::dimension::Dim;
+use simba::scalar::ComplexField;
+
+/// UDU factorization
+#[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
```
sebcrozet, line 10: Here are the bounds needed to make the compiler happy when the `serde-serialize` feature is enabled:
```suggestion
#[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
#[cfg_attr(
    feature = "serde-serialize",
    serde(bound(serialize = "VectorN<N, D>: Serialize, MatrixN<N, D>: Serialize"))
)]
#[cfg_attr(
    feature = "serde-serialize",
    serde(bound(deserialize = "VectorN<N, D>: Deserialize<'de>, MatrixN<N, D>: Deserialize<'de>"))
)]
```

src/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
+#[cfg(feature = "serde-serialize")]
+use serde::{Deserialize, Serialize};
+
+use crate::allocator::Allocator;
+use crate::base::{DefaultAllocator, MatrixN, VectorN, U1};
+use crate::dimension::Dim;
```
sebcrozet, line 6: This is needed for one of my other suggestions:
```suggestion
use crate::dimension::Dim;
use crate::storage::Storage;
```

tests/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
+use na::Matrix3;
+use na::UDU;
+
+#[test]
+#[rustfmt::skip]
+fn udu_simple() {
+    let m = Matrix3::new(
+        2.0, -1.0,  0.0,
+       -1.0,  2.0, -1.0,
+        0.0, -1.0,  2.0);
```
sebcrozet, line 10:
```suggestion
        0.0, -1.0,  2.0
);
```

tests/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+#[test]
+#[should_panic]
+#[rustfmt::skip]
+fn udu_non_sym_panic() {
+    let m = Matrix3::new(
+        2.0, -1.0,  0.0,
+        1.0, -2.0,  3.0,
+       -2.0,  1.0,  0.0);
```
sebcrozet, line 26:
```suggestion
       -2.0,  1.0,  0.0
);
```

src/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
...
+use simba::scalar::ComplexField;
+
+/// UDU factorization
+#[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug)]
```
sebcrozet, line 11:
```suggestion
#[derive(Clone, Debug, PartialEq)]
```

src/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
...
+
+impl<N: ComplexField, D: Dim> UDU<N, D>
+where
+    DefaultAllocator: Allocator<N, D> + Allocator<N, D, D>,
+{
+    /// Computes the UDU^T factorization
+    /// NOTE: The provided matrix MUST be symmetric, and no verification is done in this regard.
+    /// Ref.: "Optimal control and estimation-Dover Publications", Robert F. Stengel, (1994) page 360
+    pub fn new(p: MatrixN<N, D>) -> Self {
+        let n = p.ncols();
```
sebcrozet, line 38:
```suggestion
        assert_eq!(p.is_square(), "Cannot compute the UDU decomposition of a non-square matrix.");
        let n = p.ncols();
```

src/linalg/udu.rs
```diff
@@ -0,0 +1,78 @@
...
+    MatrixN<N, D>: Copy,
+{
+}
+
+impl<N: ComplexField, D: Dim> UDU<N, D>
+where
+    DefaultAllocator: Allocator<N, D> + Allocator<N, D, D>,
+{
+    /// Computes the UDU^T factorization
+    /// NOTE: The provided matrix MUST be symmetric, and no verification is done in this regard.
```
sebcrozet, line 35: I feel like this algorithm only reads the upper-triangular part of `p` (so it will still work if the strict-lower-triangular part of `p` contains garbage). Is it correct? In that case, this comment should be modified as follows:

```suggestion
    /// The input matrix `p` is assumed to be symmetric and this decomposition will only read the upper-triangular part of `p`.
```

sebcrozet (changes requested):
Thank you for this PR!

Will this actually work for complex matrices? Is UDU supposed to be applied to symmetric complex matrices (in which case the current implement should be fine) or hermitian complex matrices (in which case I would expect to see some `conjugate` in there)?

lebensterben:
> Will this actually work for complex matrices? Is UDU supposed to be applied to symmetric complex matrices (in which case the current implement should be fine) or hermitian complex matrices (in which case I would expect to see some `conjugate` in there)?

Some background here:

The UDU decomposition here is also called UD factorization or UDU factorization. It's actually a diagnolization method for symmetric matrix.

Notice that some people calls SVD as UDU decomposition, but they are not the same.

The UDU decomposition mentioned in this PR is a technique used in certain engineering fields and I have not heard of before.

From what I read in a few papers, it requires symmetric matrix, but not necessarily positive semi-definite matrix. ( Though due to the nature of its application, that it's used to factorize covariance matrix, who is mostly PSD in real-world application. )

To answer you question, I have not seen papers mentioned whether this would work for complex matrix or not.

lebensterben:
Prior arts on UD decomposition/factorization can be found here:
https://www.mathworks.com/matlabcentral/fileexchange/32537-ud-factorization-kalman-filtering

ChristopherRabotin:
I believe that I've implemented all of the changes. I can't see why the test-native has failed: circle-ci wants me to give them access to my private repos, but I won't ;-)

ChristopherRabotin:
@lebensterben is right: the UDU decomposition is mainly used for covariance analyzes to avoid rounding errors from causing the covariance to no longer be PSD. I also am not aware of any situation where the UDU decomposition is applied with complex numbers: maybe I should change the `N: ComplexField` to `N: RealField` then. What do you think?

lebensterben:
>@lebensterben is right: the UDU decomposition is mainly used for covariance analyzes to avoid rounding errors from causing the covariance to no longer be PSD. I also am not aware of any situation where the UDU decomposition is applied with complex numbers: maybe I should change the `N: ComplexField` to `N: RealField` then. What do you think?

@ChristopherRabotin I agree. let's be conservative

ChristopherRabotin:
Updated ;-)

ChristopherRabotin:
Hey there,

Anything I can do to help get this issue merged? I envision needed it by mid-March.

Thanks

sebcrozet:
@ChristopherRabotin I'm taking care of this today.

sebcrozet:
I took the liberty to:
- Rebase this PR to the head of `dev`.
- Fixed the quickcheck tests.
- Added a `Matrix::udu` method to compute the UDU decomposition (i.e. `UDU::new(m)` is the same as `m.udu()`).

Now waiting for the CI to pass before merging this.

sebcrozet (approved):

ChristopherRabotin:
Excellent! Thank you very much Sébastien!

Christopher Rabotin.

dayrover:
Thanks for all this work. I was looking into extending this so it has equivilent functionallity to Cholesky. Basically adding the rank_one_update, add and remove, solve and inverse functions.

UDU' factorizations have many applications where PSD matrices are involved. Generally they are always stored in a packed form, with D being on the diagonal and U in the strict upper triangle with implicit 1s. It seems very inefficent to use the storage with a matrix and a vector.

ChristopherRabotin:
@dayrover What do you recommend as storage instead of a matrix and a vector? I guess that one could store them as a data array `&[N]`, but I don't think that fits in the philosophy of nalgebra.

dayrover:
Hi @ChristopherRabotin sorry it is not so clear by what I mean by packed!
I simply mean packing U and D into a single MatrixN
For a 2D UDU' factorisation we have 
|1, ua|   |d1, 0| 
|0,   0|   |0, d2|
which can be effectiently stored in a single matrix
|d1, ua| 
| 0   d2| 

This makes using a UDU' factorisation just as efficient as a Cholesky factorisation but numerically advantagous.

ChristopherRabotin:
I see what you mean. I think there was some discussion on how it could be implemented in the issue itself, here: https://github.com/dimforge/nalgebra/issues/762 . I haven't used this UDU implementation [yet](https://gitlab.com/nyx-space/nyx/-/issues/164), but as soon as I do, I'll be curious to see if storing in a single `MatrixN` would be advantageous to my use case.

What use case do you currently have where you see an advantage of storing it as a single matrix?

dayrover:
Hi Christopher,
the obvious advantage is the we are not wasting storage and time accessing the additional vector.
I use UDU' factorisations myself for Kalman filtering. See https://sourceforge.net/p/bayesclasses/rustcode/ci/main/tree/src/linalg/cholesky.rs

ChristopherRabotin:
Oh cool! I might use your unscented implementation for inspiration in my Kalman filters (which I use for spacecraft navigation): https://gitlab.com/nyx-space/nyx/-/blob/master/src/od/kalman.rs .

dayrover:
Please do. It is MIT licenced. As mentioned in the comment. This implements the classic Duplex 'Unscented' transform.  This is mostly as I used to work with Julier and Uhlman and worked from the original papers. There are many variations that can be made to optimise for different non linear functions but the principle is the same.

PR #770, opened 2020-10-09
Remove unsafe statement in Point::deref by forwarding to Vector.

msmorgan:
Since both impls are #[inline], this should have no performance impact.


sebcrozet:
Thanks! These `transmute` date back from when `Point` contained an array instead of a `Vector`.

msmorgan:
@sebcrozet Could you label this `hacktoberfest-accepted`?

sebcrozet:
@msmorgan Sure! Label added.

PR #786, opened 2020-10-28
Make serde no_std compatible

chemicstry:
This removes `serde` default features (std) and allows `nalgebra` to be compiled with `serde-serialize` in a `no_std` environment.


sebcrozet:
Thanks!

PR #788, opened 2020-11-15
Docs improvements - part 1

sebcrozet:
This is the first PR of a series of documentation improvements for `nalgebra` (which I will be working on progressively during the next few weeks).

The goal here is to improve the documentation generated by rustdoc. In this first PR, I focused on:
- Reducing the aliasing depth to 1. This means that before, `Vector3` was an alias of `VectorN`, itself an alias of `MatrixMN`, which is an alias of `Matrix`. Now `Vector3` is an alias of `Matrix` directly. This makes it much easier to browser the documentation since everything is on the doc of `Matrix` anyways.
- The documentation of the `Matrix` type by moving, splitting, and merging impl blocks in order to organize them in a logical way. Then I added `impl` block documentation in order to get anchors to them, and these anchors are referenced at the top of the documentation page. It looks like this:

image

This should make it significantly easier to find a specific method for matrices and vectors.
I am not sure how this will render on `docs.rs`, so I will have to make a release to see if the `<span>` I use in the doc render properly.

### DIFF
--- a/nalgebra-glm/src/common.rs
+++ b/nalgebra-glm/src/common.rs
@@ -239,9 +239,9 @@ where
     x.map(|x| x.floor())
 }
 
-//// FIXME: should be implemented for TVec/TMat?
+//// TODO: should be implemented for TVec/TMat?
 //pub fn fma<N: Number>(a: N, b: N, c: N) -> N {
-//    // FIXME: use an actual FMA
+//    // TODO: use an actual FMA
 //    a * b + c
 //}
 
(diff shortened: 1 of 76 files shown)


filnet:
Commenting on a closed PR...

As you are working on the documentation, you might want to address the clippy warning about missing Safety sections.
There are 58 of them.

`warning: unsafe function's docs miss `# Safety` section`

sebcrozet:
Good to know. Right now I am mostly adding "sections" which are just impl blocks with a single-line comment like follows:

```
/// # My section title
impl Matrix<...> {
}
```

So I have not yet reached the stage where I am actually adding new content to the methods documentation.
I will keep this in mind when the time comes (I opened [an issue for this](https://github.com/dimforge/nalgebra/issues/791)).

PR #790, opened 2020-11-16
Fail test compilation if mandatory features are missing

Andlon:
The `nalgebra` test suite at a minimum requires the `compare` and `debug` features. However, if these features are not passed in, the compilation fails with a bunch of errors from which it's very hard to discern what features are missing.

As discussed on Discord, this simple addition will give a compile-time error message if any of these two mandatory features are missing.

### DIFF
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -1,7 +1,7 @@
 #[cfg(any(not(feature = "debug"), not(feature = "compare")))]
 compile_error!(
-    "Please enable the `debug` and `compare` features \
-                in order to compile and run the tests."
+    "Please enable the `debug` and `compare` features in order to compile and run the tests.
+     Example: `cargo test --features debug --features compare`"
 );
 
 #[cfg(feature = "abomonation-serialize")]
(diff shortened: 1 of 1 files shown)


Andlon:
Sorry about the test failure - had forgotten to run `cargo-fmt`. I decided to rebase and force push the change in order to avoid an extra commit for a 3-line change.

tests/lib.rs
```diff
@@ -1,3 +1,9 @@
+#[cfg(any(not(feature = "debug"), not(feature = "compare")))]
+compile_error!(
+    "Please enable the `debug` and `compare` features \
+                in order to compile and run the tests."
```
filnet, line 4: might as well say how to enable the two features ?

Andlon: Yeah, maybe that's a good idea. You mean an example with `cargo`? E.g.

> Please enable the `debug` and `compare` features in order to compile and run the tests.
   Example: `cargo test --features debug --features compares`.

filnet: That would do.

Andlon:
Thanks for the input, @filnet. Added the example, so should be ready to go now, barring any other concerns!

sebcrozet:
Thanks!

Issue #794, opened 2020-11-19
impl DoubleEndedIterator for {MatrixIter, MatrixIterMut}


Recmo:
Currently `vec.iter().rev()` fails because [`DoubleEndedIterator`](https://doc.rust-lang.org/stable/core/iter/trait.DoubleEndedIterator.html) is not implemented for [`MatrixIter`](https://docs.rs/nalgebra/0.23.0/nalgebra/base/iter/struct.MatrixIter.html). By a quick layman's glance at [the source](https://docs.rs/nalgebra/0.23.0/src/nalgebra/base/iter.rs.html#13-20) it seems straightforward to implement, but please tell me what I'm missing before I attempt this.

My usecase is Horner evaluating a coefficient vector. There are plenty workarounds, but `iter().rev()` seems the most natural to me.

sebcrozet:
Hi! Yes, this should be fairly easy to add.

ColonisationCaptain:
This looks like it should be closed.

Brandon-Huu:
Has this issue been resolved?

CicadaCinema:
> Has this issue been resolved?

I think so

PR #797, opened 2020-11-19
impl DoubleEndedIterator for {MatrixIter, MatrixIterMut}

Recmo:
Implements #794.

Also implements the [`FusedIterator`](https://doc.rust-lang.org/stable/core/iter/trait.FusedIterator.html) marker since it applies.

### DIFF
--- a/src/base/iter.rs
+++ b/src/base/iter.rs
@@ -125,16 +125,18 @@ macro_rules! iterator {
                         // element we want to return.
                         self.size -= 1;
 
+                        // Fetch strides
+                        let inner_stride = self.strides.0.value();
+                        let outer_stride = self.strides.1.value();
+
                         // Compute number of rows
-                        let inner_size = self.inner_end.offset_from(self.inner_ptr) as usize;
+                        // Division should be exact
+                        let inner_raw_size = self.inner_end.offset_from(self.inner_ptr) as usize;
+                        let inner_size = inner_raw_size / inner_stride;
 
                         // Compute rows and cols remaining
                         let outer_remaining = self.size / inner_size;
                         let inner_remaining = self.size % inner_size;
-
-                        // Fetch strides
-                        let inner_stride = self.strides.0.value();
-                        let outer_stride = self.strides.1.value();
 
                         // Compute pointer to last element
                         let last = self.ptr.offset(
--- a/src/base/iter.rs
+++ b/src/base/iter.rs
@@ -119,6 +119,15 @@ macro_rules! iterator {
                 self.size
             }
         }
+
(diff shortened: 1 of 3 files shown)


src/base/iter.rs
```diff
@@ -119,6 +119,43 @@ macro_rules! iterator {
...
+                        debug_assert_eq!(outer_stride % inner_stride, 0);
+                        let num_rows = outer_stride / inner_stride;
+
+                        // Compute rows and cols remaining
+                        let outer_remaining = self.size / num_rows;
+                        let inner_remaining = self.size % num_rows;
+
+                        // Compute pointer to last element
+                        let last = self.ptr.offset(
+                            (outer_remaining * outer_stride + inner_remaining * inner_stride)
```
Recmo, line 148: Benchmark on my machine shows this to be 300x slower than forward iteration. Is there a way to simplify / optimize the math here?

src/base/iter.rs
```diff
@@ -119,6 +119,43 @@ macro_rules! iterator {
...
+                        None
+                    } else {
+                        // Pre-decrement `size` such that it now counts to the
+                        // element we want to return.
+                        self.size -= 1;
+
+                        // Fetch strides
+                        let inner_stride = self.strides.0.value();
+                        let outer_stride = self.strides.1.value();
+                        debug_assert_eq!(outer_stride % inner_stride, 0);
```
Recmo, line 139: This assumption is probably false in MatrixSlice.

src/base/iter.rs
Recmo: After the `inner_size` fix it is only 7x slower, which is acceptable to me.

Recmo:
I'm not sure where the `check-fmt` failure comes from, the three errors are in files that this PR does not touch.

src/base/iter.rs
```diff
@@ -111,6 +112,44 @@ macro_rules! iterator {
...
+                unsafe {
+                    if self.size == 0 {
+                        None
+                    } else {
+                        // Pre-decrement `size` such that it now counts to the
+                        // element we want to return.
+                        self.size -= 1;
+
+                        // Compute number of rows
+                        let inner_size = self.inner_end.offset_from(self.inner_ptr) as usize;
```
Recmo, line 129: This needs to be divided by `inner_size`

sebcrozet:
Thank you for this PR! I took the liberty to rebase this on the head of `dev`.

Looks like the `cargo fmt` failures are gone now. I guess the failure was caused by the CI worker using a version of the compiler that changed some of the formats `cargo fmt` applies.

I'm waiting for the CI to complete before merging this now.

Issue #556, opened 2019-02-23
Get rid of `new_uninitialized` and `new_uninitialized_generic`.


sebcrozet:
Those two unsafe functions are used when we don't care about the initial value of a matrix/vector because they will be overwritten by the result of other operations. We use them mostly for performance reasons to avoid an unnecessary initialization step.

However, this is the most significant blocker for https://github.com/rustsim/nalgebra/issues/282. So if we ever want to support non-Copy types, this is the first thing to address in a way that don't hurt performance for Copy types.

aweinstock314:
I'm interested in working on this, as I want non-Copy scalars (#282) for matrices of gmp::Mpz for some cryptanalysis tooling that I'm writing. I've gotten the benchmarks running locally (requires PR #678) so I can measure performance impact.

It looks like currently like `new_uninitialized` is implemented (by the macros `impl_constructors` and `componentwise_constructors_impl`) in terms of `MatrixMN::new_uninitialized_generic`, which in turn calls `DefaultAllocator::allocate_uninitialized`. `DefaultAllocator::allocate_uninitialized` calls `mem::uninitialized` if the dimensions are static, and `Vec::{reserve_exact,set_len}` if either dimension is dynamic.

In order to support non-Copy scalars while still using the above scheme for Copy scalars (for performance), it seems necessary to add another dimension to this type-case (e.g. go from `{static, dynamic} x {static, dynamic}` to `{Copy, !Copy} x {static, dynamic} x {static, dynamic}` for `Allocator` implementations.

I've tried the following sketches of some trait-level approaches to allowing this branching, but neither quite work:

```rust
mod bools {
    pub enum True {}
    pub enum False {}
    pub trait TypeLevelBool {}
    impl TypeLevelBool for True {}
    impl TypeLevelBool for False {}
    pub trait TraitLevelTrue: TypeLevelBool {}
    impl TraitLevelTrue for True {}
    pub trait TraitLevelFalse: TypeLevelBool {}
    impl TraitLevelFalse for False {}
}
use bools::{TypeLevelBool, True, False};

pub unsafe trait IsUninitializable { type Value: TypeLevelBool; }
unsafe impl IsUninitializable for u8 { type Value = True; }
unsafe impl IsUninitializable for Vec<u8> { type Value = False; }
pub trait AllocatorIsh<S: IsUninitializable> {
    unsafe fn allocate_maybe_uninitialized(n: usize) -> Vec<S>;
}
[...]

PR #803, opened 2020-11-27
Propagate `mem::MaybeUninit` through the return types of `Allocator::…

aweinstock314:
…allocate_uninitialized` and `Matrix::new_uninitialized_generic`.

Most call sites still invoke UB through `assume_init`. Said call sites instead invoke `unimplemented!()` if the `no_unsound_assume_init` feature is enabled, to make it easier to gradually fix them.

Progress towards #556.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ compare = [ "matrixcompare-core" ]
 libm = [ "simba/libm" ]
 libm-force = [ "simba/libm_force" ]
 proptest-support = [ "proptest" ]
+no_unsound_assume_init = [ ]
 
 # This feature is only used for tests, and enables tests that require more time to run
 slow-tests = []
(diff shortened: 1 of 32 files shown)


sebcrozet (changes requested):
Thank you for this PR! This is a great first step. Some changes:
- The CI tests won't pass unless a `cargo fmt` is run.
- I think for this first PR, we should keep the same behavior as before, i.e., not use the `crate::zero_or_uninitialized_generic!` at all and always use `unimplemented!()` when `no_unsound_assume_init` is enabled. Perhaps define a macro similar to your `zero_or_uninitialized_generic!` but that does an `unimplemented!()` instead of the zero initialization?

aweinstock314:
It looks like I missed a few spots from only iterating over `cargo test`, I see the rest of the targets in `.circleci/config.yml` now.

sebcrozet (approved):

sebcrozet:
Thank you for this PR! I took the liberty to fix the remaining compilation errors.

PR #805, opened 2020-12-01
Release v0.23.2

sebcrozet:
In this release, we improved the documentation of some of the geometric types by applying changes similar to what we did in the version 0.23.1 for matrices.


sebcrozet:
The CI is failing with an ICE on nightly. Probably because of https://github.com/rust-lang/rust/issues/79560
I'll postpone this release until the compiler bug is fixed.

PR #810, opened 2020-12-16
Introduce DualQuaternion type

chinedufn:
This commit introduces the `DualQuaternion` type, in line with the plan
laid out in [#487].

[#487]: https://github.com/dimforge/nalgebra/issues/487

### DIFF
--- a/src/geometry/dual_quaternion.rs
+++ b/src/geometry/dual_quaternion.rs
@@ -1,0 +1,116 @@
+use crate::{Quaternion, SimdRealField};
+
+/// A dual quaternion.
+///
+/// # Indexing
+///
+/// DualQuaternions are stored as \[..real, ..dual\].
+/// Both of the quaternion components are laid out in `w, i, j, k` order.
+///
+/// ```
+/// # use nalgebra::{DualQuaternion, Quaternion};
+///
+/// let real = Quaternion::new(1.0, 2.0, 3.0, 4.0);
+/// let dual = Quaternion::new(5.0, 6.0, 7.0, 8.0);
+///
+/// let dq = DualQuaternion::from_real_and_dual(real, dual);
+/// assert_eq!(dq[0], 1.0);
+/// assert_eq!(dq[4], 5.0);
+/// assert_eq!(dq[6], 7.0);
+/// ```
+///
+/// NOTE:
+///  As of December 2020, dual quaternion support is a work in progress.
+///  If a feature that you need is missing, feel free to open an issue or a PR.
+///  See https://github.com/dimforge/nalgebra/issues/487
+#[repr(C)]
+#[derive(Debug, Default, Eq, PartialEq, Copy, Clone)]
+#[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
+pub struct DualQuaternion<N: SimdRealField> {
+    // [real(w, i, j, k), dual(w, i, j, k)]
+    pub(crate) dq: [N; 8],
+}
+
+impl<N: SimdRealField> DualQuaternion<N> {
+    /// Get the first quaternion component.
+    ///
+    /// # Example
+    /// ```
+    /// # #[macro_use] extern crate approx;
+    /// # use nalgebra::{DualQuaternion, Quaternion};
+    ///
+    /// let real = Quaternion::new(1.0, 2.0, 3.0, 4.0);
+    /// let dual = Quaternion::new(5.0, 6.0, 7.0, 8.0);
+    ///
+    /// let dq = DualQuaternion::from_real_and_dual(real, dual);
+    /// relative_eq!(dq.real(), real);
+    /// ```
+    #[inline]
+    pub fn real(&self) -> Quaternion<N> {
+        Quaternion::new(self[0], self[1], self[2], self[3])
+    }
+
+    /// Get the second quaternion component.
+    ///
+    /// # Example
+    /// ```
+    /// # #[macro_use] extern crate approx;
+    /// # use nalgebra::{DualQuaternion, Quaternion};
+    ///
+    /// let real = Quaternion::new(1.0, 2.0, 3.0, 4.0);
+    /// let dual = Quaternion::new(5.0, 6.0, 7.0, 8.0);
+    ///
+    /// let dq = DualQuaternion::from_real_and_dual(real, dual);
+    /// relative_eq!(dq.dual(), dual);
+    /// ```
+    #[inline]
+    pub fn dual(&self) -> Quaternion<N> {
+        Quaternion::new(self[4], self[5], self[6], self[7])
+    }
+}
+
+impl<N: SimdRealField> DualQuaternion<N>
+where
+    N::Element: SimdRealField,
+{
+    /// Normalizes this quaternion.
+    ///
+    /// # Example
+    /// ```
+    /// # #[macro_use] extern crate approx;
+    /// # use nalgebra::{DualQuaternion, Quaternion};
+    /// let real = Quaternion::new(1.0, 2.0, 3.0, 4.0);
+    /// let dual = Quaternion::new(5.0, 6.0, 7.0, 8.0);
+    /// let dq = DualQuaternion::from_real_and_dual(real, dual);
+    ///
+    /// let dq_normalized = dq.normalize();
+    ///
+    /// relative_eq!(dq_normalized.real().norm(), 1.0);
+    /// ```
+    #[inline]
+    #[must_use = "Did you mean to use normalize_mut()?"]
+    pub fn normalize(&self) -> Self {
+        let real_norm = self.real().norm();
+
+        Self::from_real_and_dual(self.real() / real_norm, self.dual() / real_norm)
+    }
+
+    /// Normalizes this quaternion.
+    ///
+    /// # Example
+    /// ```
+    /// # #[macro_use] extern crate approx;
+    /// # use nalgebra::{DualQuaternion, Quaternion};
+    /// let real = Quaternion::new(1.0, 2.0, 3.0, 4.0);
+    /// let dual = Quaternion::new(5.0, 6.0, 7.0, 8.0);
+    /// let mut dq = DualQuaternion::from_real_and_dual(real, dual);
+    ///
+    /// dq.normalize_mut();
+    ///
+    /// relative_eq!(dq.real().norm(), 1.0);
+    /// ```
+    #[inline]
+    pub fn normalize_mut(&mut self) {
+        *self = self.normalize();
+    }
+}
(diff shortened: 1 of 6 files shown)


src/geometry/dual_quaternion_ops.rs
```diff
@@ -0,0 +1,29 @@
...
+ * ===================
+ *
+ */
+
+use std::ops::{Mul};
+use crate::{DualQuaternion, Scalar, DefaultAllocator, U4, U1, SimdRealField};
+use simba::simd::SimdValue;
+use crate::base::allocator::Allocator;
+
+impl <N: Scalar + SimdValue + SimdRealField> Mul<DualQuaternion<N>> for DualQuaternion<N>
```
chinedufn, line 17: I added the `Scalar + SimdValue` constraint so that I could leverage the existing `impl Mul for Quaternion` (which consumes self, naturally).

Otherwise I could get a borrow checker error from trying to use `self.rot` below twice.

I'm not too familiar with all of the traits in nalgebra. Is there a better way that I should handle this?

src/geometry/dual_quaternion.rs
```diff
@@ -0,0 +1,48 @@
...
+///  See https://github.com/dimforge/nalgebra/issues/487
+#[repr(C)]
+#[derive(Debug)]
+pub struct DualQuaternion<N: Scalar + SimdValue> {
+    /// The rotation part of the dual quaternion.
+    pub rot: Quaternion<N>,
+    /// The translation part of the dual quaternion.
+    pub trans: Quaternion<N>
+}
+
```
chinedufn, line 19: Just added a few misc functions. Need to add more over time.

chinedufn (commented):
The instructions that you laid out were very helpful, thanks a lot

https://github.com/dimforge/nalgebra/issues/487#issuecomment-441528481

chinedufn:
@sebcrozet anything blocking this from landing? Happy to make any changes that you suggest.

src/geometry/dual_quaternion.rs
```diff
@@ -78,3 +81,46 @@ where
...
+#[cfg(feature = "serde-serialize")]
+impl<N: SimdRealField> Serialize for DualQuaternion<N>
+where
+    N: Serialize,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        [
```
chinedufn, line 94: Wasn't sure how to use a `VectorN` similar to how quaternions serialization works so I opted for this.

Let me know if this can be improved and I'll make the change(s).

src/geometry/dual_quaternion.rs
```diff
@@ -0,0 +1,126 @@
...
+
+impl<N: Scalar + SimdValue> Clone for DualQuaternion<N> {
+    #[inline]
+    fn clone(&self) -> Self {
+        Self {
+            rot: self.rot.clone(),
+            trans: self.trans.clone(),
+        }
+    }
+}
```
sebcrozet, line 52: I just made modifications a few minutes ago that should allow you to directly `#[derive]` all these traits.

src/geometry/dual_quaternion.rs
sebcrozet: If your rebase your PR on `dev`, you should be able to `derive(Serialize, Deserialize)` directly on `DualQuaternions`.

src/geometry/dual_quaternion.rs
```diff
@@ -0,0 +1,126 @@
...
+    ///
+    /// let dq = DualQuaternion::new(rot,trans);
+    ///
+    /// let dq_normalized = dq.normalize();
+    ///
+    /// relative_eq!(dq_normalized.rot.norm(), 1.0);
+    /// ```
+    #[inline]
+    #[must_use = "Did you mean to use normalize_mut()?"]
+    pub fn normalize(&self) -> Self {
```
sebcrozet, line 75: It would be great to add an implementation of `normalize_mut` too, since it is mentioned here.

sebcrozet (changes requested):
Hi! Thank you for this PR.

src/geometry/dual_quaternion_ops.rs
sebcrozet: This is odd. The `SimdRealField` bound should already imply `Scalar + SimdValue` so adding these two should not have made any difference.

chinedufn: Mmmm I removed them and it worked fine. I must've just been missing the `N::Element: SimdRealField` bounds when I first got that error or something.
This works now thanks.

src/geometry/dual_quaternion.rs
```diff
@@ -0,0 +1,116 @@
...
+/// NOTE:
+///  As of December 2020, dual quaternion support is a work in progress.
+///  If a feature that you need is missing, feel free to open an issue or a PR.
+///  See https://github.com/dimforge/nalgebra/issues/487
+#[repr(C)]
+#[derive(Debug, Default, Eq, PartialEq, Copy, Clone)]
+#[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
+pub struct DualQuaternion<N: SimdRealField> {
+    // [real(w, i, j, k), dual(w, i, j, k)]
+    pub(crate) dq: [N; 8],
```
chinedufn, line 31: Changed to store internally as an array (instead of storing real and dual part separately) so that we don't need an `if` statement when indexing.

chinedufn (commented):
@sebcrozet thanks for the helpful review - I've addressed all of your feedback.

src/geometry/dual_quaternion.rs
sebcrozet: I think what you did before was better since it made it more explicit that it is composed of two quaternions. And these two quaternions could be public fields for easier access.

Regarding indexing, you can simply:
- Add `#[repr(C)]` to `DualQuaternion`.
- Add an implementation of `AsRef` and `AsMut`. Something like:

 ```rust
 impl<N: SimdRealField> AsRef<[N; 8]> for DualQuaternion<N> {
     #[inline]
     fn as_ref(&self) -> &[N; 8] {
        unsafe { mem::transmute(self) }
    }
}

// Do something similar for AsMut.
```
and use these in the implementation of `Index` and `IndexMut`.

**Comment by sebcrozet** [comment]
> @sebcrozet thanks for the helpful review - I've addressed all of your feedback.

Thank you for addressing the feedbacks. Though keep in mind that it is generally better to push new commits instead of overwriting the previous one(s). That way it is easier to review your changes because I can see easily what changed since my last review.

**Comment by chinedufn** [comment]
> Though keep in mind that it is generally better to push new commits instead of overwriting the previous one(s)

Ah, I thought that the PR was small enough that it was fine. Okay great, noted for next time. Thank you

File: src/geometry/dual_quaternion.rs (hunk 6 above)
Comment (chinedufn) [reply src/geometry/dual_quaternion.rs:31]: Ah I like your suggestion better, thanks will change.

Question though - why are quaternions stored internally at [i,j,k,w]? I would've expected [w,i,j,k] given that `Quaternion::new(w,i,j,k)`.

No big deal, just curious.

Comment (sebcrozet) [reply src/geometry/dual_quaternion.rs:31]: > Question though - why are quaternions stored internally at [i,j,k,w]? I would've expected [w,i,j,k] given that Quaternion::new(w,i,j,k).

By similarity with the Eigen library, mostly.

**Comment by chinedufn** [comment]
@sebcrozet thanks again for the help, all feedback addressed.

File: src/geometry/dual_quaternion_ops.rs [hunk 7]
```diff
@@ -0,0 +1,110 @@
...
+
+    fn add(self, rhs: DualQuaternion<N>) -> Self::Output {
+        Self::from_real_and_dual(self.real + rhs.real, self.dual + rhs.dual)
+    }
+}
+
+impl<N: SimdRealField> Sub<DualQuaternion<N>> for DualQuaternion<N>
+where
+    N::Element: SimdRealField,
+    DefaultAllocator: Allocator<N, U4, U1> + Allocator<N, U4, U1>,
```
sebcrozet, line 103:
```suggestion
```

src/geometry/dual_quaternion_ops.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    fn mul(self, rhs: N) -> Self::Output {
+        Self::from_real_and_dual(self.real * rhs, self.dual * rhs)
+    }
+}
+
+impl<N: SimdRealField> Add<DualQuaternion<N>> for DualQuaternion<N>
+where
+    N::Element: SimdRealField,
+    DefaultAllocator: Allocator<N, U4, U1> + Allocator<N, U4, U1>,
```
sebcrozet, line 91:
```suggestion
```

src/geometry/dual_quaternion_ops.rs
```diff
@@ -0,0 +1,110 @@
...
+            self.real * rhs.real,
+            self.real * rhs.dual + self.dual * rhs.real,
+        )
+    }
+}
+
+impl<N: SimdRealField> Mul<N> for DualQuaternion<N>
+where
+    N::Element: SimdRealField + SimdValue,
+    DefaultAllocator: Allocator<N, U4, U1> + Allocator<N, U4, U1>,
```
sebcrozet, line 79:
```suggestion
    N::Element: SimdRealField
```

src/geometry/dual_quaternion_ops.rs
```diff
@@ -0,0 +1,110 @@
...
+    #[inline]
+    fn index_mut(&mut self, i: usize) -> &mut N {
+        &mut self.as_mut()[i]
+    }
+}
+
+impl<N: SimdRealField> Mul<DualQuaternion<N>> for DualQuaternion<N>
+where
+    N::Element: SimdRealField,
+    DefaultAllocator: Allocator<N, U4, U1> + Allocator<N, U4, U1>,
```
sebcrozet, line 64:
```suggestion
```

src/geometry/dual_quaternion_construction.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// let dq = DualQuaternion::from_real_and_dual(rot, trans);
+    /// assert_eq!(dq.real.w, 1.0);
+    /// ```
+    #[inline]
+    pub fn from_real_and_dual(real: Quaternion<N>, dual: Quaternion<N>) -> Self {
+        Self { real, dual }
+    }
+}
+
+impl<N: SimdRealField> DualQuaternion<N> {
```
sebcrozet, line 21:
```suggestion
```

sebcrozet (changes requested):
Thanks! There are a few things that can be removed here.

chinedufn:
@sebcrozet feedback addressed!

sebcrozet (approved):

sebcrozet:
Great, thanks!

Issue #830, opened 2021-02-11
Bytemuck implementations for statically sized matrixes.


iMplode-nZ:
Bytemuck: https://docs.rs/bytemuck/1.5.0/bytemuck/
This would be useful for computer graphics.

PR #831, opened 2021-02-11
Added bytemuck implementations for static storages.

iMplode-nZ:
See issue #830.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,6 @@ compare = [ "matrixcompare-core" ]
 libm = [ "simba/libm" ]
 libm-force = [ "simba/libm_force" ]
 
-
 [dependencies]
 typenum        = "1.12"
 generic-array  = "0.14"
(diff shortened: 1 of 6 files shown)


henkkuli:
I just happened to pass by and noticed this PR and took a quick look. I think the current proposal is unsound as there is no bound on the underlying type being `Zeroable` or `Pod`. In particular, if someone plugs a type containing a `NonZero` into a matrix, the type should not be `Zeroable`.

I think implementing `Zeroable` and `Pod` directly for `Matrix` is quite high in the hierarchy. Arguably we should implement the traits on `ArrayStorage` and then derive the traits for `Matrix` when the storage type implements the traits. One could even argue the traits should be implemented on `generic_array::GenericArray`. Deriving the trait from the underlying type allows the users to implement their own storage type with the traits, and have the matrix still implement the traits.

sebcrozet (changes requested):
Thank you for this PR!

I agree with the remarks from @henkkuli: we are missing a bound of the scalar type, and these traits should be implemented on the `ArrayStorage` first.

I agree with the fact that it would be even better if that was implemented by `generic-array` too, but I'm OK to just do it on `ArrayStorage` considering we will get rid of `generic-array` soon (when `min-const-generics` lands).

sebcrozet (approved):

sebcrozet:
I took the liberty to make the relevant changes. I also added an implementation of bytemuck traits for points.

Issue #834, opened 2021-02-21
Owned RowIter/ColumnIter, Matrix::into_row_iter()/into_column_iter()


mankinskin:
I am trying to move a matrix into an iterator over its rows or columns. row_iter() borrows the matrix, which breaks my usecase:
```rust
fn large_f<N, It: Iterator<Item=N>>(matrix_map: impl Fn(Matrix) -> It, mat: Matrix) { ... }
```
I need a a function that can map a matrix to an iterator for the same operations either on the matrix' rows or columns.
```rust
let mat = ...;
large_f(
    |mat| mat.row_iter(), // use rows
    mat);
large_fn(
    |mat| mat.column_iter(), // use columns
    mat);
```
But the error is:
```rust
251 |             |mat| mat.row_iter(),
    |                   ---^^^^^^^^^^^^^^^^^
    |                   |
    |                   returns a value referencing data owned by the current function
    |                   `mat` is borrowed here
```
So I would need something like `into_row_iter()` which moves out of `mat`.

sebcrozet:
Hi! Adding `into_row_iter` and `into_column_iter` sounds useful, yes.

PR #835, opened 2021-02-22
Owned row/column iterators

mankinskin:
https://github.com/dimforge/nalgebra/issues/834

This is an initial draft for this implementation. I only managed to implement `into_row_iter` and `into_column_iter` for `DMatrix` I think because I could not use `Owned<N, R, C> where R: Dim and C: Dim` (i.e. `R` and `C` were either static or dynamic) because of the implementations of `Allocator` on `DefaultAllocator`, which [defines `Owned`](https://docs.rs/nalgebra/0.24.1/nalgebra/base/storage/type.Owned.html):
```rust
type Owned<N, R, C> = <DefaultAllocator as Allocator<N, R, C>::Buffer;
```
So to use `Owned<N, R, C>`, `R` and `C` have to satisfy the trait bounds for any of the `impl Allocator<N, R, C> for DefaultAllocator`, but there is [no implementation](https://docs.rs/nalgebra/0.24.1/nalgebra/base/default_allocator/struct.DefaultAllocator.html) for `R: Dim` and `C: Dim`.

There are only
- `R = Dynamic, C: Dim` (dynamic, either)
- `R: DimName, C: DimName` (static, static)
- `R: Dim, C = Dynamic` (either, dynamic)

Maybe there is a way to implement this for `R: Dim, C: Dim`, but this is the current state.

I would be glad if someone could review this and help me improve this implementation.

### DIFF
--- a/src/base/allocator.rs
+++ b/src/base/allocator.rs
@@ -18,7 +18,7 @@ use crate::base::{DefaultAllocator, Scalar};
 /// same `Buffer` type.
 pub trait Allocator<N: Scalar, R: Dim, C: Dim = U1>: Any + Sized {
     /// The type of buffer this allocator can instanciate.
-    type Buffer: ContiguousStorageMut<N, R, C> + Clone;
+    type Buffer: ContiguousStorageMut<N, R, C> + Clone + IntoIterator<Item=N>;
 
     /// Allocates a buffer with the given number of rows and columns without initializing its content.
     unsafe fn allocate_uninitialized(nrows: R, ncols: C) -> Self::Buffer;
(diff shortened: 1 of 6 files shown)


jswrenn:
@mankinskin, this seems like a really interesting design space, so I took a stab at `into_row_iter`: https://github.com/jswrenn/nalgebra/commit/6e34a2bffe47d4f6724d4364d811808a0737396d (the same approach ought to work for `into_col_iter`). I don't think I ran into the same issue you did, but it's entirely possible that I did and just haven't noticed.

I wonder if there's some awkwardness here about the meaning of calling `into_row_iter` on a `&Matrix` versus on a `MatrixSlice`.

sebcrozet:
Thank s @jenanwise  and @jswrenn for the suggestions. I believe the suggestion of @jswrenn is a more adequate way of doing this since it will work for all the matrix types and don't require any additional trait bound on the allocator storage output.

PR #837, opened 2021-02-28
Replace quickcheck by proptest

sebcrozet:
This PR replaces all our tests based on quickcheck with tests based on proptest.
With its version 1.0, quickcheck has the following issue:
- It may generate any floats, including extremely large numbers or special values like `NaN`: https://github.com/BurntSushi/quickcheck/pull/240 These extreme values will obviously make most of our tests fail.
- There is (afaik) no clean way to restrict the range of the generated values, including for integers: https://github.com/BurntSushi/quickcheck/issues/267

These two issues makes quickcheck basically unusable by nalgebra

Proptest on the over hands allows us to specify the kind of values we want to test with very easily. @Andlon already did a significant amount of work to use proptest with nalgebra because proptest is already used in `nalgebra-sparse`. So even if quickcheck didn't have the two aforementioned issues, it would make sense to switch to proptest anyway.


sebcrozet:
For some reasons, CircleCI is now failing lots of tests randomly. The reason appear to be caused by not having enough memory.
So I'm switching to GitHub Actions instead which does not appear to have these issues.

PR #841, opened 2021-03-02
Make the rand dependency optional.

### LINKED ISSUE #840: compiling nalgebra 0.25 on no_std: getrandom 0.2.2 compile error "target is not supported"
With nalgebra 0.24 and earlier, I was building fine with `no_std`. With 0.25.0, I am getting the following compilation error ([full logs here](https://github.com/strawlab/cam-geom/pull/4/checks?check_run_id=2010265242)):

```
 error: target is not supported, for more information see: https://docs.rs/getrandom/#unsupported-targets
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.2/src/lib.rs:213:9
    |
213 | /         compile_error!("target is not supported, for more information see: \
214 | |                         https://docs.rs/getrandom/#unsupported-targets");
    | |_________________________________________________________________________^

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.2/src/lib.rs:235:5
    |
235 |     imp::getrandom_inner(dest)
    |     ^^^ use of undeclared crate or module `imp`

error: aborting due to 2 previous errors
```

The [code in question](https://github.com/strawlab/cam-geom) does not use random numbers at all. Might it be possible to move the getrandom dependency under a cargo feature flag so that `no_std` code could continue to use `nalgebra 0.25`? Or perhaps there is a better solution.

This issue came up due to a dependabot-issued pull request and our automated tests that check compilation on `no_std`. Given a previous no_std regression with simba (https://github.com/dimforge/simba/issues/14), it seems like it would be useful to add a compilation check for no_std with relevant parts of the dimforge stack.

sebcrozet:
This makes the `rand` dependency of `nalgebra` optional.
The `rand` crate may result in some difficulties when compiling `nalgebra` for `no-std` targets that are not supported by `getrandom`. So it is best to just make it optional so that people that don't need random don't have to take the extra steps to make their stack work.

This PR adds two new cargo features:
- `rand-no-std`: enables `rand` but not `rand/std`. This allows the use of `rand` on `no-std` environments supported by `getrandom`.
- `rand`: enables `rand/std`.

Fix #840

### DIFF
--- a/.github/Xargo.toml
+++ b/.github/Xargo.toml
@@ -1,2 +1,5 @@
 [target.x86_64-unknown-linux-gnu.dependencies]
 alloc = {}
+
+[target.thumbv7em-none-eabihf.dependencies]
+alloc = {}
--- a/.github/workflows/nalgebra-ci-build.yml
+++ b/.github/workflows/nalgebra-ci-build.yml
@@ -90,7 +90,11 @@ jobs:
           components: rustfmt
       - name: install xargo
         run: cp .github/Xargo.toml .; rustup component add rust-src; cargo install -f xargo;
-      - name: build
+      - name: build x86_64-unknown-linux-gnu
         run: xargo build --verbose --no-default-features --target=x86_64-unknown-linux-gnu;
-      - name: build --feature alloc
+      - name: build x86_64-unknown-linux-gnu --features rand-no-std
+        run: xargo build --verbose --no-default-features --target=x86_64-unknown-linux-gnu;
+      - name: build x86_64-unknown-linux-gnu --features alloc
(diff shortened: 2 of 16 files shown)


jpetkau:
The methods etc. enabled by this feature no longer appear in the docs on `docs.rs`, and it's hard to discover that they even exist. (I had some code that was using `new_random()` that failed on cargo update, and I only figured it out by going to Github and explicitly searching for prs with the word 'random'.)

Suggest including `rand` in the default features used for doc generation.

PR #845, opened 2021-03-06
Add conversions from/to glam types.

sebcrozet:
This PR is a bit big, because I moved around a few files to make the maintenance of third-party crate integration easier.

This addition has been requested multiple times by the community, so let's add it.
This adds `From` implementations to convert between nalgebra and glam types.
This adds two features:
- `convert-glam` to enable these implementations.
- `convert-glam-unchecked` to enable some extra implementation that may lead to unexpected results if used improperly. For example, this enables the conversion from a `glam::Mat4` to a `na::Isometry3`. This conversion will be cheap (without any check) but will lead to unexpected results if the glam matrix contains non-isometric components (like scaling for example).

Addresses https://github.com/bevyengine/bevy/issues/1227


Ralith:
Shouldn't this be covered by the existing `mint`-based conversions? Keeping this updated may be a semver/maintenance hazard, since every breaking update to glam will prompt a breaking update to nalgebra.

PR #853, opened 2021-03-29
Add type aliases for unit vectors

### LINKED ISSUE #843: Add type aliases for unit vectors
We should add type aliases like `type UnitVector3<N> = Unit<Matrix<N, U3, U1, Owned<N, U3, U1>>>;` to simplify their uses, e.g., in SIMD code: `UnitVector3::from([v1, v2, v3, v4])`.

guissalustiano:
Resolve #843


sebcrozet:
Thanks!

Issue #832, opened 2021-02-18
Implement some operators for boolean matrices


mpizenberg:
Today I wanted to do some logical operations on matrices of booleans. Ideally, it would be as simple as calling `m1 | m2 & m3` for example. My concrete use case is building the union of masks and I'm currently doing the following (with some additional asserts on sizes):
```rust
/// Merge multiple sparse matrices into one combining all sparsely selected pixels.
pub fn merge(matrices: &[DMatrix<bool>]) -> DMatrix<bool> {
    let mut merged = matrices[0].clone();
    for mat in matrices.iter().skip(1) {
        for (b_merged, b) in merged.iter_mut().zip(mat) {
            *b_merged |= b;
        }
    }
    merged
}
```

sebcrozet:
We could add component-wise boolean operators, yes.
Note however that you can currently do this:
```rust
pub fn merge(matrices: &[DMatrix<bool>]) -> DMatrix<bool> {
    let mut merged = matrices[0].clone();
    for mat in matrices.iter().skip(1) {
        merged.zip_map(mat, |a, b| a | b);
    }
    merged
}
```
If three matrices are involved, you can do `merged.zip_zip_apply(mat, |a, b, c| a | b & c)`.
There are also `zip_map` and `zip_zip_map` methods that do the same except that they return a new matrix instead of modifying `self` in-place.

mpizenberg:
Thanks, I think it's best to wait if other people need it to avoid adding stuff to nalgebra that only I use. But if others react to this issue maybe it will be useful. I mostly wanted to open that issue to report this usage and track its interest.

PR #854, opened 2021-03-29
Add BitAnd operator to Matrix

guissalustiano:
I'm starting implement bitwise operator (#832), and open this PR to get feadback.
I'm try use `componentwise_binop_impl` macro but simba don't have `ClosedBitAnd`

If is good, I finish implement BitOr and BitXor in the same way

### DIFF
--- a/src/base/ops.rs
+++ b/src/base/ops.rs
@@ -1,7 +1,7 @@
 use num::{One, Zero};
 use std::iter;
 use std::ops::{
-    Add, AddAssign, Div, DivAssign, Index, IndexMut, Mul, MulAssign, Neg, Sub, SubAssign,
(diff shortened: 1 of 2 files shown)


sebcrozet:
Thank you for this PR! It appears to be on the right track.
However, as mentioned in #832, it is not clear whether or not this is something that is worth adding to `nalgebra`. So it may be best to just wait if more people are interested in these bit-wise operators before going further with the implementation.
I tagged #832 as "needs community feedbacks" with the hope of attracting more feedbacks from the community.

pwnorbitals:
@sebcrozet Now may be a good time to take a decision for this PR and 832. It looks like the community doesn't really need those features, what do you think ?

Issue #758, opened 2020-08-17
Implement 2D matrix convolutions


mpizenberg:
Vector convolutions have already been implemented, I thought I'd make an issue to follow progress on 2D convolutions.

PR #855, opened 2021-03-30
Feat/matrix convolution

guissalustiano:
I hastily read #758 and went to implement. But I implement de [image convolution](https://en.wikipedia.org/wiki/Kernel_(image_processing)) instead of Signal Convolution.
As I already implemented, I opened this PR to discuss if it could be useful and where I would leave it (is provisionally in third party folder).


arthurguedes375:
Any progress ?

PR #859, opened 2021-04-05
Added pow functions

OfficialURL:
This PR adds a `pow` method and a `pow_mut` method which will raise a square matrix to an integer power using exponentiation by squaring. They might be useful as a more efficient shorthand for raising a general square matrix to a small power.


src/linalg/pow.rs
```diff
@@ -0,0 +1,52 @@
...
+{
+    /// Raises a matrix to an integer power using exponentiation by squares.
+    /// Returns `None` only when the matrix is non-invertible and raised to a
+    /// negative power.
+    pub fn pow<T: PrimInt + DivAssign>(&self, mut e: T) -> Option<Self> {
+        let zero = T::zero();
+
+        if e == zero {
+            let mut i = self.clone();
+            i.fill_with_identity();
```
vks: Creating a new identity matrix instead of cloning the original and overwriting it would be faster, especially for sparse matrices.

src/linalg/pow.rs
```diff
@@ -0,0 +1,52 @@
...
+        let two = T::from(2u8).unwrap();
+        let mut multiplier = acc.clone();
+
+        while e != zero {
+            if e % two == one {
+                acc *= &multiplier;
+            }
+
+            e /= two;
+            multiplier *= multiplier.clone();
```
vks, line 41: I don't know whether this gets optimized to only allocate once (instead of once per loop). Maybe it is more prudent to move the allocation outside of the loop and reuse the temporary multiplier?

It might be possible to further optimize this by only allocating space for one column (see [here](https://stackoverflow.com/a/25452096)), but this should probably be implemented as a public `square_in_place` method.

lebensterben:
The description of the PR and documentations of new interfaces is not clear enough.

Refer to the specification https://www.mathworks.com/help/matlab/ref/mpower.html for some ideas

src/linalg/pow.rs
OfficialURL: I'm currently trying to code the `square_in_place` method. I think it fits nicely in this same module. I'll add it to the pull request as soon as it's done.

OfficialURL: Actually, I think there may have been a misunderstanding. The linked Stack Overflow article suggests a method to use a single new column, but only in the case where the result is to be stored in a different place than the original matrix. If we want to avoid extra allocations, it might be a better idea to instead have a single "buffer" matrix where the square of `multiplier` gets computed into, and copy its result back into `multiplier`.

OfficialURL: Perhaps, to make things easier in the future, we could add some sort of function to add/multiply/etc. two matrices with one another, but store the result in another given matrix. This would (assuming the compiler doesn't somehow optimize it out) save one allocation over the naive `A = B * C` approach.

OfficialURL:
I updated the description and modified a few comments in the code as well.

src/linalg/pow.rs
```diff
@@ -0,0 +1,72 @@
...
+where
+    D: DimMin<D, Output = D>,
+    DefaultAllocator: Allocator<N, D, D>,
+    DefaultAllocator: Allocator<N, D>,
+{
+    /// Attempts to raise this matrix to an integral power `e` in-place. If this
+    /// matrix is non-invertible and `e` is negative, it leaves this matrix
+    /// untouched and returns `false`. Otherwise, it returns `true` and
+    /// overwrites this matrix with the result.
+    pub fn pow_mut<T: PrimInt + DivAssign>(&mut self, mut e: T) -> bool {
```
vks, line 19: Instead of returning a `bool`, it might be better to return a `Result<(), ()>` or `Option<()>` for improved semantics (`unwrap` etc.) and the `must_use` annotation.

OfficialURL: I've implemented this suggestion. `pow_mut` now returns a `Result<(), ()>`.

sebcrozet:
Thanks @OfficialURL!
Thank you @vks for the reviews.

PR #860, opened 2021-04-08
Add determinant method to Cholesky

jturner314:
This adds a `.determinant()` method to `Cholesky`, which is computed as the square of the product of the diagonal elements in the Cholesky factor. The implementation is based on `LU`'s implementation of `.determinant()`, and the tests simply compare to the determinant computed by `LU`.


sebcrozet:
Thanks!

PR #861, opened 2021-04-09
Fix Cholesky::determinant for complex elements

jturner314:
I realized that I forgot about complex element types when writing #860. I find it odd that the tests didn't catch this issue. Maybe `RandomSDP` happens to generate matrices such that the determinant of the Cholesky factor is always purely real, or maybe that's the case for all symmetric positive definite matrices?

The previous implementation was correct only for real elements. The Cholesky decomposition is `L L^H`, so the determinant is `det(L) * det(L^H)`. Since `L` is a triangular matrix, `det(L)` is the product of the diagonal elements of `L`. Since `L^H` is triangular and its diagonal elements are the conjugates of the diagonal elements of `L`, `det(L^H)` is the conjugate of `det(L)`. So, the overall determinant is the product of the diagonal elements of `L` times its conjugate.

### DIFF
--- a/src/linalg/cholesky.rs
+++ b/src/linalg/cholesky.rs
@@ -140,13 +140,13 @@ where
     }
 
     /// Computes the determinant of the decomposed matrix.
-    pub fn determinant(&self) -> N {
+    pub fn determinant(&self) -> N::SimdRealField {
         let dim = self.chol.nrows();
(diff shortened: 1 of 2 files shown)


sebcrozet:
Good catch, thanks!

PR #862, opened 2021-04-10
Better error message when running `cargo test`

vks:
The `nalgebra` tests are designed to require the "compare", "debug" and "rand" features. Due to this, running `cargo test` fails.

To hint users at this behavior, there is a custom compile error. However, this error is not shown in practice due to other compilation failures. This PR tries to rectify this by making some compilations conditional and using Cargo's `required-features` key.


sebcrozet:
Thanks!

PR #863, opened 2021-04-10
Use the new Cargo resolver from Rust 1.50

vks:
This should make it possible to add the dev-dependency on criterion
without breaking `no_std` builds.


sebcrozet:
Thanks!

PR #864, opened 2021-04-10
More idiomatic use of Rand

vks:
This include the Rand-related changes from #622. It does not include any of the reworked Cargo features or removed `extern crate ...` statements from #622.

### DIFF
--- a/src/base/construction.rs
+++ b/src/base/construction.rs
@@ -853,6 +853,7 @@ where
     }
 }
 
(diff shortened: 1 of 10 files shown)


sebcrozet:
Thanks! The CI fails because of `cargo fmt`. I have a pretty pig PR to merge after this one, so I will take care of this at the same time.

vks:
> The CI fails because of cargo fmt.

Sorry about this, I was on my Windows machine and had some trouble to run `cargo fmt` due to weird path errors.

Issue #344, opened 2018-05-24
Rotations: add constructors from euler angles with other convensions.


sebcrozet:
Currently, `::from_euler_angle(...)` applies the rotations in the following order: "roll" first, "pitch" second, and "yaw" third. We should provide a way for the user to specify the convention to be applied.

See the initial request on the [forum](http://users.nphysics.org/t/specify-rotation-order-in-from-euler-angle-r-p-y-method/194).

PR #867, opened 2021-04-12
add euler order support

IsaacTru:
These changes attempt to address #344 and uses the [wiki](https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles) article as a reference for conversion.

However, a difference between the documentation of the `from_euler_angles()` method and what is shown in the reference was discovered. The order used in the wiki page is ZYX and is self-consistent. However, the current implementation within nalgebra states that the order is 1 roll - 2 pitch - 3 yaw, but uses the same calculation as the wiki article, whose input arguments take the order yaw, pitch roll. Therefore, I believe the calculation in nalgebra is correct, but the naming of the arguments and documentation are out of sync with the actual code.

This pull request also changes the arguments to the method, so it is not backwards compatible with prior versions.

To implement the order, I used an enum with the 12 possible conventions (both classic and Tait-Bryan). The enum might be better put in a different file, but I wasn't sure about the source structure. I did the quaternion algebra by hand, so checks on its validity are greatly welcomed...

### DIFF
--- a/src/geometry/quaternion_construction.rs
+++ b/src/geometry/quaternion_construction.rs
@@ -274,6 +274,35 @@ where
         Self::new_normalize(q)
     }
 
+    /// Define the order that the Euler angles are applied. Note that these are the
+    /// combination of Proper and Tait-Bryan
+    pub enum EulerOrder {
+        /// Rotation about Z, then new X, then new Z. A classic Euler format.
+        ZXZ,
+        /// Rotation about X, then new Y, then new X. A classic Euler format.
+        XYX,
+        /// Rotation about Y, then new Z, then new Y. A classic Euler format.
+        YZY,
+        /// Rotation about Z, then new Y, then new Z. A classic Euler format.
+        ZYZ,
+        /// Rotation about X, then new Z, then new X. A classic Euler format.
+        XZX,
+        /// Rotation about Y, then new X, then new Y. A classic Euler format.
+        YXY,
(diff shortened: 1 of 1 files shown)


n3vu0r:
I would like to test this on the weekend, especially regarding the angle ranges. If I remember correctly, the current implementation assigns the pitch, yaw, roll rotation axes to the x, y, z vector components with full-intrinsic rotation order roll, pitch, yaw (z, x', y'') resulting in angle ranges ±π, ±π, ±π/2. I wonder if the current proposed changes allow a combination of the current axes assignment and rotation order but with different angle ranges of for example ±π, ±π/2, ±π, so that the pitch instead of the yaw is limited to ±π/2.

IsaacTru:
That sounds great. It definitely needs testing! Which reminds me, I didn't update the doc test, which will fail now due to the flip of the arguments. I will update them.

I think we may be using different naming conventions, or I'm misunderstanding what the wiki article is stating. as I was associating roll, pitch, yaw with (x, y', z'') if they were applied in that order, but the current implementation performs them in yaw, pitch, roll (z, y', x'') order.

I haven't looked into the angle ranges, but I'll think about that some if you want to have a discussion. Thanks!

j-hui:
Hey there! It's been a little while since this PR was opened, but I wanted to see if there's anything blocking this feature from being merged. I'm also interested in whether support for additional euler angle orders are available for `Rotation3`.

PR #868, opened 2021-04-12
Don't enable serde by default.

sebcrozet:
- Don't enable `serde/std` by default.
- Add a `serde-serialize-no-std` feature to enable serde without its default features.

At first I wantetd to rename `serde-serialize` to just `serde`, just like I did with the `rand` dependency. However this doesn't work because of the way serde serives operate: https://github.com/serde-rs/serde/issues/953


Andlon:
Just a quick interjection: it would perhaps make sense to document the difference between the two different `serde-*` features? While the name gives some of the intention away, the exact semantics are not entirely clear to me.

sebcrozet:
@Andlon Thank you for your feedback, I added some comments.

PR #871, opened 2021-04-18
Updates Makefile to build tests, benchmarks.

mookerji:
- Updates the Makefile's various target feature flag lists so that the
  targets build, similar to the CI configuration. Currently the test
  target fails because the `proptest!` feature flag is missing.
- Make non-file targets PHONY; update `make all` to build all targets.

/cc @sebcrozet


mookerji:
Some context: I ran into this error while trying to build the tests locally on nightly (`rustc 1.53.0-nightly (392ba2ba1 2021-04-17`).

Andlon:
`Makefile` no longer exists in `nalgebra`, so I believe that this PR is no longer relevant. I will therefore close it.

PR #874, opened 2021-04-27
More conversions

sebcrozet:
### Added
- Conversion from an array `[T; D]` to an isometry `Isometry<T, _, D>` (as a translation).
- Conversion from a static vector `SVector<T; D>` to an isometry `Isometry<T, _, D>` (as a translation).
- Conversion from a point `Point<T; D>` to an isometry `Isometry<T, _, D>` (as a translation).
- Conversion of an array `[T; D]` from/to a translation `Translation<T, D>`.
- Conversion of a point `Point<T, D>` to a translation `Translation<T, D>`.
- Conversion of the tuple of glam types `(Vec3, Quat)` from/to an `Isometry2` or `Isometry3`.
- Conversion of a glam type `Vec2/3/4` from/to a `Translation2/3/4`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ documented here.
 
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [0.26.2]
+###Added
+- Conversion from an array `[T; D]` to an isometry `Isometry<T, _, D>` (as a translation).
+- Conversion from a static vector `SVector<T; D>` to an isometry `Isometry<T, _, D>` (as a translation).
+- Conversion from a point `Point<T; D>` to an isometry `Isometry<T, _, D>` (as a translation).
+- Conversion of an array `[T; D]` from/to a translation `Translation<T, D>`.
+- Conversion of a point `Point<T, D>` to a translation `Translation<T, D>`.
+- Conversion of the tuple of glam types `(Vec3, Quat)` from/to an `Isometry2` or `Isometry3`.
+- Conversion of a glam type `Vec2/3/4` from/to a `Translation2/3/4`.
+
 ## [0.26.1]
 Fix a regression introduced in 0.26.0 preventing `DVector` from being serialized with `serde`.
 
--- a/src/base/conversion.rs
+++ b/src/base/conversion.rs
@@ -3,7 +3,6 @@ use alloc::vec::Vec;
 use simba::scalar::{SubsetOf, SupersetOf};
 use std::convert::{AsMut, AsRef, From, Into};
 use std::mem;
-use std::ptr;
 
 use simba::simd::{PrimitiveSimdValue, SimdValue};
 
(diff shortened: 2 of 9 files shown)


Andlon:
I feel that perhaps some of these conversions are a bit too "aggressive"? For example, `From<[[T; R]; C]>` for `SMatrix` might cause a lot of confusion about the fact that the nested array

```rust
[[1, 2],
 [3, 4]]
```
is in fact **not** the matrix
```
[ 1, 2 ]
[ 3, 4 ]
```
but rather its transpose. Similarly, is it completely obvious that a Point is directly transformed into an Isometry? An Isometry is documented in terms of a *translation*, so I suppose implementing `From` for a vector makes sense. But a point is not a translation, it's a position. Sure, it may be interpreted as such, but doesn't that defeat the purpose of having a semantic distinction between `Point` and `Vector` in the first place?

sebcrozet:
> For example, From<[[T; R]; C]> for SMatrix might cause a lot of confusion about the fact that the nested array.

We also have `From<[[T; R]; C]>` implementation for `SMatrix`. So if there is an ambiguity, the compiler should catch it and generate an error.

> Similarly, is it completely obvious that a Point is directly transformed into an Isometry? An Isometry is documented in terms of a translation, so I suppose implementing From for a vector makes sense. But a point is not a translation, it's a position. Sure, it may be interpreted as such, but doesn't that defeat the purpose of having a semantic distinction between Point and Vector in the first place?

I guess, this point conversion may sound odd. However, isometries are very often used to represent a position (especially throughout our own code-bases). So I think this makes the Points -> Isometry relevant.

sebcrozet:
Sorry, I misread your comment about `From<[[T; R]; C]>`. Yeah, I understand this can cause a confusion. However it appears that this conversion already existed (see the `impl_from_into_asref_2D` macro removed by this PR); we are just making its implementation nicer.

Andlon:
> > Similarly, is it completely obvious that a Point is directly transformed into an Isometry? An Isometry is documented in terms of a translation, so I suppose implementing From for a vector makes sense. But a point is not a translation, it's a position. Sure, it may be interpreted as such, but doesn't that defeat the purpose of having a semantic distinction between Point and Vector in the first place?
> 
> I guess this point conversion may sound odd. However, isometries are very often used to represent a position (especially throughout our own code-bases). So I think this makes the Points -> Isometry relevant.

Fair enough, I don't use isometries that often.

> Sorry, I misread your comment about `From<[[T; R]; C]>`. Yeah, I understand this can cause a confusion. However it appears that this conversion already existed before (see the `impl_from_into_asref_2D` macro partially removed by this PR); we are just making its implementation nicer here.

Ah. Hmm, well, in that case I would even suggest to remove it altogether, in favor of a method that makes the storage order explicit. I.e. `from_col_major_array` or similar.

sebcrozet:
> Ah. Hmm, well, in that case I would even suggest to remove it altogether, in favor of a method that makes the storage order explicit. I.e. from_col_major_array or similar.

Yeah, that would definitely make sense (and would allow us to mark this `from_col_major_array` as `const`).

Andlon:
As a final comment, upon further reflection I think a better name would even be `from_col_major_nested_array`, because `from_col_major_array` could also indicate that the data is laid out in a flat array.

PR #886, opened 2021-05-03
Matrix macro

### LINKED ISSUE #873: How to construct a DMatrix from a slice of slices ?
I'd like to construct a DMatrix (or a Matrix which size is auto-deduced) with the following syntax. The reason I'd like the matrix size to be auto-deduced is that I'm lazy and don't want to count the size (can be very large numbers), and the reason I'd like to use this syntax is because it's (I think) easy to deduce the matrix size from it, and would be close to being to compatible with Numpy. Of course it can be a variation of this syntax. 

**Does this exist already ? If not, are there plans to implement it ?**

Thank you for your consideration

PS : bonus points if you find the algorithm this matrix is from ;)

<details>
<summary>Code</summary>

    let a = DMatrix::from_something([
        [
            1.0 / 18.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
        ],
        [
            1.0 / 48.0,
            1.0 / 16.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
        ],
        [
            1.0 / 32.0,
            0.0,
            3.0 / 32.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
        ],
        [
            5.0 / 16.0,
            0.0,
            -75.0 / 64.0,
            75.0 / 64.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
        ],
        [
            3.0 / 80.0,
            0.0,
            0.0,
            3.0 / 16.0,
            3.0 / 20.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
        ],
        [
            29443841.0 / 614563906.0,
            0.0,
            0.0,
            77736538.0 / 692538347.0,
            -28693883.0 / 1125000000.0,
            23124283.0 / 1800000000.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
        ],
        [
            16016141.0 / 946692911.0,
            0.0,
            0.0,
            61564180.0 / 158732637.0,
            22789713.0 / 633445777.0,
            545815736.0 / 2771057229.0,
            -180193667.0 / 1043307555.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
        ],
        [
            39632708.0 / 573591083.0,
            0.0,
            0.0,
            -433636366.0 / 683701615.0,
            -421739975.0 / 2616292301.0,
            100302831.0 / 723423059.0,
            790204164.0 / 839813087.0,
            800635310.0 / 3783071287.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
        ],
        [
            246121993.0 / 1340847787.0,
            0.0,
            0.0,
            -37695042795.0 / 15268766246.0,
            -309121744.0 / 1061227803.0,
            -12992083.0 / 490766935.0,
            6005943493.0 / 2108947869.0,
            393006217.0 / 1396673457.0,
            123872331.0 / 1001029789.0,
            0.0,
            0.0,
            0.0,
            0.0,
        ],
        [
            -1028468189.0 / 846180014.0,
            0.0,
            0.0,
            8478235783.0 / 508512852.0,
            1311729495.0 / 1432422823.0,
            -10304129995.0 / 1701304382.0,
            -48777925059.0 / 3047939560.0,
            15336726248.0 / 1032824649.0,
            -45442868181.0 / 3398467696.0,
            3065993473.0 / 597172653.0,
            0.0,
            0.0,
            0.0,
        ],
        [
            185892177.0 / 718116043.0,
            0.0,
            0.0,
            -3185094517.0 / 667107341.0,
            -477755414.0 / 1098053517.0,
            -703635378.0 / 230739211.0,
            5731566787.0 / 1027545527.0,
            5232866602.0 / 850066563.0,
            -4093664535.0 / 808688257.0,
            3962137247.0 / 1805957418.0,
            65686358.0 / 487910083.0,
            0.0,
            0.0,
        ],
        [
            403863854.0 / 491063109.0,
            0.0,
            0.0,
            -5068492393.0 / 434740067.0,
            -411421997.0 / 543043805.0,
            652783627.0 / 914296604.0,
            11173962825.0 / 925320556.0,
            -13158990841.0 / 6184727034.0,
            3936647629.0 / 1978049680.0,
            -160528059.0 / 685178525.0,
            248638103.0 / 1413531060.0,
            0.0,
            0.0,
        ],
    ]);

</details>

Andlon:
This PR introduces four new macros to `nalgebra`. They are available under the (now default) `macros` feature, which makes `nalgebra` export the macros from the procedural macro crate `nalgebra-macros`. The macros are:

- `matrix![]` for constructing fixed-size matrices with a MATLAB-like syntax.
- `dmatrix![]` for constructing *dynamic* matrices with the same syntax as `matrix!`
- `vector![]` for constructing small fixed-size vectors (syntax basically the same as `vec!`)
- `dvector![]` for constructing dynamically allocated vectors with the same syntax as `vector`.

The intention here is to simplify matrix/vector construction by improving readability, ease-of-use and ensuring that the matrix construction is as fast as possible.

Example:

```rust
use nalgebra::matrix;

// Directly constructs an SMatrix<_, 2, 3>
let a = matrix![1, 2, 3;
                4, 5, 6];
```

Benefits over e.g. `MatrixRxC::new` constructors:

- Number of rows and columns are inferred from the data, so that when changing the data (e.g. increasing column count) you don't need to update both the data and the type.
- More succinct and obvious.
- Can be combined with e.g. `#[rustfmt::skip::macros(matrix)]` to prevent `rustfmt` from messing up carefully aligned syntax.
- Works for matrices of any (practical) size, not just those that have `new` constructors.
- Semi-colons clearly delineate where each row begins.

Disadvantages:

- Using the macros adds `syn` and `quote` to the dependency tree. On my 6-7 year old laptop, this adds ~10 secs to clean build compile time. Note that if you use `serde` or possibly some of the other optional dependencies, you might already have `syn`/`quote` in your dependency tree, so you pay effectively no penalty. If users don't need the macro, it's also possible to opt out by disabling the `macros` feature.
- I can't think of any others, but please let me know if I've missed any.

The user guide should probably be updated to accommodate these macros. My suggestion is to make these macros the *de-facto* way of constructing matrices with `nalgebra`.

Resolves #873.


nalgebra-macros/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "nalgebra-macros"
+version = "0.1.0"
+authors = ["Andreas Longva <andreas.b.longva@gmail.com>"]
+edition = "2018"
+
```
Andlon, line 6: `Cargo.toml` needs some work: more metadata, remove the comments. License term needs to be resolved: Currently there's a discrepancy in that Cargo.toml for `nalgebra` says BSD-3 whereas `LICENSE` is Apache 2.0.

Andlon:
I'm missing one or more tests that verify that the macros work for all built-in types. I will add this as soon as I have time.

Andlon:
I'm also missing tests where instead of just numbers, we have arbitrary Rust expressions (though one of the doc tests has something like this). I will add this later too.

Ralith:
What's the case for `vector![1, 2, 3]` existing as a distinct thing from `matrix![1; 2; 3]`?

Andlon:
> What's the case for `vector![1, 2, 3]` existing as a distinct thing from `matrix![1; 2; 3]`?

Semantics, mostly. Although a vector is also a matrix, matrices and vectors often play different roles in applications, so at least in my opinion it benefits readability to distinguish them at the syntax level (although you can indeed use `matrix!` to construct vectors as well, if you are so inclined).

Second, `dvector![1, 2, 3]` and `dmatrix![1; 2; 3]` do *not* produce the same type. In the first case, you get `DVector<T>`, in the second case `DMatrix<T>`. It seems to me that for symmetry purposes it is then natural to have `vector!` mirror `dvector!` and `matrix!` mirror `dmatrix!`.

Andlon:
All right, I finished up the things I wanted to address, so I suppose this PR should be ready now.

sebcrozet:
Hey @Andlon, this is an amazing PR. I will review this within the next few days.

> The user guide should probably be updated to accommodate these macros. My suggestion is to make these macros the de-facto way of constructing matrices with nalgebra.

I agree. I will take care of this later this month.

Andlon:
Thanks, @sebcrozet! I also wanted to add that I don't have much experience with proc macros, so if there's any experienced proc macro person with some good suggestions for improvement I'd also be happy to hear about that.

src/base/array_storage.rs
```diff
@@ -299,3 +300,14 @@ where
...
+// TODO: Where to put this impl block?
+impl<T, const R: usize, const C: usize> SMatrix<T, R, C> {
+    /// Creates a new statically-allocated matrix from the given [ArrayStorage].
+    #[inline(always)]
+    pub const fn from_array_storage(storage: ArrayStorage<T, R, C>) -> Self {
+        // This is sound because the row and column types are exactly the same as that of the
+        // storage, so there can be no mismatch
+        unsafe { Self::from_data_statically_unchecked(storage) }
+    }
+}
```
sebcrozet, line 313: I would suggest to put this on `matrix.rs`, bellow the impl block containing `from_data_statically_unchecked`.
Perhaps we should add a `TODO` comment saying that this should be removed in favor of `Matrix::from_data` once `const-fn` support generic trait bounds.

src/base/vec_storage.rs
```diff
@@ -410,3 +411,21 @@ impl<T> Extend<T> for VecStorage<T, Dynamic, U1> {
...
+}
+
+impl<T> DVector<T> {
+    /// Creates a new heap-allocated matrix from the given [VecStorage].
+    pub const fn from_vec_storage(storage: VecStorage<T, Dynamic, U1>) -> Self {
+        // This is sound because the dimensions of the matrix and the storage are guaranteed
+        // to be the same
+        unsafe { Self::from_data_statically_unchecked(storage) }
+    }
+}
```
sebcrozet, line 431: I would suggest to put this on `matrix.rs`, bellow the impl block containing `from_data_statically_unchecked`.
Perhaps we should add a `TODO` comment saying that this should be removed in favor of `Matrix::from_data` once `const-fn` support generic trait bounds.

sebcrozet (changes requested):
Thank you again for this PR. This looks great. I'm not a proc-macro expert myself so I can't really judge the quality of the implementation. But it looks OK to me.

Ralith:
The proc macro stuff LGTM.

Andlon:
@sebcrozet: I moved the impl blocks, but I needed to add some conditional compilation for the `VecStorage`-related stuff in `matrix.rs`. Not sure if this is the way you prefer it. Let me know if you want it done differently.

@Ralith: thanks for taking a look!

sebcrozet (approved):

sebcrozet:
@Andlon Looks greatk thanks!

Issue #850, opened 2021-03-13
Row-by-row version of from_iterator


RReverser:
Forgive me if this already exists, because I couldn't find it in the docs, but is there a way to construct a matrix row-by-row using an iterator? (for cases when source data structure already represents them in row-by-row form)

Currently the only way I found is creating matrix `NxM` using `from_iterator`, and then invoking `transpose` to get `MxN`, but this is rather inefficient - first, it needs to fill in matrix in incorrect order, then construct a separate matrix (`transpose_mut` fails because it says it works only on square matrices), and then walk over all the temporary data rearranging items in correct order.

sebcrozet:
Hi! You are correct, there is no direct constructor from an iterator yielding data in a row-major order. The workaround you found `Matrix::from_iterator(iter).transpose()` is probably the easiest way of doing this right now. An alternative is to collect the iterator into a `Vec` and use `Matrix::from_row_slice`.
We should add a `::from_row_iterator` constructor.

chammika-become:
@sebcrozet 
I had a go at this. First I thought this is going to be a convenience function around `Matrix::from_iterator(iter).transpose()` as hinted above. But it end up being little bit more... so I am not sure if it's the right approach. 
My changes https://github.com/dimforge/nalgebra/compare/dev...chammika-become:from_row_iterator required following.

- add `allocate_from_row_iterator` to `Allocator` trait
- Implement the this for `DefaultAllocator` variations (static x static, static x dynamic and dynamic x static)
- Use it to implement `::from_row_iterator`

Let me know if there is a simpler approach to this. Otherwise I am glad to make a PR from this.

Trolloldem:
@sebcrozet
I noticed that the changes requested in #888 are not being addressed since when they were added to the PR a couple of months ago. I had a go at implementing them. The PR walks through the steps that I've followed.

Let me know if these are all the requested changes. Otherwise I can improve from that code.

PR #888, opened 2021-05-06
Implemented from_row_iterator 

chammika-become:
PR for  #850
- added `from_row_iterator` with doc tests
- Notable difference from `from_iterator`  (due to use of `iter.take(nrows*ncols)` ) 
```rust
let dv1 = DVector::from_iterator(3, (0..100).into_iter());  // Panics : iter.collect() => "the iterator did not yield the correct number of elements"
let dv2 = DVector::from_row_iterator(3, (0..100).into_iter()); // Ok: Only take(3)

// However
let m1 = Matrix2x3::from_iterator((0..100).into_iter()); // Ok : because of zip()
let m2 = Matrix2x3::from_row_iterator((0..100).into_iter()); // Ok: as well
```
Perhaps `DVector::from_iterator` shouldn't panic when more elements are present ?

### DIFF
--- a/src/base/default_allocator.rs
+++ b/src/base/default_allocator.rs
@@ -65,6 +65,40 @@ impl<T: Scalar, const R: usize, const C: usize> Allocator<T, Const<R>, Const<C>>
 
         res
     }
+
+    #[inline]
+    fn allocate_from_row_iterator<I: IntoIterator<Item = T>>(
+        nrows: Const<R>,
+        ncols: Const<C>,
+        iter: I,
+    ) -> Self::Buffer {
+        #[cfg(feature = "no_unsound_assume_init")]
+        let mut res: Self::Buffer = unimplemented!();
+        #[cfg(not(feature = "no_unsound_assume_init"))]
+        let mut res = unsafe { Self::allocate_uninitialized(nrows, ncols).assume_init() };
+        let mut count = 0;
+        let res_ptr = res.as_mut_slice();
+
+        for (i, e) in iter
+            .into_iter()
+            .take(ncols.value() * nrows.value())
+            .enumerate()
+        {
+            unsafe {
+                *res_ptr
+                    .get_unchecked_mut((i % ncols.value()) * nrows.value() + i / ncols.value()) = e;
+            }
+            // res_ptr[(i % ncols.value()) * nrows.value() + i / ncols.value()] = e;
+            count += 1;
+        }
+
+        assert!(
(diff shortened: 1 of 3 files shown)


RReverser:
@sebcrozet Is there a chance for this (or a variation of it) to still get merged?

src/base/default_allocator.rs
```diff
@@ -96,6 +130,32 @@ impl<T: Scalar, C: Dim> Allocator<T, Dynamic, C> for DefaultAllocator {
...
+        iter: I,
+    ) -> Self::Buffer {
+        let it = iter.into_iter().take(nrows.value() * ncols.value());
+        let mut res: Vec<T> = Vec::with_capacity(nrows.value() * ncols.value());
+        let res_ptr = res.as_mut_ptr();
+        let mut count = 0;
+
+        unsafe {
+            for (i, e) in it.enumerate() {
+                *res_ptr.add((i % ncols.value()) * nrows.value() + i / ncols.value()) = e;
```
sebcrozet, line 147: This is unsound because it will run the destructor of the uninitialized entry we are overwriting.
I think this constructior can be constructed using something like this:
1. Allocate the uninitialized storage using `Self::new_uninitialized`.
2. Initialize the storage’s entries using `MaybeUninit`.
3. Call `Self::assume_init` on the storage.

src/base/default_allocator.rs
```diff
@@ -126,6 +186,32 @@ impl<T: Scalar, R: DimName> Allocator<T, R, Dynamic> for DefaultAllocator {
...
+                count += 1;
+            }
+            res.set_len(nrows.value() * ncols.value());
```
sebcrozet, line 206: Same remark as above.

sebcrozet (changes requested):
Thank you for this PR!
The implementation of `DefaultAllocator::allocate_from_row_iterator` is technically unsound because of the invalid `assume_init`:

```rust
        let mut res = unsafe { Self::allocate_uninitialized(nrows, ncols).assume_init() };
```

To get rid of this problem, we need to:
- Rebase this PR against `master`, which contains the abstractions needed to make a sound impl of this kind of constructor.
- Rewrite `allocate_from_row_iterator` in a way similar to our new `DefaultAllocator::allocate_from_iterator` impl.

RReverser:
@chammika-become Are you planning to come back to this? It would be really nice to have this method.

sebcrozet:
Closing this as it’s replaced by #1133

PR #889, opened 2021-05-06
Fix potential unsoundness after deserializing a DVector with a bad number or elements.

### LINKED ISSUE #883: Deserialize is not sound for VecStorage (and possibly others)
`VecStorage` currently looks like this:

```rust
/// A Vec-based matrix data storage. It may be dynamically-sized.
#[repr(C)]
#[derive(Eq, Debug, Clone, PartialEq)]
#[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
pub struct VecStorage<T, R: Dim, C: Dim> {
    data: Vec<T>,
    nrows: R,
    ncols: C,
}
```

`VecStorage` carries an implicit invariant in that `nrows * ncols == data.len()`, and I believe there is some unsafe code that relies on this invariant. However, since `Deserialize` is derived, it does not know about this invariant. Therefore it is possible to break this invariant in safe code by deserializing invalid data. This is a potential security issue, since a malicious attacker might potentially use this to read/write to invalid memory locations.

In order to fix this, we should write a custom `Deserialize` implementation that validates the input.

There may also be other soundness issues caused by invariants not being upheld by derived `Deserialize` implementations throughout the library.

sebcrozet:
Fix #883 and adds tests that would have triggered unsoundness.

### DIFF
--- a/src/base/vec_storage.rs
+++ b/src/base/vec_storage.rs
@@ -13,6 +13,12 @@ use crate::base::storage::{
 };
 use crate::base::{Scalar, Vector};
 
+#[cfg(feature = "serde-serialize-no-std")]
+use serde::{
+    de::{Deserialize, Deserializer, Error},
+    ser::{Serialize, Serializer},
+};
+
 #[cfg(feature = "abomonation-serialize")]
 use abomonation::Abomonation;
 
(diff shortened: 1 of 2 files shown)


Andlon (approved):
Looks good to me!

PR #892, opened 2021-05-07
Update matrixcompare version to 0.3.0

Andlon:
Fixes some warnings produced by earlier versions since Rust 1.51.
See https://github.com/Andlon/matrixcompare/pull/5 for more details.

Note that we only update dev-dependencies, so there is no API breakage.


sebcrozet:
Thanks!

PR #894, opened 2021-05-09
Fix docs on csc_data(), csc_data_mut()

dfarnham:
CSC had the CSR docs for these methods


Andlon:
Thanks for catching this! I did try my best to catch copy-paste errors for the CSR/CSC docs, but clearly I was not diligent enough!

PR #895, opened 2021-05-10
Update the glam 0.14 optional dependency.

sebcrozet:
Update `glam` to its version 0.14. I'll merge this at the end of this month since this is considered a breaking change.


sebcrozet:
This will be replaced by #900 which is much better as it allows the support of multiple version of `glam` conversion.

Issue #791, opened 2020-11-16
Fix clippy warnings for missing Safety sections in documentation


sebcrozet:
See https://github.com/dimforge/nalgebra/pull/788#issuecomment-727899092

filnet:
Some of the missing Safety sections warnings will be fixed through fixing other clippy warnings.
So please don't rush to address them...

Issue #913, opened 2021-06-16
Too many clippy warnings


OfficialURL:
Currently, running `cargo clippy` on `nalgebra` yields 581 warnings.

A few of these are false positives, and some others are just complaints about the complicated types of `nalgebra` (for which not much can be done). However, many of these should be strictly good improvements to the code. Many of these fixes would also be breaking changes, so we should be aware of that.

I believe that all of these warnings should either be fixed or turned off. That way, if any future sub-par code props up, we'll quickly be able to detect it and fix it.

sebcrozet:
Yeah, `nalgebra` started being developed before `clippy` existed, and I never actually used since then unfortunately. There have been some efforts to fix clippy warning [on other merged PRs](https://github.com/dimforge/nalgebra/pulls?q=is%3Apr+is%3Aclosed+clippy) but obviously we haven't fixed everything yet.

Andlon:
I think this is a really worthwhile endeavor.  Once we have gotten to the "0 baseline" (either by fixing problems or judiciously ignoring individual special cases where we are doing the right thing), we can also consider adding stricter lints on top.

For example, we recently had a case where an automatic derive of `Deserialize` caused an unsoundness hole due to invariants used in unsafe code not being upheld. It turns out that there's a clippy lint (not turned on by default) that would have caught this issue, see https://github.com/RustSec/advisory-db/issues/880 for more information.

PR #914, opened 2021-06-16
Clippy fixing

OfficialURL:
I'll be trying to decrease the number of clippy warnings (with the end goal of closing #913.) I'll fix a few at a time to make sure I don't introduce any bugs, and I'll start with those that aren't breaking changes.

### DIFF
--- a/src/base/blas.rs
+++ b/src/base/blas.rs
@@ -1,5 +1,6 @@
 use crate::SimdComplexField;
 #[cfg(feature = "std")]
+#[allow(clippy::single_component_path_imports)]
 use matrixmultiply;
 use num::{One, Zero};
 use simba::scalar::{ClosedAdd, ClosedMul};
--- a/src/base/blas.rs
+++ b/src/base/blas.rs
@@ -1,6 +1,5 @@
 use crate::SimdComplexField;
 #[cfg(feature = "std")]
-#[allow(clippy::single_component_path_imports)]
 use matrixmultiply;
 use num::{One, Zero};
 use simba::scalar::{ClosedAdd, ClosedMul};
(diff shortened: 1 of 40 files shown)


OfficialURL:
I've got no idea what happened in Checkpoint 3. Apparently `nalgebra` didn't like it when I took out those... clones? Surely that shouldn't be having a noticeable effect.

OfficialURL:
Should I revert the changes I did where I replaced clones by dereferences? I don't see how these could be inequivalent, but since these were in a macro, I guess anything goes.

sebcrozet:
Thank you for this PR! Regarding the `clone` issue: I think the problem comes from the [trait-bound on the implementation](https://github.com/dimforge/nalgebra/blob/dev/src/geometry/translation.rs#L41) of `Copy` for `Translation`:
```rust
impl<T: Scalar + Copy, const D: usize> Copy for Translation<T, D> where Owned<T, Const<D>>: Copy {}
```

That `Owned<T, Const<D>>: Copy` should always be satisfied, but for some reasons the compiler fails to see it in some context. Just removing that trait bound (which is actually no longer necessary, this is just a remain from before the switch to const-generics):
```rust
impl<T: Scalar + Copy, const D: usize> Copy for Translation<T, D> {}
```
will allow you to get rid of the `.clone()`.

OfficialURL:
I'm having some trouble with `test-nalgebra` specifically. I must've accidentally reverted something important. How can I check this on my side? Running `cargo test --features debug,compare,rand,macros` doesn't give me any errors.

OfficialURL:
This PR will fix all clippy warnings, save for those that would be fixed by PR #918, and those that would close Issue #791 (the `#Safety` sections).

sebcrozet:
This is great, thank you!

PR #916, opened 2021-06-16
`Result<(), ()>` to `bool`

OfficialURL:
All other methods in `nalgebra` that mutate an argument and may possibly fail, return a boolean. Furthermore, ever since writing the pow code I've learned that it's actually bad practice to use unit structs as error types. 

This change would be breaking.


sebcrozet:
Thanks!

PR #917, opened 2021-06-16
Fixes #909

### LINKED ISSUE #909: Bug in Rotation3 try_slerp.
https://github.com/dimforge/nalgebra/blob/dev/src/geometry/rotation_interpolation.rs#L73 This code causes a stack overflow due to calling itself.

OfficialURL:
It seems like `try_slerp` was mistakenly calling itself, when it should have called the `UnitQuaternion` impementation.


sebcrozet:
Looks like a copy-past error… Thanks!

PR #918, opened 2021-06-17
`&self` → `self` in `to_X` methods in `Copy` structs

OfficialURL:
A new `clippy` lint enforces that any methods whose name starts with `to` in types that implement `Copy` must take `self` instead of `&self`. I've fixed all instances of this in the code. I believe this should be a breaking change, though in practice it should hardly be disruptive at all.

It might be worth going through the code and checking if any other methods like these should take the plain argument instead of a reference as well. Or if perhaps, the methods should be renamed instead.


sebcrozet:
This looks legitimate, thanks! It looks like merging #914 introduced some conflicts with that branch.

Andlon:
Aren't these types only `Copy` when `T: Copy`? Wouldn't this be unfortunate if `T` is *not* `Copy` (I know at least @aweinstock314 has been working with non-copy types).

sebcrozet:
@Andlon These methods require the `T: RealField` bound which derives from `Copy`.

Andlon:
Ah, I missed that they're all `T: RealField`. Is this something we'd like to enforce also in the future though? I have a use case right now where I'd like to use arbitrary precision floats (from e.g. the `rug` crate) in order to test numerical gradients by computing finite difference approximations in extended precision (otherwise precision is a huge problem due to catastrophic cancellation). If we move in this direction we're moving away from relaxing these bounds on `RealField`. I'm not sure what other implications removing `Copy` from `RealField` would have though...

OfficialURL:
I'm sorry, while trying to fix the merge conflict, I completely screwed up this branch. I've made a new PR over at #923 that does what this one was supposed to do.

Issue #920, opened 2021-06-17
Allow recovery of slices used in SliceStorage and SliceStorageMut


2pt0:
The `Storage` and `StorageMut` traits require implementation of `as_slice` and `as_mut_slice` methods, respectively. However, the returned slices have the lifetime of the reference to the storage object, not the underlying lifetime. No methods exist in `MatrixSlice` for recovery of the original slice.

It is sometimes necessary that the slice needs to be recovered. For example, when trying to reclaim the mutable slice transferred to a `SliceStorageMut` object to use `MatrixSliceMut` and friends. Consider the following example:

```rust
struct Buffer<'buf> {
    buffer: &'buf [f64],
    // ...
}

impl<'buf> Buffer<'buf> {
    fn to_vector_mut(mut self) -> Vector<'buf>{
        Vector { vector: DVectorSliceMut::from(self.buffer) }
    }
}

struct Vector<'buf> {
    vector: DVectorMut<'buf, f64>,
    // ...
}

impl<'buf> Vector<'buf> {
    fn to_buffer(mut self) -> Buffer<'buf> {
        // Can't implement. No way to extract the &'buf mut [f64] slice.
    }
}
```

Is such a feature possible to implement? I also have [this](https://stackoverflow.com/questions/68023386/extract-original-slice-from-slicestorage-and-slicestoragemut) open issue on SO.

PR #922, opened 2021-06-18
Allow to extract the inner slices of SliceStorage

Cryptjar:
This PR adds an inherent `into_slice` and `into_slice_mut` method to `SliceStorage` and `SliceStorageMut` (the latter gets both). Also, to complement the `From` impls which convert slices to `DVectorSlice` and `DVectorSliceMut`, this PR adds corresponding `From` impls to convert back to the respective slice type.

The implementation is essentially a generalization of [`as_mut_slice_unchecked`](https://github.com/dimforge/nalgebra/blob/38add0b00df2bfba89b8e753e7cb16fcf4ae93c3/src/base/matrix_slice.rs#L190-L198). I just noticed https://github.com/dimforge/nalgebra/pull/905, I'm not sure how this relates to this new functions, but I guess they should be not affected because they consume the respective storage struct entirely.

This would implement #920

_From a consistency point of view, I wonder whether it would make sense to also add an analog method to `VecStorage`._

### DIFF
--- a/src/base/conversion.rs
+++ b/src/base/conversion.rs
@@ -452,10 +452,22 @@ impl<'a, T: Scalar + Copy> From<&'a [T]> for DVectorSlice<'a, T> {
     }
 }
 
+impl<'a, T: Scalar> From<DVectorSlice<'a, T>> for &'a [T] {
+    fn from(vec: DVectorSlice<'a, T>) -> &'a [T] {
+        vec.data.into_slice()
+    }
+}
+
 impl<'a, T: Scalar + Copy> From<&'a mut [T]> for DVectorSliceMut<'a, T> {
     #[inline]
     fn from(slice: &'a mut [T]) -> Self {
         Self::from_slice(slice, slice.len())
+    }
+}
+
+impl<'a, T: Scalar> From<DVectorSliceMut<'a, T>> for &'a mut [T] {
+    fn from(vec: DVectorSliceMut<'a, T>) -> &'a mut [T] {
+        vec.data.into_slice_mut()
     }
 }
 
--- a/src/base/matrix_slice.rs
+++ b/src/base/matrix_slice.rs
@@ -80,6 +80,8 @@ macro_rules! slice_storage_impl(
 
         impl <'a, T: Scalar, R: Dim, C: Dim, RStride: Dim, CStride: Dim>
             $T<'a, T, R, C, RStride, CStride>
+        where
+            Self: ContiguousStorage<T, R, C>
         {
             /// Extracts the original slice from this storage
             pub fn into_slice(self) -> &'a [T] {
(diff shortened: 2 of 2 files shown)


sebcrozet:
Thank you for this PR. This could lead to the same kind undefined-behavior as in #905 when doing this:
```rust
let (mut row1, mut row2) = matrix.rows_range_pair_mut(0, 1);
let slice1 = row1.into_slice_mut();
let slice2 = row2.into_slice_mut();
some_function(&mut slice1[1], &mut slice2[0]);
```

So all these conversions should only be implemented for matrix slices which have a contiguous internal storage. However, right now, the `ContiguousStorage` isn't implemented for any of the matrix slice storages (by lack of specialization I think becaute it would depends on the strides and dimensions, it's been a while since I last attempted to implement it).

Cryptjar:
@sebcrozet  Ok, I looked into it in more detail. As I understand it, `rows_range_pair_mut` might return non-continues _things_, which results in phony slices when using `into_slice` (and friends). I reproduced it with following code:

```rust
use nalgebra::Matrix3;
fn main() {
    let mut matrix = Matrix3::<f64>::new(1., 2., 3., 4., 5., 6., 7., 8., 9.);

    let (row1, row2) = matrix.rows_range_pair_mut(0, 1);

    let slice1 = row1.data.into_slice_mut();
    let slice2 = row2.data.into_slice_mut();

    some_function(&mut slice1[1], &mut slice2[0]);
}
fn some_function(a: &mut f64, b: &mut f64) {
    println!("{:p}, {:p}", a, b);
    assert!(!core::ptr::eq(a, b));
}
```

It compiles on 801d8a6 and gives in deed aliased slices, i.e. UB!

So, I though, I just add a `ContiguousStorage` and `ContiguousStorageMut` bound respectively, because, as I understand it, this trait gives exactly the guarantee we need here. And in deed, this bound prevents my above test code from compiling.

> However, right now, the ContiguousStorage isn't implemented for any of the matrix slice storages

Well, if I understand the docs correctly, `ContiguousStorage` is in fact implemented, well, on [certain `SliceStorage`  types](https://docs.rs/nalgebra/0.27.1/nalgebra/base/storage/trait.ContiguousStorage.html#impl-ContiguousStorage%3CT%2C%20R%2C%20Const%3C1_usize%3E%3E). However, I wrote the following code to test the use-case of #920:

```rust
use nalgebra::DVectorSliceMut;

struct Buffer<'buf>(&'buf mut [f64]);
impl<'buf> Buffer<'buf> {
    fn to_vector_mut(self) -> Vec<'buf> {
        Vec(DVectorSliceMut::from(self.0))
    }
}
struct Vec<'buf>(DVectorSliceMut<'buf, f64>);
impl<'buf> Vec<'buf> {
    fn to_buffer(self) -> Buffer<'buf> {
        Buffer(self.0.into())
    }
}

fn main() {
    let mut array = [1., 2., 3.];
    foo(Buffer(&mut array));
}
fn foo<'a>(buffer: Buffer<'a>) {
    let mut vec = buffer.to_vector_mut();
    vec.0.normalize_mut();
    let mut buffer2: Buffer<'a> = vec.to_buffer();
    println!("{:?}", buffer2.0);
}
```

And it still works as expected with the trait bound.

_I would also add it as an example or test case, but I'm not sure where to put it._

sebcrozet:
Thanks!

> Well, if I understand the docs correctly, ContiguousStorage is in fact implemented, well, on certain SliceStorage types.
You are correct, I forgot about these.

PR #923, opened 2021-06-18
`&self` → `self` on `Copy` types.

OfficialURL:
I screwed up #918 pretty badly, this PR does what that one was supposed to do.


sebcrozet:
Thanks!

PR #925, opened 2021-06-29
Impl bytemuck traits for geometry types

### LINKED ISSUE #830: Bytemuck implementations for statically sized matrixes.

CAD97:
Adds impls for `bytemuck::Zeroable` and `bytemuck::Pod` to the `na::geometry::*` types that don't already have them. All added impls just fully gate themselves on the fields' impls for these traits, so they should all be sound, unless I missed one of the following conditions:

- The type has invalid values (semantically, in which case the traits shouldn't be impl'd); or
- The type is not stable layout (`#[repr(C)]` or `#[repr(transparent)]`; or
- The type is stable layout but may have padding (for _any_ value of `T` or other type parameters)

I believe this should reasonably close #830.

### DIFF
--- a/src/geometry/dual_quaternion.rs
+++ b/src/geometry/dual_quaternion.rs
@@ -242,10 +242,20 @@ where
 }
 
 #[cfg(feature = "bytemuck")]
-unsafe impl<T> bytemuck::Zeroable for DualQuaternion<T> where Quaternion<T>: bytemuck::Zeroable {}
+unsafe impl<T> bytemuck::Zeroable for DualQuaternion<T>
+where
+    T: Scalar + bytemuck::Zeroable,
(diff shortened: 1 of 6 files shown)


sebcrozet:
Thank you for this PR! The CI is currently failing because some of the impls are missing trait-bounds.

CAD97:
That's what I get for not checking `--all-features` 😅

Some places that are bounded on `RealField` could probably be loosened further to just `Scalar`, but that should probably wait until merged with #932.

CAD97:
Resolved merge conflicts; should be good to merge.

sebcrozet:
Thanks!

PR #929, opened 2021-07-05
Fix some clippy warnings and typo

remilauzier:


src/sparse/cs_matrix_cholesky.rs
```diff
@@ -293,7 +293,7 @@ where
     ) -> (CsMatrix<T, D, D>, CsMatrix<T, D, D>) {
         let etree = Self::elimination_tree(m);
         let (nrows, ncols) = m.data.shape();
-        let mut rows = Vec::with_capacity(m.len());
+        let rows = Vec::with_capacity(m.len());
```
sebcrozet, line 296: This change makes the CI fail. It should be mutable because `&mut rows` is used bellow at line 305.

sebcrozet (changes requested):
Thank you for this PR!

sebcrozet (approved):

sebcrozet:
Thanks!

PR #931, opened 2021-07-06
Rotation::matrix_mut_unchecked may not cause UB

CAD97:
...because it isn't unsafe, so clarify the documentation to match.


sebcrozet:
Thanks!

PR #932, opened 2021-07-06
Remove scalar bound from geometry type defs

CAD97:
This was inconsistently applied, with some types having <T>, some having
<T: Scalar>, and some having <T: RealField>.
This unifies all types to match the convention of Matrix:
Just declare <T> at type def time, and apply bounds on impls only.

A significant advantage of this approach is const fn construction. Const
fn generics currently still can't have trait bounds, so any generic
const fn needs to only move opaque types around. Construction methods
such as new_unchecked or from_parts can be made const by removing their
generic bounds after this PR.

Actual constification is left to a follow-up PR.

Note that na::Transform is _not_ loosened here, as it has more complicated
definition requirements.

### DIFF
--- a/src/geometry/dual_quaternion.rs
+++ b/src/geometry/dual_quaternion.rs
@@ -38,12 +38,21 @@ use simba::scalar::{ClosedNeg, RealField};
 ///  If a feature that you need is missing, feel free to open an issue or a PR.
 ///  See https://github.com/dimforge/nalgebra/issues/487
 #[repr(C)]
-#[derive(Debug, Eq, PartialEq, Copy, Clone)]
-pub struct DualQuaternion<T: Scalar> {
+#[derive(Debug, Copy, Clone)]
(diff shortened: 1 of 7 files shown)


sebcrozet:
Thank you for this PR! This is a good idea.

I think the CI is failing when nalgebra is built with the `serde` feature enabled. We probably need to add to the serde macro all the trait-bounds your removed from the type definition.

CAD97:
That's what I get for not checking `--all-features` 😅

Should work now.

CAD97:
I don't think that proptest failure is even possibly caused by my changes.

sebcrozet:
Yeah, this happens from time to time if we are unlucky with the test inputs (and get a particularly badly conditioned matrix).
I have re-run the tests.

sebcrozet:
It succeeds now, thanks!

PR #933, opened 2021-07-06
Add cargo-features resolver

tcharding:
Attempting to run `cargo udeps` throws the following warning:

```
cargo +nightly udeps
error: failed to parse manifest at `/home/tobin/build/github.com/tcharding/nalgebra/Cargo.toml`

Caused by:
  feature `resolver` is required

  consider adding `cargo-features = ["resolver"]` to the manifest
```

This can be resolved by adding `cargo-features = ["resolver"]` to the
manifest file as suggested.

_Why_ this is needed by cargo udeps when it is the default behaviour now
I do not know.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,5 @@
+cargo-features = ["resolver"]
+
 [package]
 name    = "nalgebra"
 version = "0.27.1"


tcharding:
I discovered this when trying to run `cargo udeps` on a code base that has `nalgebra` as a dependency. We get the same error then.

tcharding:
This is actually not needed, re-installing udeps resolves the issue.

```
cargo uninstall cargo-udeps
cargo install cargo-udeps
```

PR #934, opened 2021-07-06
point_dim_name: Fix serde impls and Distribution impl for OPoint

Andlon:
This adds some missing fixes on top of the `point_dim_name` branch that were necessary for `serde` and `rand` integrations to work correctly.

Hopefully we can run the full CI suite to uncover additional problems.


Andlon:
@sebcrozet: Hmm, looks as if CI only runs against the `dev` branch. Should I re-submit the PR against the dev branch, or how do you prefer to proceed?

sebcrozet:
Thank you for the PR! I will just merge this, and open a PR for the point stuffs. The CI will run there.

Issue #216, opened 2017-01-12
Implementing borrow for fixed size types


arturoc:
Right now most types implement AsRef like Vector3 implements AsRef<[T;3]>.

There's a case where being able to accept both a &[f32;3] or a Vec3 would be useful. For example to create universal apis that can work with any library like:

```
fn clear_color<C: Borrow<&[f32;4]>(c: &C)
```

can be called as:

```
clear_color( &[1., 1., 1., 1.]);
clear_color(&na::Vector4(1., 1., 1., 1.));
clear_color(&color::Rgba(1., 1., 1., 1.));
```

while AsRef only would support the last 2 cases, it's impossible to take &[f32;4] as AsRef without having some really verbose trait implementation while Borrow has a blanket implementation for any &T.

This would allow to create apis that could take any type with the correct memory structure without having dependencies on any specific library.

Not sure though if this is a correct usage of Borrow or there's a better way to accomplish the same.

I can create a PR request implementing Borrow for any type where it makes sense if this is something that seems useful

phaazon:
That’s interesting. If you have the PR, I think it’s worth it. If you don’t, I can help bringing the feature up.

aepsil0n:
The book and standard library documentation distinguish those traits mostly in terms of the intent they express. One example where they differ would be `String`, that can be turned into `&[u8]` or `&str` using `AsRef`, but only into `&str` using `Borrow`.

However, we do not really know, whether people want to convert vectors into references or borrow their contents. It makes perfect sense to implement `Borrow` just as we implement `AsRef` in this case, I think.

@arturoc would you still be willing to create a PR for this?

cc @sebcrozet just in case you feel we should not implement `Borrow`.

sebcrozet:
I have no objection. Of course, just like `AsRef`, this should be implemented only for vectors with contiguous storage.

arturoc:
I haven't looked into this further at some point i managed to make things work just by using AsRef (mostly once the ambiguities where solved) and i didn't needed it anymore but i'll try to implement it if i have a moment

jsmith628:
I saw this was collecting dust, so I decided to just do it in case you all still wanted this

PR #937, opened 2021-07-07
Implemented Borrow and BorrowMut on fixed size matrices

jsmith628:
I saw #216 and thought I'd just take five minutes and do it. I know it's kinda an old issue, so feel free to close this if the consensus there has changed.

Also, I saw that `AsRef`/`AsMut`/`Borrow`/`BorrowMut` isn't `impl`'d for `Quaternion` or `Complex` either. I could also go ahead and do those as well if no one can see a reason they shouldn't be.


sebcrozet:
Thank you for this PR! I think the CI is failing because you need to run `cargo fmt`.

sebcrozet:
Thanks!

PR #938, opened 2021-07-08
Make display generical for Omatrix

Nateckert:
In the previous version, the test added in tests/core/matrix.rs failed to compile.

This version remove the dependency to the `Allocator<usize>`, so that the display implementation does not depend of the allocator of `nalgebra`, making it generical for `OMatrix`.

Not relying on `Omatrix<usize, R, C>` implies that some computation have to be done twice.
However, the performance hit is only in `Display`, which I find acceptable (the use case will be for debugging/writing logs)

The difference with the previous version is that `val_width` is called twice per entry instead of once.

### DIFF
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -1819,7 +1819,6 @@ macro_rules! impl_fmt {
         where
             T: Scalar + $trait,
             S: Storage<T, R, C>,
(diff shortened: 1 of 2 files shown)


sebcrozet:
Thanks!

PR #940, opened 2021-07-08
Removed unnecessary `unsafe` markers

OfficialURL:
I removed some `unsafe` markers from methods that didn't need them, and added a bit of safety documentation to related methods.


sebcrozet:
Thanks!

PR #941, opened 2021-07-08
Attempt at fixing UB in `allocate_uninitialized`

OfficialURL:
Currently, `allocate_uninitialized` uses a `set_len` call to make an "uninitialized" vector. However, this is undefined behavior per the Rust docs. I'm currently attempting to remedy this by using `MaybeUninit`. 

Tests are currently crashing, so either my implementation is wrong, or there's UB elsewhere (or both).


OfficialURL:
This PR will (hopefully) be superseded by a much more general fix.

PR #948, opened 2021-07-15
Add Transform × UnitComplex & friends

CAD97:
... following the same implementation strategy as was used for `UnitQuaternion`.


sebcrozet:
Thanks!

PR #950, opened 2021-07-22
Make `from_matrix_unchecked`es const

CAD97:
Now that [`T` is unbounded](https://github.com/dimforge/nalgebra/pull/932), the unchecked constructors can be `const`.

### DIFF
--- a/src/geometry/rotation.rs
+++ b/src/geometry/rotation.rs
@@ -130,10 +130,10 @@ where
     }
 }
 
-impl<T: Scalar, const D: usize> Rotation<T, D> {
+impl<T, const D: usize> Rotation<T, D> {
     /// Creates a new rotation from the given square matrix.
     ///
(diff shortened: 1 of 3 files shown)


src/geometry/rotation.rs
```diff
@@ -154,12 +154,7 @@ impl<T: Scalar, const D: usize> Rotation<T, D> {
     /// assert_eq!(*rot.matrix(), mat);
     /// ```
     #[inline]
-    pub fn from_matrix_unchecked(matrix: SMatrix<T, D, D>) -> Self {
-        assert!(
-            matrix.is_square(),
-            "Unable to create a rotation from a non-square matrix."
-        );
```
CAD97, line 161: Note that squareness is still statically enforced, since we take `SMatrix<_, D, D>`, which doesn't allow for dynamic sizing.

sebcrozet:
Thanks!

PR #953, opened 2021-07-26
Do not require T: (De)Serialize for OPoint impl

Andlon:
The bounds recently got a little too strict.

I probably messed this up when I fixed serde compilation during the recent `OPoint` introduction?


sebcrozet:
Thanks!

PR #954, opened 2021-07-26
Include "rand" in feature list for docs.rs

jpetkau:
Random support was gated by the "rand" feature in version 0.25.1, but not added to the docs.rs list, causing the gated functions to disappear from docs.


sebcrozet:
Thanks!

PR #956, opened 2021-07-27
Fix some warnings

remilauzier:


nalgebra-sparse/src/csc.rs
```diff
@@ -260,7 +260,7 @@ impl<T> CscMatrix<T> {
     /// let triplets: Vec<_> = csc.triplet_iter().map(|(i, j, v)| (i, j, *v)).collect();
     /// assert_eq!(triplets, vec![(0, 0, 1), (2, 0, 3), (1, 1, 2), (0, 2, 4)]);
     /// ```
-    pub fn triplet_iter(&self) -> CscTripletIter<T> {
+    pub fn triplet_iter(&self) -> CscTripletIter<'_, T> {
```
Andlon, line 263: Why add explicit (unnamed) lifetimes here (and elsewhere)? Is this a clippy lint?

remilauzier: It's rust_2018_idioms. Undeclared anonymous lifetime are deprecated and might be remove in 2021 version.

remilauzier: https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#elided-lifetimes-in-paths

Andlon: Huh, that's a shame. Adds a significant amount of noise to our APIs without much benefit in return, in my opinion. For example we have a bunch of APIs taking `DVectorSlice<T>` or `DVectorSliceMut<T>` as input arguments. Oh well, if that's the way it's going to be, then I guess that's how it will be.

Thanks for explaining! :)

sebcrozet:
Thanks!

PR #959, opened 2021-07-30
Fix a typo in the doc comment for `Mat2x3`

zedseven:
Currently, the doc comment for `Mat2x3` incorrectly describes it as a 2x2 matrix. Obviously this is a very minor issue, but I figured it was worth fixing.


sebcrozet:
Thanks!

PR #962, opened 2021-08-04
Tha clone wars: allow non-copy scalar types everywhere

sebcrozet:
This PR illustrates what needs to be done to make `nalgebra` work if we decide to replace the `SimdComplexField: Copy` requirement by `SimdComplexField: Clone`. This has no impact on (optimized) performances, but adds a lot of noise to the codebase (I had to add about 2000 calls to `.clone()`).

This also removes our `.inlined_clone()` thing as it has only a ~10% performance impact on debug performances. This is negligible compared to the fact that the debug run takes between 20000% and 100000% more time than release (yes, that much in our synthetic benchmarks according to `criterion`!) The main benefit of removing `.inlined_clone()` is that we can have a blanket impl of `Scalar` for all the `Clone` types.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,7 @@ num-traits     = { version = "0.2", default-features = false }
 num-complex    = { version = "0.4", default-features = false }
 num-rational   = { version = "0.4", default-features = false }
 approx         = { version = "0.5", default-features = false }
-simba          = { version = "0.5", default-features = false }
+simba          = { version = "0.6", default-features = false }
 alga           = { version = "0.9", default-features = false, optional = true }
 rand_distr     = { version = "0.4", default-features = false, optional = true }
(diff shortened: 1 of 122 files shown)


Andlon:
Love the new title by the way :joy:

pwnorbitals:
this would make nalgebra compatible with types like Rug::Float and I would love it

Andlon:
> this would make nalgebra compatible with types like Rug::Float and I would love it

That is precisely the motivation :-)

sebcrozet:
This will be merged & released today. Note that types from `rug` won’t be supported yet when `nalgebra` expects `RealField/ComplexField` because these traits have not been implemented yet.

pwnorbitals:
This is true, though I believe it is now possible to use them through a newtype. Previously this couldn't be done

PR #963, opened 2021-08-07
Fix wrong reference in docs

mariusknaust:
The `row_sum` documentation incorrectly referred to `row_variance_tr` instead of `row_sum_tr`.


sebcrozet:
Thanks!

PR #969, opened 2021-08-15
Fix some clippy warnings

remilauzier:


src/base/matrix.rs
```diff
@@ -357,8 +357,10 @@ mod rkyv_impl {
 }
 
 impl<T, R, C, S> Matrix<T, R, C, S> {
-    /// Creates a new matrix with the given data without statically checking that the matrix
-    /// dimension matches the storage dimension.
+    /// Creates a new matrix with the given data.
+    ///
+    /// # Safety
+    /// The function doesnt verify that the dimension of the matrix correspond to the storage dimension.
```
filnet, line 363: typo: doesn't

src/base/matrix_slice.rs
```diff
@@ -29,7 +29,10 @@ macro_rules! slice_storage_impl(
         {}
 
         impl<'a, T, R: Dim, C: Dim, RStride: Dim, CStride: Dim> $T<'a, T, R, C, RStride, CStride> {
-            /// Create a new matrix slice without bound checking and from a raw pointer.
+            /// Create a new matrix slice..
+            ///
+            /// # Safety
+            /// The function doesnt do bound checking and use a raw pointer.
```
filnet, line 35: typo: doesn't

filnet: typo: uses

examples/unit_wrapper.rs
```diff
@@ -23,5 +23,5 @@ fn main() {
     let l1 = length_on_direction_with_unit(&v, &Vector3::y_axis());
     let l2 = length_on_direction_without_unit(&v, &Vector3::y());
 
-    assert_eq!(l1, l2)
+    assert!((l1 - l2).abs() < f32::EPSILON)
```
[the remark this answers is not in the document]
sebcrozet: The whole point of this example is to show that we get the exact same result with the `Unit<Vector>` and the `Vector` (both having the exact same components).

sebcrozet (approved):

PR #970, opened 2021-08-15
Implement a few handy functions for `Quaternion` and `DualQuaternion`

Ewpratten:
In a recent project, I stumbled across a few quaternion related functions I wished this library had. So, here is my pull request to try to add them.

I would appreciate feedback on my work here if anybody has some to give.

## What's new

 - ~`Quaterion::rotate(&self, Vector3) -> Vector3`~
   - ~Given any `Vector3`, rotate it by the quaternion and return the rotated vector result~
 - `DualQuaternion::from_real_and_translation(Quaterion, Vector3) -> DualQuaternion`
   - Construct a `DualQuaternion` from a real part (rotation) and a vector representing a 3d translation
 - `DualQuaternion::translation(&self) -> Vector3`
   - Calculate the 3d translation that any `DualQuaternion` represents

### DIFF
--- a/src/geometry/dual_quaternion.rs
+++ b/src/geometry/dual_quaternion.rs
@@ -8,8 +8,7 @@ use crate::{
 use approx::{AbsDiffEq, RelativeEq, UlpsEq};
 #[cfg(feature = "serde-serialize-no-std")]
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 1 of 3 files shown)


Ewpratten:
Edit:

While double-checking my work I noticed my implementation `Quaternion::rotate` already exists as `UnitQuaternion::transform_vector`

PR #979, opened 2021-08-30
Allow setting Cholesky field directly; more lax decomposition method

Fuuzetsu:
Please see individual commits. This makes working with Cholesky quite a bit more flexible.

### DIFF
--- a/src/linalg/cholesky.rs
+++ b/src/linalg/cholesky.rs
@@ -171,7 +171,32 @@ where
     ///
     /// Returns `None` if the input matrix is not definite-positive. The input matrix is assumed
     /// to be symmetric and only the lower-triangular part is read.
-    pub fn new(mut matrix: OMatrix<T, D, D>) -> Option<Self> {
+    pub fn new(matrix: OMatrix<T, D, D>) -> Option<Self> {
+        Self::new_internal(matrix, None)
+    }
+
+    /// Attempts to approximate the Cholesky decomposition of `matrix` by
+    /// replacing non-positive values on the diagonals during the decomposition
+    /// with the given `substitute`.
+    ///
+    /// [`try_sqrt`](ComplexField::try_sqrt) will be applied to the `substitute`
+    /// when it has to be used.
+    ///
+    /// If your input matrix results only in positive values on the diagonals
+    /// during the decomposition, `substitute` is unused and the result is just
+    /// the same as if you used [`new`](Cholesky::new).
+    ///
+    /// This method allows to compensate for matrices with very small or even
+    /// negative values due to numerical errors but necessarily results in only
+    /// an approximation: it is basically a hack. If you don't specifically need
+    /// Cholesky, it may be better to consider alternatives like the
+    /// [`LU`](crate::linalg::LU) decomposition/factorization.
+    pub fn new_with_substitute(matrix: OMatrix<T, D, D>, substitute: T) -> Option<Self> {
+        Self::new_internal(matrix, Some(substitute))
+    }
+
+    /// Common implementation for `new` and `new_with_substitute`.
+    fn new_internal(mut matrix: OMatrix<T, D, D>, substitute: Option<T>) -> Option<Self> {
         assert!(matrix.is_square(), "The input matrix must be square.");
 
         let n = matrix.nrows();
@@ -187,17 +212,25 @@ where
                 col_j.axpy(factor.conjugate(), &col_k, T::one());
             }
 
-            let diag = unsafe { matrix.get_unchecked((j, j)).clone() };
-            if !diag.is_zero() {
-                if let Some(denom) = diag.try_sqrt() {
-                    unsafe {
-                        *matrix.get_unchecked_mut((j, j)) = denom.clone();
-                    }
-
-                    let mut col = matrix.slice_range_mut(j + 1.., j);
-                    col /= denom;
-                    continue;
+            let sqrt_denom = |v: T| {
+                if v.is_zero() {
+                    return None;
                 }
+                v.try_sqrt()
+            };
+
+            let diag = unsafe { matrix.get_unchecked((j, j)).clone() };
+
+            if let Some(denom) =
+                sqrt_denom(diag).or_else(|| substitute.clone().and_then(sqrt_denom))
+            {
+                unsafe {
+                    *matrix.get_unchecked_mut((j, j)) = denom.clone();
+                }
+
+                let mut col = matrix.slice_range_mut(j + 1.., j);
+                col /= denom;
+                continue;
             }
 
             // The diagonal element is either zero or its square root could not
--- a/src/linalg/cholesky.rs
+++ b/src/linalg/cholesky.rs
@@ -74,6 +74,14 @@ where
         Cholesky { chol: matrix }
     }
 
+    /// Uses the given matrix as-is without any checks or modifications as the
+    /// Cholesky decomposition.
+    ///
+    /// It is up to the user to ensure all invariants hold.
+    pub fn pack_dirty(matrix: OMatrix<T, D, D>) -> Self {
+        Cholesky { chol: matrix }
+    }
+
     /// Retrieves the lower-triangular factor of the Cholesky decomposition with its strictly
     /// upper-triangular part filled with zeros.
     pub fn unpack(mut self) -> OMatrix<T, D, D> {
--- a/tests/linalg/cholesky.rs
+++ b/tests/linalg/cholesky.rs
@@ -1,5 +1,16 @@
 #![cfg(all(feature = "proptest-support", feature = "debug"))]
 
+#[test]
+// #[rustfmt::skip]
+fn cholesky_with_substitute() {
+    // Make a tiny covariance matrix with a small covariance value.
+    let m = na::Matrix2::new(1.0, f64::NAN, 1.0, 1e-32);
+    // Show that the cholesky fails for our matrix. We then try again with a substitute.
+    assert!(na::Cholesky::new(m).is_none());
+    // ...and show that we get some result this time around.
+    assert!(na::Cholesky::new_with_substitute(m, 1e-8).is_some());
+}
+
 macro_rules! gen_tests(
     ($module: ident, $scalar: ty) => {
         mod $module {


src/linalg/cholesky.rs
```diff
@@ -200,6 +207,50 @@ where
         Some(Cholesky { chol: matrix })
     }
 
+    /// Attempts to compute the Cholesky decomposition of `matrix`, using the
+    /// given element in places where the decomposition would otherwise fail.
+    ///
+    /// The result can be garbage if your value is non-positive, resulting in
+    /// non-PSD matrix. Same caveats as with `new_unchecked` apply.
+    pub fn new_with_substitute_unchecked(mut matrix: OMatrix<T, D, D>, substitute: T) -> Self
```
sebcrozet, line 215: Having to copy the decomposition code for this is unfortunate. We should just modify the `Cholesky::new(matrix)` to this instead:
```rust
pub fn new(mut matrix: OMatrix<T, D, D>, substitute: Option<T>) -> Option<Self> {
```

sebcrozet (changes requested):
Thank you for this PR!

src/linalg/cholesky.rs
Andlon: I'd say this would be unfortunate as in 99+% of the time, users don't need this functionality, so `new` should ideally just take a single matrix.

Perhaps instead we can implement `new` and the new function in terms of some internal function `new_with_substitute` that has the type signature that @sebcrozet proposed?

Andlon:
Interesting... Do you have any examples for when this is actually useful? Why you use a direct solver if the solution you get might be garbage?

src/linalg/cholesky.rs
sebcrozet: > 
> 
> I'd say this would be unfortunate as in 99+% of the time, users don't need this functionality, so `new` should ideally just take a single matrix.
> 
> Perhaps instead we can implement `new` and the new function in terms of some internal function `new_with_substitute` that has the type signature that @sebcrozet proposed?

I think that would be unnecessarily complicated. Passing `Matrix::new(matrix, None)` doesn’t sound too noisy compared to having more construction methods.

sebcrozet:
@Andlon I believe that’s answered on [that commit comment](https://github.com/dimforge/nalgebra/pull/979/commits/8ddb2da863b2ebefb23f04d3c85a999d5882ab56). Basically, very small numerical error may result in slightly negative diagonal entries that would break Cholesky for no legitimate reason.
@Fuuzetsu It would be great to add a test where using this epsilon proves useful.

src/linalg/cholesky.rs
Andlon: I'm sorry, but I completely disagree. If I see a function call `Cholesky::new(matrix)` it's clear what's going on. If I see `Cholesky::new(matrix, None)`, I have no idea what's going on and I have to look up the docs to see what the `None` is.

Andlon:
I fear that this feature is not very useful to users in its current form, because the docs are not specific as to what "substitute" is, and what an appropriate "substitute" would be. I think in order to remedy this the docs would need to be expanded a little to explain what it does, and some guidelines on how to choose an appropriate substitute.

src/linalg/cholesky.rs
```diff
@@ -74,6 +74,13 @@ where
         Cholesky { chol: matrix }
     }
 
+    /// Uses the given matrix as-is without any checks or modifications as the
+    /// Cholesky decomposition. It is up to the user to ensure and and all
```
Andlon, line 78: a couple of extra `and` here, I think?

src/linalg/cholesky.rs
sebcrozet: Ok, that sounds legitimate, to make reading code easier. Let’s do it as you suggested then. We can add `pub fn new_with_substitute(matrix, substitute)` and then implement

```rust
pub fn new(matrix) -> Option<Self> { 
    Self::new_with_substitute(matrix, None) 
}
```

Andlon:
I skimmed through the motivating issues in `friedrichs`: https://github.com/nestordemeure/friedrich/issues/43

I still haven't seen any convincing evidence that this is a "sound" numerical operation. How do you know that the errors don't propagate causing a vastly "wrong" decomposition? Sure - your Cholesky will not return an error anymore, but have you thoroughly determined that the returned decomposition is actually reasonable accurate?

The fundamental issue, to me, is that your matrix is in fact for all intents and purposes singular. In this case the Cholesky factor is not well-defined in the first place, so hacking Cholesky to return a meaningful result might be misleading to the user.

I'm sorry if I sound overly negative here - I'd be happy to engage in a discussion to help find a good solution to your problem, I just worry that the solution proposed here is only *perhaps* appropriate for Gaussian processes and rarely otherwise.

Andlon:
As a follow-up, this short note on Cholesky factorization by Nicholas Higham (a renowned expert on numerical linear algebra) may be of interest: https://www.maths.manchester.ac.uk/~higham/papers/high09c.pdf

In particular, it describes how using *complete pivoting* with Cholesky can be used to decompose semi-definite matrices.

nestordemeure:
I see this operation as a *forced* Cholesky decomposition: it is useful when the choice is between this potentially dirty decomposition and no decomposition at all, killing further steps in an algorithm. One thing that is clear is that the doc should definitely emphatise the point that the result is an approximation (the name of the method might also be selected to further reflect that).

If there is interest in it, to at least experimentally validate the approach, I could measure the accuracy of the modified method on some singular matrices (porting the code to C++ or Julia to analyse it with [Shaman](https://gitlab.com/numerical_shaman/shaman/-/tree/master), treating `expected_diagonal_value - substitute` as additional numerical error).

A pivoting Cholesky would definitely be interesting to try ([it has been explored as a solution to decompose the matrices produced by kernel methods](https://arxiv.org/abs/1505.06195)) but there is no implementation in nalgebra as far as I know and I do not have the time to contribute one at the moment.

nestordemeure:
> Interesting... Do you have any examples for when this is actually useful? Why you use a direct solver if the solution you get might be garbage?

To give you more context, the [code](https://github.com/nestordemeure/friedrich) that triggered this PR solves a linear system but also inverses a matrix and makes use of the `L` sub-matrix (depending on the funcitonality that is called by a user) so the Cholesky decomposition serves a variety of purposes that would be hard to replace with an alternative.

src/linalg/cholesky.rs
Fuuzetsu: I wanted to keep existing API as-is so that making a small release would be faster.

I did want to make `new` just call `new_with_substitute`, just forgot. Will do.

Fuuzetsu:
I have added a commit that makes `new` use the new function, just setting substitute to `None`. I also changed the behaviour on the substitute very slightly so that it's subject to the same non-negativity check while dropping the explicit Signed constraint which would cause an API change.

Fuuzetsu:
> I fear that this feature is not very useful to users in its current form, because the docs are not specific as to what "substitute" is, and what an appropriate "substitute" would be. I think in order to remedy this the docs would need to be expanded a little to explain what it does, and some guidelines on how to choose an appropriate substitute.

I don't myself know how to pick a good substitute. I saw some other places using values like 1e-8 or 1e-10 but I don't know how to justify it in a way that looks useful in docs.

If there are any concrete ideas about what the docs should say, please let me know and I can add them (or feel free to add them directly). For the most part, I kind of just expected that people who don't need this feature just keep using `new` and those that run into the issue and want to apply similar work-around can find `new_with_substitute` and use that, picking the value as they see fit.

Basically, it should be seen as a function for only those users that look for it while everyone else keeps using `new` as always and don't worry about `new_with_substitute` existing.

Andlon:
@nestordemeure, @Fuuzetsu: thanks for the additional context! I can see that this might be a workable "hack" in lieu of more principled methods. I'm not opposed to adding this to `nalgebra`, I just think communicating some of the context we discussed here is important: in particular, these methods should clearly be documented as providing no guarantees as to the quality of the approximation, and perhaps even suggest to use other methods unless the user really knows what they are doing. In particular, we might suggest that they could use LU decomposition - or even `FullPivLU` if they need the added stability.

Additionally, I think if we add such a method, then perhaps adding a little bit more detail as to what exactly the substitute does would be good. In particular, it would be good to clarify that it essentially substitutes non-positive entries encountered along the diagonal during decomposition. My wording here is not very good - I don't have the Cholesky algorithm perfectly memorized.

Fuuzetsu:
> @nestordemeure, @Fuuzetsu: thanks for the additional context! I can see that this might be a workable "hack" in lieu of more principled methods. I'm not opposed to adding this to `nalgebra`, I just think communicating some of the context we discussed here is important: in particular, these methods should clearly be documented as providing no guarantees as to the quality of the approximation, 

Yes, I suppose we can add it. I think we can mark it as a hack explicitly and then there should be no complains if it does something surprising.

>and perhaps even suggest to use other methods unless the user really knows what they are doing. In particular, we might suggest that they could use LU decomposition - or even `FullPivLU` if they need the added stability.
 
 I don't know if just using LU decomposition is a substitute in every situation: I'm not qualified at all in the area and the PR is basically just mimicking what I saw some other libraries do. I can add a mention anyway if you feel it's useful though.
 
> Additionally, I think if we add such a method, then perhaps adding a little bit more detail as to what exactly the substitute does would be good. In particular, it would be good to clarify that it essentially substitutes non-positive entries encountered along the diagonal during decomposition. My wording here is not very good - I don't have the Cholesky algorithm perfectly memorized.

I already added something at https://github.com/dimforge/nalgebra/pull/979/files#diff-ad542399a1cc2be4f534cfde616de3355434c5f4fdaff0c0f247a61bb81542d3R179 but maybe that's not clear enough?

Andlon:
> I don't know if just using LU decomposition is a substitute in every situation: I'm not qualified at all in the area and the PR is basically just mimicking what I saw some other libraries do. I can add a mention anyway if you feel it's useful though.

Yeah, you're right in that it's not a substitute for all situations - in particular, if you *need* something like a Cholesky factorization it wouldn't fill that role. But if someone only wants to solve a linear system, then it would be good to point them in that direction perhaps?

> I already added something at https://github.com/dimforge/nalgebra/pull/979/files#diff-ad542399a1cc2be4f534cfde616de3355434c5f4fdaff0c0f247a61bb81542d3R179 but maybe that's not clear enough?

Yeah, that's pretty good already, although it's still a little ambiguous: For example it might read as if the substitute is put in place of *any* `0` value, whereas it's really only the ones that appear on the diagonal during factorization.

src/linalg/cholesky.rs
```diff
@@ -163,7 +170,23 @@ where
...
+    /// If `substitute` is given, any values during the decomposition that are 0
+    /// or otherwise fail `ComplexField::try_sqrt` will be replaced with
+    /// `substitute` instead. The `substitute` itself is subject to the same
+    /// checks and will by `try_sqrt`'d before being used.
+    ///
+    /// This allows to compensate for matrices with very small or even negative
+    /// values due to numerical errors.
+    pub fn new_with_substitute(
+        mut matrix: OMatrix<T, D, D>,
+        substitute: Option<T>,
```
Andlon, line 188: What's the case for taking `Option<T>` in the public API instead of `T`?

nestordemeure:
> Additionally, I think if we add such a method, then perhaps adding a little bit more detail as to what exactly the substitute does would be good. In particular, it would be good to clarify that it essentially substitutes non-positive entries encountered along the diagonal during decomposition. My wording here is not very good - I don't have the Cholesky algorithm perfectly memorized.

One thing that might be interesting to point out is that it will be *identical* to the usual Cholesky decomposition if no problem (that would cause `None` to be returned by the default implementation) is encountered.

Also the Conjugate Gradient might be suggested as another alternative if the user just needs to solve a positive-definite system.

src/linalg/cholesky.rs
Fuuzetsu: There was none, except that we'd need a private method to not repeat the implementation with `new`. I added a private method now, PTAL.

Fuuzetsu:
>But if someone only wants to solve a linear system, then it would be good to point them in that direction perhaps?

I added a mention and a doc link to the `LU` in the crate.

>Yeah, that's pretty good already, although it's still a little ambiguous: For example it might read as if the substitute is put in place of any 0 value, whereas it's really only the ones that appear on the diagonal during factorization.

I updated the comment to specifical mention diagonals during decomposition.

>One thing that might be interesting to point out is that it will be identical to the usual Cholesky decomposition if no problem (that would cause None to be returned by the default implementation) is encountered.

I added a mention.

>Also the Conjugate Gradient might be suggested as another alternative if the user just needs to solve a positive-definite system.

I didn't add this because it feels out of place on `new_with_substitute`: if you already have positive-definite system so that CG is applicable, you shouldn't even be looking at `new_with_substitute`. It might have a place on `new` or `Cholesky` but this PR is only really concerned about `new_with_substitute`. I can add something if others disagree with me though.

nestordemeure:
> > Also the Conjugate Gradient might be suggested as another alternative if the user just needs to solve a positive-definite system.
> 
> I didn't add this because it feels out of place on `new_with_substitute`: if you already have positive-definite system so that CG is applicable, you shouldn't even be looking at `new_with_substitute`. It might have a place on `new` or `Cholesky` but this PR is only really concerned about `new_with_substitute`. I can add something if others disagree with me though.

`Cholesky` can fail on a badly conditionned but positive-definite system (due to numerical error producing spurious negative diagonal terms) whereas a conjugate gradient would usually keep running (but might not converge and produce garbage).
It is actually what we are seing with the Gaussian process.

Fuuzetsu:
> > > Also the Conjugate Gradient might be suggested as another alternative if the user just needs to solve a positive-definite system.
> > 
> > 
> > I didn't add this because it feels out of place on `new_with_substitute`: if you already have positive-definite system so that CG is applicable, you shouldn't even be looking at `new_with_substitute`. It might have a place on `new` or `Cholesky` but this PR is only really concerned about `new_with_substitute`. I can add something if others disagree with me though.
> 
> `Cholesky` can fail on a badly conditionned but positive-definite system (due to numerical error producing spurious negative diagonal terms) whereas a conjugate gradient would usually keep running (but might not converge and produce garbage).
> It is actually what we are seing with the Gaussian process.

I see. I can add the mention though `nalgebra` doesn't implement conjugate gradient method (that I can see) so it might be a bit confusing to suggest it.

nestordemeure:
> I can add the mention though `nalgebra` doesn't implement conjugate gradient method (that I can see) so it might be a bit confusing to suggest it.

Damn, you are right. Then you might has well not mention it.

src/linalg/cholesky.rs
```diff
@@ -74,6 +74,13 @@ where
         Cholesky { chol: matrix }
     }
 
+    /// Uses the given matrix as-is without any checks or modifications as the
+    /// Cholesky decomposition. It is up to the user to ensure and all
+    /// invariants hold.
```
Andlon, line 79: Very minor, but maybe put the second sentence in a new paragraph, since the first line is used as a "summary" in the docs. Also, the double "and" is removed, but I think maybe the "and" here also shouldn't be there...? That is, shouldn't it say "to ensure all invariants hold"?

Andlon (changes requested):
Looks good to me, just have one very minor remark about a doc comment that I think would be worth fixing!

src/linalg/cholesky.rs
Fuuzetsu: You're right, done.

Fuuzetsu:
As it looks like we're done iterating, I have squashed the various fix-ups back into a single commit. PTAL.

Andlon:
@Fuuzetsu: do you have any thoughts on a test case for this, as @sebcrozet suggested?

Fuuzetsu:
> @Fuuzetsu: do you have any thoughts on a test case for this, as @sebcrozet suggested?

I missed the request. I added something very simple, just to demonstrate that it "works". Not sure if there's a better test we should be doing.

Andlon:
Yeah, I also can't really think of anything substantially better. At least the code path is being called by the test suite now!

@sebcrozet: any other concerns before merging? Looks good to me at least.

sebcrozet:
Looks good! Thank you all!

Issue #984, opened 2021-09-08
The Mat3x3 new function has a bug?


DarkTong:
https://github.com/dimforge/nalgebra/blob/1bc919e0dbf2f7e83fab8aed37c1a06fe1b4fc93/nalgebra-glm/src/constructors.rs#L106-L110

sebcrozet:
Uh, yeah, that’s clearly wrong. Thank you for pointing this out!

loyd:
Why tests don't catch that?

PR #985, opened 2021-09-10
Attempt to fix issue #984

InnocentusLime:
Hello!

I have seen issue #984, noticed that it's a `good first issue`, decided to commit :)

What I did was simply swapping the two lines around.

EDIT
I have `grep`ed the whole rep and `mat3` doesn't seem to be called anywhere in the whole lib, so this change shouldn't break `nalgebra` itself


Andlon:
Hey @InnocentusLime, that's great, thanks for fixing this! Would you perhaps also be interested in adding a test case to make sure we don't make the same mistake again in the future? :)

sebcrozet:
I added a doc test. Thank you @InnocentusLime for fixing this.

PR #986, opened 2021-09-12
Concise Debug impls

Ralith:
Replace the verbose derived (or nearly equivalent) `Debug` impls for several newtypes with explicit impls that forward to the inner type, making readable diagnostics logging much easier.

I've found absolutely every time I'm trying to debug-log a nalgebra type I need to manually dig out the underlying array to have any hope of being able to scan a large amount of logged data to find irregularities. This is particularly awkward when I'm trying to log a structure that contains various nalgebra types, and hence can't just have `.coords.data` or whatever slapped on the end. This is also consistent with the existing `Debug` impls for e.g. `Orthographic3` and `Perspective3`, and with e.g. `Vec` in the stdlib.

In most cases, formatting is directly forwarded to the contents of each newtype. When a type wrapping a matrix is necessarily always wrapping a column vector, I instead forwarded to `as_slice()` as the extra layer of `[]` would add no information. I also considered extending the impl for `Matrix` to omit the outer `[]` for column vectors, but I'm not sure if that's too dynamic. Thoughts?

Before:
`Isometry { rotation: Unit { value: Quaternion { coords: Matrix { data: [[0.0, 0.0, 0.0, 1.0]] } } }, translation: Translation { vector: Matrix { data: [[0.0, 0.0, 0.0]] } } }`

After:
`Isometry { rotation: [0.0, 0.0, 0.0, 1.0], translation: [0.0, 0.0, 0.0] }`

### DIFF
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -195,10 +195,7 @@ pub struct Matrix<T, R, C, S> {
 
 impl<T, R: Dim, C: Dim, S: fmt::Debug> fmt::Debug for Matrix<T, R, C, S> {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        formatter
-            .debug_struct("Matrix")
-            .field("data", &self.data)
-            .finish()
+        self.data.fmt(formatter)
     }
 }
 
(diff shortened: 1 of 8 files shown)


Ralith:
CI failure looks unrelated.

Ralith:
Another big win here is that `dbg!(...)` output, which uses `{:#?}` form, is not only a lot easier to read, it takes *way* fewer lines, since every layer of newtype wrapping adds two additional lines and indentation.

sebcrozet:
Thanks!

PR #990, opened 2021-09-18
Remove some assertions via specialization

MaxVerevkin:


src/base/matrix.rs
```diff
@@ -53,8 +53,7 @@ pub type VectorSum<T, R1, R2> =
     Matrix<T, SameShapeR<R1, R2>, U1, SameShapeStorage<T, R1, U1, R2, U1>>;
 
 /// The type of the result of a matrix cross product.
-pub type MatrixCross<T, R1, C1, R2, C2> =
-    Matrix<T, SameShapeR<R1, R2>, SameShapeC<C1, C2>, SameShapeStorage<T, R1, C1, R2, C2>>;
+pub type MatrixCross<T, R, C> = Matrix<T, R, C, Owned<T, R, C>>;
```
MaxVerevkin, line 56: By the way, I'd like to ask some questions. Why do we need this type alias? I don't think it simplifies much.

src/base/matrix.rs
```diff
@@ -2019,108 +2018,90 @@ mod tests {
...
-        assert!(
-            (shape.0 == 3 && shape.1 == 1) || (shape.0 == 1 && shape.1 == 3),
-            "Vector cross product dimension mismatch: must be (3, 1) or (1, 3) but found {:?}.",
-            shape
-        );
+        unsafe {
+            // TODO: soooo ugly!
+            let nrows = SameShapeR::<U3, U3>::from_usize(3);
+            let ncols = SameShapeC::<U1, U1>::from_usize(1);
+            let mut res = Matrix::uninit(nrows, ncols);
```
MaxVerevkin, line 2050: Why does `Matrix::uninit()` take arguments?

MaxVerevkin:
Oh, I just realized that now `orthonormal_subspace_basis`, `perp` and `cross` don't work for dynamic vectors any more. Should `perp` and `cross` be implemented for dynamic vectors?

MaxVerevkin:
Given that current implementation of `perp()` contains a huge bug (it doesn't check for rhs shape) and this code runs:
```rust
let v1 = nalgebra::Vector2::<f32>::new_random();
let v2 = nalgebra::DVector::<f32>::new_random(1);
dbg!(&v1);
dbg!(&v2);
dbg!(v1.perp(&v2));
```
and doesn't panic: 
```
[src/main.rs:16] &v1 = Matrix {
    data: [
        [
            0.7467802,
            0.38603526,
        ],
    ],
}
[src/main.rs:17] &v2 = Matrix {
    data: VecStorage {
        data: [
            0.8369892,
        ],
        nrows: Dynamic {
            value: 1,
        },
        ncols: Const,
    },
}
[src/main.rs:18] v1.perp(&v2) = -0.32310736
```

I think it should be okay not to implement `perp` and `cross` for dynamic vectors.

PR #991, opened 2021-09-19
Fix UB in `Matrix::perp()`

MaxVerevkin:


sebcrozet:
Thank you for this PR. These assertions definitely need to be added.

However, the refactoring to avoid duplicate code isn’t valid because `.get_unchecked` will ignore the matrix stride. So if the input vectors are slices with non-default strides, this will result in an invalid output.

MaxVerevkin:
Thanks for your feedback @sebcrozet. I've rolled back that commit.

sebcrozet:
Thanks!

PR #992, opened 2021-09-20
exp.rs: factorial(): use precomputed factorial array

MaxVerevkin:
Gives a little performance improvement (-13% for `Matrix3`, -21% for `Matrix2`, -0% for `Matrix1`).


sebcrozet:
Thanks!

PR #993, opened 2021-09-25
deny unused_mut and unused_variables

MaxVerevkin:


sebcrozet:
Thanks!

Andlon:
I'm a bit late since this was already merged, but for lints like these, wouldn't it generally be better to deny warnings on CI than denying them outright? Often when developing I temporarily have many unused variables or similar, and it seems very annoying if this leads to intermediate compilation failures... Not a big deal since you can of course temporarily remove the `deny`, I don't know :shrug:

MaxVerevkin:
Maybe we can allow them in debug builds and deny in release? I'll check it it's possible.

Ralith:
+1 to making this sort of thing CI-only; applying strict linting to WIP code doesn't make a lot of sense.

PR #994, opened 2021-09-25
Deprecate clamp() in favour of standard functions

MaxVerevkin:
And by the way: why does `lib.rs` contain functions like `max()`, `min()`, `partial_cmp()`, `partial_lt()`, etc.?


sebcrozet:
The `clamp` function exists because there is no generic `clamp` function that accept both integers and float values.
However, I agree some of the other functions like the ones you mentioned are useless since the are equivalent to what the stdlib offers.

MaxVerevkin:
Thanks for clearifying.

PR #996, opened 2021-09-26
Simplify Matrix::is_identity while also improving performance

MaxVerevkin:
This benchmark shows ~4% performance improvement.

```rust
use nalgebra as na;
use rand::prelude::*;
use std::time::{Duration, Instant};

fn main() {
    let mut time = Duration::ZERO;
    const N: usize = 500_000;
    for _ in 0..N {
        time += bench();
    }
    dbg!(time);
}

fn bench() -> Duration {
    let mut rng = thread_rng();
    let m1 = na::DMatrix::<f32>::new_random(rng.gen_range(10..100), rng.gen_range(10..100));

    let timer = Instant::now();
    let _ = m1.is_identity(1.0);
    timer.elapsed()
}
```


sebcrozet:
Thanks!

Issue #957, opened 2021-07-28
Scalar trait unsoundness


Andlon:
It struck me the other day that the `Scalar` trait in its current form is responsible for a source of unsoundness.

In particular, since `Scalar` can be safely implemented, the following compiles:

```rust
#[derive(Clone, Debug, PartialEq)]
struct S;

impl Scalar for S {
    fn is<T: Scalar>() -> bool { true }
}
```

IIRC, `nalgebra` relies on this function in several cases to specialize for e.g. `f32` or `f64`, which may lead to undefined behavior with a `Scalar` implementation like the above.

I don't think this is a big problem, because I can't imagine there's anyone out there overriding `is`, but it is nonetheless unsound. The easiest way to fix this is either to move `is` out of `Scalar` or to mark `Scalar` `unsafe`.

sebcrozet:
That’s correct.  We should get rid of this method anyway since we can just compare `TypeId` directly instead.

MaxVerevkin:
This is solved now, isn't it?

sebcrozet:
That’s correct, it’s solved now. Closing.

PR #997, opened 2021-09-27
Remove `Dim::is()` function because it may be used in unsound ways

MaxVerevkin:
Very similar to #957. This part of `nalgebra` will cause UB for a poorly implemented `Dim`:  https://github.com/dimforge/nalgebra/blob/7f236d88aab0c8310d759720c2d8b13bee8e5aeb/src/base/blas.rs#L48-L57

By the way, I think it should be fine to remove those lines. Thanks to monomorphisation, this function will (probably) get optimized anyway https://github.com/dimforge/nalgebra/blob/7f236d88aab0c8310d759720c2d8b13bee8e5aeb/src/base/matrix_slice.rs#L170-L175

### DIFF
--- a/src/base/blas.rs
+++ b/src/base/blas.rs
@@ -7,7 +7,7 @@ use crate::base::blas_uninit::{axcpy_uninit, gemm_uninit, gemv_uninit};
 use crate::base::constraint::{
     AreMultipliable, DimEq, SameNumberOfColumns, SameNumberOfRows, ShapeConstraint,
 };
-use crate::base::dimension::{Const, Dim, Dynamic, U1, U2, U3, U4};
+use crate::base::dimension::{Const, Dim, Dynamic, U1};
 use crate::base::storage::{Storage, StorageMut};
 use crate::base::uninit::Init;
 use crate::base::{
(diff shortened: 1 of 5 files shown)


Ralith:
Alternatively, `Dim` could be made an unsafe trait.

Andlon:
> Alternatively, `Dim` could be made an unsafe trait.

I'd prefer this solution as removing it might unnecessarily break user code (iirc I also specialize for dimensions in some places). Besides, `D::is::<U2>()` has much nicer syntax than e.g. `TypeId::of::<D>() == TypeId::of::<U2>()`.

MaxVerevkin:
Okay, so instead of removing this function, I've created a new unsafe trait which is implemented for all `'static`types. I think it makes more sense because now it can be used with any type.

Ralith:
Wouldn't that still break the same user code?

sebcrozet:
I think it’s best to just make the `Dim` trait unsafe here. It sounds reasonable that unsafe code reading matrix dimensions should be able to rely on any of the `Dim` functions (like `Dim::value()`) to behave properly.

> By the way, I think it should be fine to remove those lines. Thanks to monomorphisation, this function will (probably) get optimized anyway

This needs to be benchmarked. It’s my experience that the compiler’s optimizer is annoyingly picky about what’s in the `blas` module, so it is not certain that the monomorphised code will really be optimized properly.

MaxVerevkin:
> This needs to be benchmarked

Well, I'm not sure how to do it, so I'll leave it as it is for now.

PR #1000, opened 2021-10-02
CSR/CSC: Provide constructor for unsorted but otherwise valid data

aarsenij:


nalgebra-sparse/src/csr.rs
```diff
@@ -170,6 +170,24 @@ impl<T> CsrMatrix<T> {
...
+        col_indices: Vec<usize>,
+        values: Vec<T>,
+    ) -> Result<Self, SparseFormatError> {
+        let sorted_num_cols: Vec<usize> = row_offsets[0..row_offsets.len() - 1]
+            .iter()
+            .enumerate()
+            .flat_map(|(index, &offset)| {
+                Self::sorted(col_indices[offset..row_offsets[index + 1]].to_vec())
+            })
+            .collect();
```
Andlon, line 187: Here we need to keep in mind that the input we receive has *not* been validated. In particular, we *cannot* assume that `row_offsets.len() > 0` and that `col_index[offset..row_offsets[index + 1]]` won't panic. So we need to do the same input validation as `::try_from_csr_data`, with the exception of the requirement that column indices need to be sorted.

Moreover, we allocate a `Vec` in each `flat_map` invocation, and similarly the whole result is collected (so the whole thing is allocated again). This seems a little unnecessary: after all we don't need any more storage, we only need to sort subcollections in the input array.

So instead of using `.flat_map` and `.collect()` you could instead directly sort the corresponding entries in `col_indices`, i.e. something like
```rust
for (index, &offset) in ... {
   col_indices[offset .. row_offsets[index + 1]).sort_unstable();
}
```

nalgebra-sparse/src/csr.rs
```diff
@@ -170,6 +170,24 @@ impl<T> CsrMatrix<T> {
         Self::try_from_pattern_and_values(pattern, values)
     }
 
+    /// Try to construct a CSR matrix from raw CSR data with unsorted columns.
```
Andlon, line 173: I think the documentation should explain the invariants and error conditions like the documentation for `::try_from_csr_data`

nalgebra-sparse/src/csr.rs
```diff
@@ -190,6 +208,15 @@ impl<T> CsrMatrix<T> {
...
     }
 
+    /// Return sorted vector.
+    #[inline]
+    #[must_use]
+    pub fn sorted(row_offsets: Vec<usize>) -> Vec<usize> {
+        let mut sorted = row_offsets.clone();
+        sorted.sort();
+        return sorted;
+    }
```
Andlon, line 218: this should probably not be a method of `CsrMatrix` as it does not require any access to `self` and seems unrelated to a `CsrMatrix` in general - and certainly not a public method.

Andlon (changes requested):
Thanks for looking into this @aarsenij! I think we're off to a good start here. I have some concerns: see below!

nalgebra-sparse/src/csr.rs
aarsenij: Thank you for feedback @Andlon!
I've changed my initial proposal according to your concerns. I hope it will bring the PR closer to a merge.
For example, I created a `first_and_last_offsets_are_ok` function to the base::helper which can be used in order to check whether the row_offsets is valid. But I was still not sure if it would be the best place for a new helper function.

src/base/helper.rs
```diff
@@ -29,3 +29,15 @@ where
...
+#[inline]
+#[must_use]
+pub fn first_and_last_offsets_are_ok(
+    major_offsets: &Vec<usize>,
+    minor_indices: &Vec<usize>,
+) -> bool {
+    let first_offset_ok = *major_offsets.first().unwrap() == 0;
+    let last_offset_ok = *major_offsets.last().unwrap() == minor_indices.len();
+    return first_offset_ok && last_offset_ok;
+}
```
Andlon, line 43: In general helper functionality should reside somewhere in `nalgebra-sparse`, since `nalgebra` itself does not have any knowledge of sparse matrices.

Andlon (changes requested):
Hey @aarsenij, thanks for addressing my concerns. I realize now in my initial review that I was a little too hasty and forgot to comment on some very important details. First, let's take a step back:

If a CSR matrix has unsorted - but otherwise valid - column indices, then it's not sufficient to sort only the column indices, but we have to permute the *values* array as well. So in general we have to compute a permutation of the column indices and apply it to both the column indices *and* the values.

The *best* way to do this would probably be to do sorting at the same time as checking the pattern (see the implementation of `SparsityPattern::try_from_offsets_and_indices`), but I think we can get away with a much simpler approach, very similar to what you have done already.

One way to do this is the following algorithm:

for each row `i` (indicated by `row_offsets`) with `n_i` entries:
   1. let `col_idx_in_row` be the `n_i` column indices in row `i` (NB! Need to check that offsets are not out of bounds w.r.t. column indices, and that they give a valid index range in the column indices array).
   2. Initialize the identity permutation `p = [0, ..., n - 1]`
   3. Sort `p` by key `col_idx_in_row[p[i]]`
   4. Copy over `col_idx_in_row` to a temporary buffer
   5. Permute `col_idx_in_row` by the permutation `p`

After this we can pass our permuted data into `CsrMatrix::try_from_csr_data` (which will detect other errors, such as out of bounds column indices). Note that we still have to ensure that the `col_indices` array and `values` array have the same length.

I hope that makes sense? Let me know what you think!

Andlon:
Oh, and: please don't force push your changes! It destroys the history and makes it harder (or impossible) to keep track of what was changed in each commit, thus making it harder for me to review.

PR #1004, opened 2021-10-06
fix linear indexing of matrix slices

### LINKED ISSUE #750: Linear indexing of a matrix slice does not account for stride.
By "linear indexing", I mean using a `usize` index to access the elements of the matrix in column-major order, via `Matrix::index::<usize>`. The implementation of `MatrixIndex` for `usize` doesn't account for the stride of the underlying matrix storage; it [calls `Storage::get_unchecked_linear`](https://github.com/rustsim/nalgebra/blob/2ab82befe4b2fd8826d307048939902089f48cb3/src/base/indexing.rs#L577) directly, which then uses this index as an [offset into the storage buffer](https://github.com/rustsim/nalgebra/blob/2ab82befe4b2fd8826d307048939902089f48cb3/src/base/storage.rs#L75). 

This works fine for matrices whose strides are equal to 1 (their storage is contiguous), but for other (sub)matrices, this produces an incorrect result. Here's an example case demonstrating what happens ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5621ecfc2709167479978eecce70e3eb)):

```rust
use nalgebra::*;

fn main() {
    let mat = DMatrix::from_row_slice(3, 3, &[
        0.0, 3.0, 6.0,
        1.0, 4.0, 7.0,
        2.0, 5.0, 8.0,
    ]);
    
    // prints each value in row 1 - [1, 4, 7]
    for i in 0..3 {
        print!("{} ", mat.index((1, i)));
    }
    println!();
    
    // get a reference to row 1
    let submat = mat.index((1, ..));
    // should print the same as above,
    // instead prints contiguous values from underlying buffer,
    // not accounting for stride - [1, 2, 3]
    for i in 0..3 {
        print!("{} ", submat.index(i));
    }
    println!();
}
```

A workaround that produces the expected behavior is to just use `(row, col)` indexing - use `(0, i)` instead of `i` to access the elements of the submatrix in the example above.

To fix this, some modular arithmetic is needed - compute the row and column from the index `i` and pass those to `Storage::get_unchecked` instead which does properly apply the stride:

```rust
let (row, col) = (i / nrows, i % nrows)
```

Alternatively, if this is too much overhead, then stricter trait bounds can be added for using `usize` as an index, so that the matrix storage has to be contiguous - note this is a breaking change:

```rust
impl<...> MatrixIndex<...> for usize
where
    S: ContiguousStorage<...>
```

MaxVerevkin:
Closes #750

### DIFF
--- a/src/base/indexing.rs
+++ b/src/base/indexing.rs
@@ -567,7 +567,10 @@ where
     #[doc(hidden)]
     #[inline(always)]
     unsafe fn get_unchecked(self, matrix: &'a Matrix<T, R, C, S>) -> Self::Output {
-        matrix.data.get_unchecked_linear(self)
+        let nrows = matrix.shape().0;
+        let row = self % nrows;
+        let col = self / nrows;
+        matrix.data.get_unchecked(row, col)
     }
 }
 
--- a/src/base/matrix_slice.rs
+++ b/src/base/matrix_slice.rs
@@ -73,7 +73,6 @@ macro_rules! slice_storage_impl(
                       S: $Storage<T, RStor, CStor>,
                       RStride: Dim,
                       CStride: Dim {
-
                 $T::from_raw_parts(storage.$get_addr(start.0, start.1), shape, strides)
             }
         }


sebcrozet:
Thanks!

PR #1007, opened 2021-10-13
WIP: Add Scaling type to represent non-uniform scale transformations

Yuri6037:
This is my first PR on this project, so not sure if I properly have respected the conventions here. My IDE actually broke while coding this: My IDE cannot figure out impls of OScaling with Const<somevalue_usize>.

I've made this PR as requested in "Use for BP3D" issue. In this issue it has been acknowledged a missing feature to support non-uniform scales.
I write this PR in order to simplify working with non-uniform scales by introducing a new type called "Scaling". This new type allows direct multiplication with vectors and points. When multiplied with a point, the result is a scaled point. When multiplied by a Vector or another Scaling the result is a new Scaling which represents a cumulation of both scales.

Any review on this is appreciated. Should we also add an overload for multiply in Transform to allow multiplying with a Scaling?

EDIT: Should we have a set of operators (*, *=, /, /=) for Scaling with a single scalar so that we can keep support for uniform scaling within Scaling?

### DIFF
--- a/src/geometry/scaling.rs
+++ b/src/geometry/scaling.rs
@@ -1,14 +1,20 @@
 use crate::{SVector, Scalar};
 
 /// A scaling represents a non-uniform scale transformation
-pub struct Scaling<T: Scalar, const D: usize>(pub SVector<T, D>);
+pub struct Scaling<T: Scalar, const D: usize>
+{
+    /// The inner vector
+    pub vector: SVector<T, D>
+}
 
 impl<T, const D: usize> From<SVector<T, D>> for Scaling<T, D>
     where T: Scalar
 {
     fn from(other: SVector<T, D>) -> Self
     {
-        return Scaling::<T, D>(other);
+        return Scaling {
+            vector: other
+        };
     }
 }
 
@@ -17,6 +23,6 @@ impl<T, const D: usize> Into<SVector<T, D>> for Scaling<T, D>
 {
     fn into(self) -> SVector<T, D>
     {
-        return self.0;
+        return self.vector;
     }
 }
--- a/src/geometry/scaling_construction.rs
+++ b/src/geometry/scaling_construction.rs
@@ -6,7 +6,9 @@ impl<T: Scalar> Scaling<T, 1>
     /// Initializes this scaling from its components.
     pub fn new(x: T) -> Scaling<T, 1>
     {
-        return Scaling(SVector::<T, 1>::new(x));
+        return Scaling {
+            vector: SVector::<T, 1>::new(x)
+        };
     }
 }
 
@@ -15,7 +17,9 @@ impl<T: Scalar> Scaling<T, 2>
     /// Initializes this scaling from its components.
     pub fn new(x: T, y: T) -> Scaling<T, 2>
     {
-        return Scaling(SVector::<T, 2>::new(x, y));
+        return Scaling {
+            vector: SVector::<T, 2>::new(x, y)
+        };
     }
 }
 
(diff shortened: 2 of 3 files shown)


Yuri6037:
> Thank you for this PR, this is a great start! Some remarks (to make it coherent with the rest of `nalgebra`.
> 
>     * All the trait implementations for operator overloading should go in a file called `scaling_ops.rs`
> 
>     * All the functions for creating a `Scaling` (i.e. the `fn new`) should go in a file called `scaling_construction.rs`.
> 
>     * There is no need to introduce the `OScaling` type. I assume you took some inspiration from the `OPoint` type. This `OPoint` only exists for backward compatibility (from a time whene const-generics didn’t exist). So for `Scaling` you can directly define:
> 
> 
> ```rust
> struct Scaling<T, const D: usize> {
>     pub vector: SVector<T, D>
> }
> ```
> 
> Overall, I strongly suggest starting again from scratch by doing the following in order to implement `Scaling` very quickly and more comprehensively:
> 
>     1. Copy all the files with a name beginning with `translation` (e.g. `translation.rs`, `translation_construction.rs`, etc.); replacing `translation` to `scaling` on these file names. The `Translation` impl looks a lot like `Scaling` so it will be a great stating point.
> 
>     2. On all these copied files replace all occurrences of `Translation` to `Scaling`, of `translation` to `scaling` etc.
> 
>     3. Adjust the few math impls (e.g., use a vector filled with 1.0 instead of 0.0 for the identity, use componentwise multiplication instead of addition, etc.)
> 
>     4. Adjust the doc-tests.

Thanks for your answer, I prefer to not start from scratch. I will follow what you said about the conventions I had no idea about these conventions.

About the OScaling. I did that because I could not find any way to be compatible with points. If you want to multiply a scaling with a point then I figured that rustc refused at all costs to convert any type of vector to a point except this OVector.
I don't know how I can use SVector. If you know a way to turn SVector into OPoint then I'd appreciate and this would allow using SVector instead of OVector.

sebcrozet:
> Thanks for your answer, I prefer to not start from scratch. I will follow what you said about the conventions I had no idea about these conventions.

Please, do restart using the methodology I suggested. If you wont, you will:
1. Miss a lot of functions that should exist for `Scaling` (`identity`,  `inverse`, `to_homogeneous`, etc.)
2. Miss a lot of trait impls (serialization traits, coordinate view traits, etc.)
3. Will have to rewrite all the documentation and doc-tests manually.
4. The `Scaling` struct should have a `vector` field, just like `Translation`, instead of being a newtype.

This will also make it much easier to maintain in the future because the code structure will remain similar everywhere; and it will make it easier for me to help if you have any problem because I already know the way it looks like overall.

----

Remarks regarding the operations:
- `Scaling * Vector` should return a `Vector` (just like `Scaling * Point` returns a `Point`).
- `Scaling / Vector` or `Scaling /  Point` shouldn’t be allowed as they don’t make any sense. The mathematically sound way of doing this would be `scaling.inverse() * vector`, or `scaling.inverse() * point`.
- `Scaling *= Vector` shouldn’t exist because it doesn’t make sense (because `Scaling * Vector` returns a `Vector`, not a `Scaling`).

sebcrozet:
> And even if we use Scaling * Scaling, then I maintain that Scaling *= Scaling makes sense. Why shouldn't it be?

`Scaling *= Scaling` is perfectly fine. It was `Scaling *= Vector` that I said didn’t make sense.

> A vector can represent a scale but if you prefer using Scaling * Scaling then in this case Scaling * Vector should not exist and be replaced by a .into() to convert the scaling as a vector.

Why shouldn’t `Scaling * Vector` exist? Say you have a vector `{ 1, 2, 3 }` and you need to scale it’s magnitude by `10`. Then you just do `10 * { 1, 2, 3 } = { 10, 20, 30 }`. So the same makes sense for non-uniform scaling: `Scaling(10, 100, 1000) * { 1, 2, 3 } = { 10, 200, 3000 }`.

PR #1012, opened 2021-10-19
Scale

Yuri6037:
New PR for non uniform scaling. This is the new replacement PR for https://github.com/dimforge/nalgebra/pull/1007.

Per my last 2 comments, I'll name the type Scale.

### DIFF
--- a/src/geometry/scale_conversion.rs
+++ b/src/geometry/scale_conversion.rs
@@ -102,7 +102,7 @@ where
     fn from_superset_unchecked(
         m: &OMatrix<T2, DimNameSum<Const<D>, U1>, DimNameSum<Const<D>, U1>>,
     ) -> Self {
-        let v = m.fixed_slice::<D, D>(0, 0).into_owned().diagonal();
+        let v = m.fixed_slice::<D, D>(0, 0).diagonal();
         Self {
             vector: crate::convert_unchecked(v),
         }


Yuri6037:
> This is a great start, thank you!
> 
> A decision has to be taken here. If a scaling contains a 0, then it becomes non inversible. This means that we can’t have a `fn inverse()` or `fn inverse_mut()` method. Instead, we need to have:
> 
>     * `fn try_inverse(&self) -> Option<Scale<T, D>>` which returns `None` if any of the component of `self` is zero.
> 
>     * `fn try_inverse_mut(&mut self) -> bool` which returns `false` if the inversion can’t happen because the scaling contains a zero.
> 
> 
> And we could also consider adding (if that sounds useful):
> 
>     * `fn inverse_unchecked(&self) -> Scale<T, D>` which does what this method does (i.e. inverse without any check.)
> 
>     * `fn pseudo_inverse(&self) -> Scale<T, D>` which only inverts the non-zero component of `scale`.
> 
> 
> An alternative would be to disallow instances of `Scale` that contain a zero. In that case, we guarantee that inversion is always valid. This implies that the `Scale::new` constructors should check that the input values don’t have any zero. Also, a `Scale::new_unchecked` constructor that skips these checks could be added.
> 
> So the question is, does it sound reasonable to say that a `Scale` should not contain any zero in the vast majority of use-cases?

The answer to this question is obviously that a scale may be 0. What if you have a mod that wants to make some kind of dissolve effect on a model. The way you'd do this among some shader work is to subtract some very small value times the delta time out of the scale which would be initialzed at (1, 1, 1). Then inevitably your scale would eventually reach 0 at which point the mod could do `if scale <= 0 then object:Destroy()`.

Issue #879, opened 2021-04-29
CSR/CSC: Provide constructor for unsorted but otherwise valid data


Andlon:
The main constructor from existing CSR/CSC data is e.g. [try_from_csr_data](https://docs.rs/nalgebra-sparse/0.2.0/nalgebra_sparse/csr/struct.CsrMatrix.html#method.try_from_csr_data). However, this deliberately rejects unsorted input. However, if the only issue with the data is that the column indices are not sorted, then this is fixable. We should provide a similarly named `try_from_unsorted_csr_data` which would sort the column indices when needed. Similarly for `CscMatrix`, as well as an analogous constructor for `SparsityPattern`.

PR #1015, opened 2021-10-25
CSC: Create constructor for unsorted but otherwise valid data

aarsenij:
related to https://github.com/dimforge/nalgebra/issues/879

- [x] try to port tests from #1019 for cases when `range_start`/`range_end`  are out of bounds

- [x] avoid allocations on "happy path"

### DIFF
--- a/nalgebra-sparse/src/convert/serial.rs
+++ b/nalgebra-sparse/src/convert/serial.rs
@@ -14,6 +14,7 @@ use crate::coo::CooMatrix;
 use crate::cs;
 use crate::csc::CscMatrix;
 use crate::csr::CsrMatrix;
+use crate::utils::apply_permutation;
 
 /// Converts a dense matrix to [`CooMatrix`].
 pub fn convert_dense_coo<T, R, C, S>(dense: &Matrix<T, R, C, S>) -> CooMatrix<T>
@@ -388,15 +389,6 @@ fn sort_lane<T: Clone>(
 
     apply_permutation(minor_idx_result, minor_idx, permutation);
     apply_permutation(values_result, values, permutation);
-}
-
-// TODO: Move this into `utils` or something?
-fn apply_permutation<T: Clone>(out_slice: &mut [T], in_slice: &[T], permutation: &[usize]) {
-    assert_eq!(out_slice.len(), in_slice.len());
-    assert_eq!(out_slice.len(), permutation.len());
-    for (out_element, old_pos) in out_slice.iter_mut().zip(permutation) {
-        *out_element = in_slice[*old_pos].clone();
-    }
 }
 
 /// Given *sorted* indices and corresponding scalar values, combines duplicates with the given
--- a/nalgebra-sparse/src/convert/serial.rs
+++ b/nalgebra-sparse/src/convert/serial.rs
@@ -218,7 +218,7 @@ where
 /// Converts a [`CscMatrix`] to a [`CsrMatrix`].
 pub fn convert_csc_csr<T>(csc: &CscMatrix<T>) -> CsrMatrix<T>
 where
-    T: Scalar + Zero,
+    T: Scalar,
 {
     let (offsets, indices, values) = cs::transpose_cs(
         csc.ncols(),
--- a/nalgebra-sparse/src/convert/serial.rs
+++ b/nalgebra-sparse/src/convert/serial.rs
@@ -218,7 +218,7 @@ where
 /// Converts a [`CscMatrix`] to a [`CsrMatrix`].
 pub fn convert_csc_csr<T>(csc: &CscMatrix<T>) -> CsrMatrix<T>
 where
-    T: Scalar,
+    T: Scalar + Zero,
 {
     let (offsets, indices, values) = cs::transpose_cs(
         csc.ncols(),
--- a/nalgebra-sparse/src/convert/serial.rs
+++ b/nalgebra-sparse/src/convert/serial.rs
@@ -14,7 +14,7 @@ use crate::coo::CooMatrix;
 use crate::cs;
 use crate::csc::CscMatrix;
 use crate::csr::CsrMatrix;
-use crate::utils::apply_permutation;
+use crate::utils::{apply_permutation, compute_sort_permutation};
 
 /// Converts a dense matrix to [`CooMatrix`].
 pub fn convert_dense_coo<T, R, C, S>(dense: &Matrix<T, R, C, S>) -> CooMatrix<T>
@@ -377,15 +377,7 @@ fn sort_lane<T: Clone>(
     assert_eq!(values.len(), workspace.len());
 
     let permutation = workspace;
-    // Set permutation to identity
-    for (i, p) in permutation.iter_mut().enumerate() {
-        *p = i;
-    }
-
-    // Compute permutation needed to bring minor indices into sorted order
-    // Note: Using sort_unstable here avoids internal allocations, which is crucial since
-    // each lane might have a small number of elements
-    permutation.sort_unstable_by_key(|idx| minor_idx[*idx]);
+    compute_sort_permutation(permutation, minor_idx);
 
     apply_permutation(minor_idx_result, minor_idx, permutation);
     apply_permutation(values_result, values, permutation);
--- a/nalgebra-sparse/src/cs.rs
+++ b/nalgebra-sparse/src/cs.rs
@@ -601,6 +601,9 @@ where
         }
     }
 
+    // Set up required buffers up front
+    let mut minor_idx_buffer: Vec<usize> = Vec::new();
+    let mut values_buffer: Vec<T> = Vec::new();
     let mut minor_index_permutation: Vec<usize> = Vec::new();
 
     // Test that each lane has strictly monotonically increasing minor indices, i.e.
@@ -630,10 +633,6 @@ where
             // to ensure that we only visit each minor index once
             let mut prev = None;
             let mut nonmonotonic = false;
-
-            // Set up *all* buffers up front (i.e. at the beginning of function, e.g. `Vec::new()`
-            let mut minor_idx_buffer: Vec<usize> = Vec::new();
-            let mut values_buffer: Vec<T> = Vec::new();
 
             for &minor_idx in minor_idx_in_lane {
                 if minor_idx >= minor_dim {
--- a/nalgebra-sparse/src/cs.rs
+++ b/nalgebra-sparse/src/cs.rs
@@ -560,7 +560,9 @@ where
     let mut values_option = values;
     if values_option.is_none() {
         if sort {
-            unreachable!("Internal error: Sorting currently not supported if no values are present.");
+            unreachable!(
+                "Internal error: Sorting currently not supported if no values are present."
+            );
         }
     }
 
--- a/nalgebra-sparse/src/cs.rs
+++ b/nalgebra-sparse/src/cs.rs
@@ -1,4 +1,3 @@
-use std::iter::FromIterator;
 use std::mem::replace;
 use std::ops::Range;
 
@@ -212,15 +211,7 @@ impl<T> CsMatrix<T> {
         )
         .expect("Internal error: Sparsity pattern must always be valid.");
 
-        // permute values
-        let sorted_values: Vec<T> = Vec::from_iter(
-            (new_pattern
-                .minor_index_permutation
-                .iter()
-                .map(|i| &new_values[*i]))
-            .cloned(),
-        );
-        Self::from_pattern_and_values(new_pattern, sorted_values)
+        Self::from_pattern_and_values(new_pattern, new_values)
     }
 
     /// Returns the diagonal of the matrix as a sparse matrix.
--- a/nalgebra-sparse/src/cs.rs
+++ b/nalgebra-sparse/src/cs.rs
@@ -1,3 +1,4 @@
+use std::iter::FromIterator;
 use std::mem::replace;
 use std::ops::Range;
 
@@ -211,7 +212,15 @@ impl<T> CsMatrix<T> {
         )
         .expect("Internal error: Sparsity pattern must always be valid.");
 
-        Self::from_pattern_and_values(new_pattern, new_values)
+        // permute values
+        let sorted_values: Vec<T> = Vec::from_iter(
+            (new_pattern
+                .minor_index_permutation
+                .iter()
+                .map(|i| &new_values[*i]))
+            .cloned(),
+        );
+        Self::from_pattern_and_values(new_pattern, sorted_values)
     }
 
     /// Returns the diagonal of the matrix as a sparse matrix.
--- a/nalgebra-sparse/src/cs.rs
+++ b/nalgebra-sparse/src/cs.rs
@@ -655,10 +655,7 @@ where
                 minor_index_permutation.resize(range_size, 0);
                 compute_sort_permutation(&mut minor_index_permutation, &minor_idx_in_lane);
                 minor_idx_buffer.clear();
-                minor_idx_buffer.reserve(range_size);
-                for &value in minor_idx_in_lane.iter() {
-                    minor_idx_buffer.push(value);
-                }
+                minor_idx_buffer.extend_from_slice(&minor_idx_in_lane);
                 apply_permutation(
                     &mut minor_indices[range_start..range_end],
                     &minor_idx_buffer,
@@ -671,15 +668,14 @@ where
                         panic!(
                             "Internal error: Sorting currently not supported if values are empty."
                         );
-                    }
-                    if minor_indices.len() != values.len() {
-                        panic!("Number of values and minor indices must be the same.");
+                    } else if minor_indices.len() != values.len() {
+                        return Err(SparseFormatError::from_kind_and_msg(
+                            SparseFormatErrorKind::InvalidStructure,
+                            "Number of values and minor indices must be the same.",
+                        ));
                     }
                     values_buffer.clear();
-                    values_buffer.reserve(range_size);
-                    for index in range_start..range_end {
-                        values_buffer.push(values[index].clone());
-                    }
+                    values_buffer.extend_from_slice(&values[range_start..range_end]);
                     apply_permutation(
                         &mut values[range_start..range_end],
                         &values_buffer,
--- a/nalgebra-sparse/src/cs.rs
+++ b/nalgebra-sparse/src/cs.rs
@@ -545,7 +545,7 @@ pub fn convert_counts_to_offsets(counts: &mut [usize]) {
     }
 }
 
-/// Validates cs data, sorts minor indices and values
+/// Validates cs data, optionally sorts minor indices and values
 pub(crate) fn validate_and_optionally_sort_cs_data<T>(
     major_dim: usize,
     minor_dim: usize,
@@ -569,12 +569,10 @@ where
             value_refs = values;
         }
         None => {
-            if sort && minor_indices.len() > 0 {
-                return Err(SparseFormatError::from_kind_and_msg(
-                    SparseFormatErrorKind::InvalidStructure,
-                    "No values provided for sorting.",
-                ));
-            }
+            assert!(
+                sort && minor_indices.len() > 0,
+                "No values provided for sorting."
+            );
         }
     }
 
@@ -630,12 +628,11 @@ where
 
             // We test for in-bounds, uniqueness and monotonicity at the same time
             // to ensure that we only visit each minor index once
-            let mut iter = minor_idx_in_lane.iter();
             let mut prev = None;
             let mut nonmonotonic = false;
 
-            while let Some(next) = iter.next().copied() {
-                if next >= minor_dim {
+            for &minor_idx in minor_idx_in_lane {
+                if minor_idx >= minor_dim {
                     return Err(SparseFormatError::from_kind_and_msg(
                         SparseFormatErrorKind::IndexOutOfBounds,
                         "A minor index is out of bounds.",
@@ -643,17 +640,23 @@ where
                 }
 
                 if let Some(prev) = prev {
-                    if prev == next {
+                    if prev == minor_idx {
                         return Err(SparseFormatError::from_kind_and_msg(
                             SparseFormatErrorKind::DuplicateEntry,
                             "Input data contains duplicate entries.",
                         ));
                     }
-                    if prev > next {
+                    if prev > minor_idx {
+                        if !sort {
+                            return Err(SparseFormatError::from_kind_and_msg(
+                                SparseFormatErrorKind::InvalidStructure,
+                                "Minor indices are not monotonically increasing within each lane.",
+                            ));
+                        }
                         nonmonotonic = true;
                     }
                 }
-                prev = Some(next);
+                prev = Some(minor_idx);
             }
 
             // sort if indices are nonmonotonic and sorting is expected
--- a/nalgebra-sparse/src/cs.rs
+++ b/nalgebra-sparse/src/cs.rs
@@ -1,8 +1,7 @@
-use std::iter::FromIterator;
 use std::mem::replace;
 use std::ops::Range;
 
-use num_traits::One;
+use num_traits::{One, Zero};
 
 use nalgebra::Scalar;
 
@@ -555,7 +554,7 @@ pub(crate) fn validate_and_optionally_sort_cs_data<T>(
     sort: bool,
 ) -> Result<(), SparseFormatError>
 where
-    T: Scalar,
+    T: Scalar + Zero,
 {
     let mut value_refs: &mut [T] = &mut Vec::new();
     match values {
@@ -631,6 +630,10 @@ where
             let mut prev = None;
             let mut nonmonotonic = false;
 
+            // Set up *all* buffers up front (i.e. at the beginning of function, e.g. `Vec::new()`
+            let mut minor_idx_buffer: Vec<usize> = Vec::new();
+            let mut values_buffer: Vec<T> = Vec::new();
+
             for &minor_idx in minor_idx_in_lane {
                 if minor_idx >= minor_dim {
                     return Err(SparseFormatError::from_kind_and_msg(
@@ -672,16 +675,19 @@ where
                     let y = &minor_idx_in_lane[*b - range_start];
                     x.partial_cmp(y).unwrap()
                 });
-                let unsorted_indices: Vec<usize> = minor_indices[range_start..range_end]
-                    .iter()
-                    .map(|i| i.clone())
-                    .collect();
-                let unsorted_values: Vec<T> =
-                    Vec::from_iter(((range_start..range_end).map(|i| &value_refs[i])).cloned());
+
+                minor_idx_buffer.resize(range_end, 0);
+                values_buffer.resize(range_end, T::zero());
+                for (index, &value) in minor_indices[range_start..range_end].iter().enumerate() {
+                    minor_idx_buffer[index + range_start] = value;
+                    let v: T = value_refs[index + range_start].clone();
+                    values_buffer[index + range_start] = v;
+                }
+                
                 for (index, &offset) in minor_index_permutation.iter().enumerate() {
-                    let i: usize = unsorted_indices[offset - range_start];
+                    let i: usize = minor_idx_buffer[offset];
                     minor_indices[index + range_start] = i;
-                    let v: T = unsorted_values[offset - range_start].clone();
+                    let v: T = values_buffer[offset].clone();
                     value_refs[index + range_start] = v;
                 }
             }
--- a/nalgebra-sparse/src/cs.rs
+++ b/nalgebra-sparse/src/cs.rs
@@ -6,7 +6,7 @@ use num_traits::{One, Zero};
 use nalgebra::Scalar;
 
 use crate::pattern::SparsityPattern;
-use crate::utils::apply_permutation;
+use crate::utils::{apply_permutation, compute_sort_permutation};
 use crate::{SparseEntry, SparseEntryMut, SparseFormatError, SparseFormatErrorKind};
 
 /// An abstract compressed matrix.
@@ -545,23 +545,6 @@ pub fn convert_counts_to_offsets(counts: &mut [usize]) {
     }
 }
 
-fn compute_sort_permutation(minor_index_permutation: &mut [usize], minor_idx_in_lane: &[usize]) {
-    assert_eq!(
-        minor_index_permutation.len(),
-        minor_idx_in_lane.len(),
-        "Length of permutation slice and number of indices must be the same."
-    );
-    for (index, value) in (0..minor_idx_in_lane.len()).enumerate() {
-        minor_index_permutation[index] = value;
-    }
-    // sort permutation values for monotonic index order
-    minor_index_permutation.sort_by(|a, b| {
-        let x = &minor_idx_in_lane[*a];
-        let y = &minor_idx_in_lane[*b];
-        x.partial_cmp(y).unwrap()
-    });
-}
-
 /// Validates cs data, optionally sorts minor indices and values
 pub(crate) fn validate_and_optionally_sort_cs_data<T>(
     major_dim: usize,
@@ -682,33 +665,28 @@ where
 
             // sort if indices are nonmonotonic and sorting is expected
             if nonmonotonic && sort {
-                minor_index_permutation.resize(range_end - range_start, 0);
-                compute_sort_permutation(
-                    &mut minor_index_permutation,
-                    &minor_indices[range_start..range_end],
-                );
-
-                minor_idx_buffer.resize(range_end, 0);
-
-                for (index, &value) in minor_indices[range_start..range_end].iter().enumerate() {
-                    minor_idx_buffer[index + range_start] = value;
+                let range_size = range_end - range_start;
+                minor_index_permutation.resize(range_size, 0);
+                compute_sort_permutation(&mut minor_index_permutation, &minor_idx_in_lane);
+                minor_idx_buffer.resize(range_size, 0);
+                for (index, &value) in minor_idx_in_lane.iter().enumerate() {
+                    minor_idx_buffer[index] = value;
                 }
-
                 apply_permutation(
                     &mut minor_indices[range_start..range_end],
-                    &minor_idx_buffer[range_start..range_end],
+                    &minor_idx_buffer,
                     &minor_index_permutation,
                 );
 
+                // sort values if they exist
                 if !value_refs.is_empty() {
-                    values_buffer.resize(range_end, T::zero());
+                    values_buffer.resize(range_size, T::zero());
                     for index in range_start..range_end {
-                        let v: T = value_refs[index].clone();
-                        values_buffer[index] = v;
+                        values_buffer[index - range_start] = value_refs[index].clone();
                     }
                     apply_permutation(
                         &mut value_refs[range_start..range_end],
-                        &values_buffer[range_start..range_end],
+                        &values_buffer,
                         &minor_index_permutation,
                     );
                 }
--- a/nalgebra-sparse/src/cs.rs
+++ b/nalgebra-sparse/src/cs.rs
@@ -558,14 +558,21 @@ where
     T: Scalar,
 {
     let mut values_option = values;
-    if values_option.is_none() {
-        if sort {
+
+    if sort {
+        if let Some(values) = values_option.as_mut() {
+            if minor_indices.len() != values.len() {
+                return Err(SparseFormatError::from_kind_and_msg(
+                    SparseFormatErrorKind::InvalidStructure,
+                    "Number of values and minor indices must be the same.",
+                ));
+            }
+        } else {
             unreachable!(
                 "Internal error: Sorting currently not supported if no values are present."
             );
         }
     }
-
     if major_offsets.len() == 0 {
         return Err(SparseFormatError::from_kind_and_msg(
             SparseFormatErrorKind::InvalidStructure,
@@ -637,7 +644,7 @@ where
                         if !sort {
                             return Err(SparseFormatError::from_kind_and_msg(
                                 SparseFormatErrorKind::InvalidStructure,
-                                "Minor indices are not monotonically increasing within each lane while sorting is not expected.",
+                                "Minor indices are not strictly monotonically increasing in each lane.",
                             ));
                         }
                         monotonic = false;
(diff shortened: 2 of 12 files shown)


aarsenij:
With the approach you proposed, it looks like we would need to provide mutability of `indices` and `values` in `try_from_csc_data` or we would need to copy the values in order to make them mutable in the new validation function.
Otherwise, I get "Cannot borrow `row_indices` as mutable, as it is not declared as mutable" error. Not sure which path should be taken here...

Andlon:
@aarsenij: if I understand what you mean correctly, then `try_from_csc_data` already takes the input arrays by value - so we can add a `mut` qualifier on them without changing the signature for *users*.

That is:
```rust
pub fn try_from_csc_data(
    num_rows: usize,
    num_cols: usize,
    col_offsets: Vec<usize>,
    row_indices: Vec<usize>,
    values: Vec<T>
) -> Result<Self, SparseFormatError> {
```
is exactly the same as
```rust
pub fn try_from_csc_data(
    num_rows: usize,
    num_cols: usize,
    mut col_offsets: Vec<usize>,
    mut row_indices: Vec<usize>,
    mut values: Vec<T>
) -> Result<Self, SparseFormatError>
```
the `mut` qualifiers only apply to the *implementation*, they do not affect the user-facing API. Does that answer your question?

aarsenij:
Thank you, I will check it : )

aarsenij:
I looks like it works, but on each construction, it still has to traverse a matrix twice as `validate_and_optionally_sort_cs_data` is required to be called in `csc::try_from_csc_data` and in `SparsityPattern::try_from_offset_and_indices`.

Andlon:
Yeah, we might want to provide an unsafe `SparsityPattern::from_offset_and_indices_unchecked` method that only checks that the arrays have the correct length, but otherwise performs no additional checking.

The alternative would be to instead of "validating" several slices, we could directly construct `(SparsityPattern, Option<T>)` and return this. This would be a better option *right now*, but further down the road we'd like to also be able to borrow existing CSR/CSC/pattern data, so it would be useful to have functionality for validating borrowed data.

aarsenij:
According to my understanding of your idea, I added  `validate_and_optionally_sort_cs_data` and `from_offset_and_indices_unchecked` by my last commit.
`validate_and_optionally_sort_cs_data` was added to SparsityPattern in order to simplify error handling.

Maybe it would be possible to combine all the required functionality in single function, e.g. in `SparsityPattern::try_from_offset_and_indices` if we add optional arguments to it like here:
```
pub fn try_from_offsets_and_indices(
        major_dim: usize,
        minor_dim: usize,
        major_offsets: Vec<usize>,
        mut minor_indices: Vec<usize>,
	values: Option<&mut [T]>,
	sort: Option<bool>,
    ) -> Result<Self, SparsityPatternFormatError>
```

Or there are some objections I don't see?

nalgebra-sparse/src/pattern.rs
```diff
@@ -184,6 +186,131 @@ impl SparsityPattern {
...
+
+        Ok(Self {
+            major_offsets,
+            minor_indices,
+            minor_dim,
+        })
+    }
+
+    /// Validates cs data, sorts minor indices and values
+    pub fn validate_and_optionally_sort_cs_data<T>(
```
Andlon, line 222: I don't see any reason this should be in `SparsityPattern`. Also, we definitely don't want to make this a public method. At least my intention was to have this be an internal implementation detail - maybe rather put it in `cs.rs`?

nalgebra-sparse/src/pattern.rs
```diff
@@ -184,6 +186,131 @@ impl SparsityPattern {
...
+    /// Try to construct a sparsity pattern from the given dimensions, major offsets
+    /// and minor indices.
+    ///
+    /// Returns an error if the data does not conform to the requirements.
+    pub fn from_offset_and_indices_unchecked(
+        major_dim: usize,
+        minor_dim: usize,
+        major_offsets: Vec<usize>,
+        minor_indices: Vec<usize>,
+    ) -> Result<Self, SparsityPatternFormatError> {
```
Andlon, line 198: This should be an `unsafe` method, because if the variants are not upheld downstream code may lead to undefined behavior. Moreover, I don't think we should return a `Result` here: simply panic (by way of assertions) if the most basic invariants are not upheld.

nalgebra-sparse/src/pattern.rs
```diff
@@ -184,6 +186,131 @@ impl SparsityPattern {
...
+            for lane_idx in 0..major_dim {
+                let range_start = major_offsets[lane_idx];
+                let range_end = major_offsets[lane_idx + 1];
+
+                // Test that major offsets are monotonically increasing
+                if range_start > range_end {
+                    return Err(NonmonotonicOffsets);
+                }
+
+                let minor_indices2 = &minor_indices[range_start..range_end];
```
Andlon, line 263: Maybe rather name the variable `minor_idx_in_lane` to make it clear that this is a slice of the minor indices in the current lane?

Also, the previous implementation didn't check the `range_start` and `range_end` (see discussion in #1019). Maybe we can incorporate the fix in this PR: you'd need to check that `range_start`/`range_end` are actually in bounds for `minor_indices`. You could then also port the test from #1019 to explicitly test for this case. What do you think?

nalgebra-sparse/src/pattern.rs
```diff
@@ -184,6 +186,131 @@ impl SparsityPattern {
...
+        // Check that the first and last offsets conform to the specification
+        {
+            let first_offset_ok = *major_offsets.first().unwrap() == 0;
+            let last_offset_ok = *major_offsets.last().unwrap() == minor_indices.len();
+            if !first_offset_ok || !last_offset_ok {
+                return Err(InvalidOffsetFirstLast);
+            }
+        }
+
+        let mut minor_index_permutation: Vec<usize> = (0..minor_indices.len()).collect();
```
Andlon, line 248: *Always* allocating a vector here is very unfortunate. We need to make sure we don't have any heap allocations on the "happy path" (i.e. when the data is already in a valid format). One way to ensure this is to instead think of the permutation as a buffer, and then only permute *locally*. Roughly like this:

- Start with empty vector `Vec::new()`
- *If* we should sort and a lane is unsorted, then clear and populate the permutation buffer with `(0 .. lane_size)`
- Apply the permutation to the minor indices, and possibly the values if they are provided

This way the vector stays empty if there is no need to sort (and thus never needs to heap allocate). Makes sense?

nalgebra-sparse/src/convert/serial.rs
```diff
@@ -218,7 +218,7 @@ where
 /// Converts a [`CscMatrix`] to a [`CsrMatrix`].
 pub fn convert_csc_csr<T>(csc: &CscMatrix<T>) -> CsrMatrix<T>
 where
-    T: Scalar,
+    T: Scalar + Zero,
```
Andlon, line 221: What is the reason for the additional bound?

nalgebra-sparse/src/csc.rs
```diff
@@ -170,6 +170,46 @@ impl<T> CscMatrix<T> {
...
+
+        let pattern = SparsityPattern::try_from_offsets_and_indices(
+            num_cols,
+            num_rows,
+            col_offsets,
+            row_indices,
+        )
+        .map_err(pattern_format_error_to_csc_error)?;
+
+        return Self::try_from_pattern_and_values(pattern, values);
```
Andlon, line 210: Here we should just validate and then call an unsafe method that skips further checks to avoid redundant checking

Andlon (commented):
Thanks for this @aarsenij! We're on the right path here - I do have a number of comments however. Please see my comments!

Andlon:
> According to my understanding of your idea, I added `validate_and_optionally_sort_cs_data` and `from_offset_and_indices_unchecked` by my last commit. `validate_and_optionally_sort_cs_data` was added to `SparsityPattern` in order to simplify error handling.
> 
> Maybe it would be possible to combine all the required functionality in single function, e.g. in `SparsityPattern::try_from_offset_and_indices` if we add optional arguments to it like here:
> 
> ```
> pub fn try_from_offsets_and_indices(
>         major_dim: usize,
>         minor_dim: usize,
>         major_offsets: Vec<usize>,
>         mut minor_indices: Vec<usize>,
> 	values: Option<&mut [T]>,
> 	sort: Option<bool>,
>     ) -> Result<Self, SparsityPatternFormatError>
> ```
> 
> Or are there any objections I don't see?

Well, `SparsityPattern` has no conception of "values", so it doesn't really belong here in my opinion. Also, validation of the pattern/CSR/CSC data should probably be an internal implementation detail and not leak into the API. I think we should try to keep our current API as long as there are no compelling arguments for changing it.

nalgebra-sparse/src/pattern.rs
aarsenij: I think it makes sense and try to apply this idea.

nalgebra-sparse/src/pattern.rs
aarsenij: `minor_idx_in_lane` sounds better and I agree `range_start` and `range_end` should be checked.
 It looks like changes from https://github.com/dimforge/nalgebra/pull/1019 do not directly relate to the case of unsorted indices as we handle it in separate functions. I will check if I can add analogous tests for unsorted data constructor.

nalgebra-sparse/src/convert/serial.rs
aarsenij: It was required after I introduced optional sorting in `try_from_csc_data`, now it should not be necessary.

nalgebra-sparse/src/pattern.rs
aarsenij: I reduced allocations of permutations in `validate_and_optionally_sort_cs_data`. My implementation might deviate a bit from your idea description but I hope it does roughly the same thing. :)

Further, I think the expectation on `validate_and_optionally_sort_cs_data` could be more clearer.
At the moment, if sorting is expected then values should be provided as well and an error is thrown otherwise.
Or should this behavior be changed?
Maybe we could even get rid of the boolean `sort` argument and let the `validate_and_optionally_sort_cs_data` sort data always when given indices are unsorted.
If we use `validate_and_optionally_sort_cs_data` without expecting it to sort data then its functionality seems to be redundant. One could use `try_from_csr_data`/`try_from_csc_data` instead.

nalgebra-sparse/src/csc.rs
aarsenij: Should be fixed now.

Andlon:
@aarsenij: I don't quite have the overview here. Is this ready for review or are there some remaining items that you want to address first?

aarsenij:
> @aarsenij: I don't quite have the overview here. Is this ready for review or are there some remaining items that you want to address first?

@Andlon :
I think it is ready for a review, but it is possible that I missed something

nalgebra-sparse/src/cs.rs
```diff
@@ -543,3 +544,146 @@ pub fn convert_counts_to_offsets(counts: &mut [usize]) {
...
+                    "Number of values and minor indices must be the same",
+                ));
+            }
+            value_refs = values;
+        }
+        None => {
+            if sort && minor_indices.len() > 0 {
+                return Err(SparseFormatError::from_kind_and_msg(
+                    SparseFormatErrorKind::InvalidStructure,
+                    "No values provided for sorting.",
```
Andlon, line 575: It seems to me that it would make total sense to sort without `values` if we're constructing e.g. a `SparsityPattern` from unsorted indices. But I guess this is not available yet, right? So it's fine that we have this restriction for now!

I think it's also OK to panic in this case instead of returning an error, because this is an *internal error* - we're the only one who can call this method, and it's an error on our part if this path gets triggered. (So could just replace with an assert)

nalgebra-sparse/src/cs.rs
```diff
@@ -543,3 +544,146 @@ pub fn convert_counts_to_offsets(counts: &mut [usize]) {
...
+                ),
+            )?;
+
+            // We test for in-bounds, uniqueness and monotonicity at the same time
+            // to ensure that we only visit each minor index once
+            let mut iter = minor_idx_in_lane.iter();
+            let mut prev = None;
+            let mut nonmonotonic = false;
+
+            while let Some(next) = iter.next().copied() {
```
Andlon, line 637: Can maybe just replace this with `for &minor_idx in &minor_idx_in_lane {` and remove the `iter`?

nalgebra-sparse/src/cs.rs
```diff
@@ -543,3 +544,146 @@ pub fn convert_counts_to_offsets(counts: &mut [usize]) {
...
+                    }
+                    if prev > next {
+                        nonmonotonic = true;
+                    }
+                }
+                prev = Some(next);
+            }
+
+            // sort if indices are nonmonotonic and sorting is expected
+            if nonmonotonic && sort {
```
Andlon, line 660: What if the indices are `nonmonotonic` but we're *not* asked to sort? In this case we need to return failure, right?

nalgebra-sparse/src/cs.rs
```diff
@@ -543,3 +544,146 @@ pub fn convert_counts_to_offsets(counts: &mut [usize]) {
...
+                minor_index_permutation.resize(range_end - range_start, 0);
+                for (index, value) in (range_start..range_end).enumerate() {
+                    minor_index_permutation[index] = value;
+                }
```
Andlon, line 664: Didn't we have some functions for applying permutations already? I think this whole part would be much easier to understand if we put the permutation stuff into separate functions. For example (not in any way tested):

```rust
// Set up *all* buffers up front (i.e. at the beginning of function, e.g. `Vec::new()`
let mut minor_idx_buffer = Vec::new();
let mut values_buffer = Vec::new();

// ...

compute_sort_permutation(&mut minor_index_permutation, &minor_idx_in_lane);
minor_idx_buffer.resize(minor_idx_in_lane.len());
permute(&minor_index_permutation, &minor_idx_in_lane, &mut minor_idx_buffer);
minor_idx_in_lane.copy_from_slice(&minor_idx_buffer);

if Some(values) = values {
   permute(&minor_index_permutation, &values, &mut values_buffer);
   values.clone_from_slice(&values_buffer);
}
```

What do you think?

nalgebra-sparse/src/csc.rs
```diff
@@ -170,6 +171,48 @@ impl<T> CscMatrix<T> {
...
+            num_rows,
+            &col_offsets,
+            &mut row_indices,
+            Some(&mut values),
+            true,
+        );
+
+        match result {
+            Ok(()) => unsafe {
+                let pattern = SparsityPattern::from_offset_and_indices_unchecked(
```
Andlon, line 204: Could maybe move the `unsafe` down to only encapsulate the `unchecked` function

nalgebra-sparse/src/csc.rs
```diff
@@ -518,7 +561,7 @@ impl<T> CscMatrix<T> {
     #[must_use]
     pub fn transpose(&self) -> CscMatrix<T>
     where
-        T: Scalar,
+        T: Scalar + Zero,
```
Andlon, line 564: Is this necessary?

nalgebra-sparse/src/cs.rs
Andlon: Doesn't have to look exactly like this, but the most important part is to remove the complete allocation of a buffer on every lane, so that we can amortize allocations over all lanes.

nalgebra-sparse/src/csr.rs
```diff
@@ -184,62 +185,33 @@ impl<T> CsrMatrix<T> {
...
         num_cols: usize,
         row_offsets: Vec<usize>,
-        col_indices: Vec<usize>,
-        values: Vec<T>,
+        mut col_indices: Vec<usize>,
+        mut values: Vec<T>,
     ) -> Result<Self, SparseFormatError>
     where
-        T: Scalar,
+        T: Scalar + Zero,
```
Andlon, line 192: I guess the `Zero` bound comes from needing to possible construct a buffer with e.g. `Vec::<T>resize`, is this correct?

nalgebra-sparse/src/csr.rs
```diff
@@ -184,62 +185,33 @@ impl<T> CsrMatrix<T> {
...
-            sorted_values,
+            &row_offsets,
+            &mut col_indices,
+            Some(&mut values),
+            true,
         );
+
+        match result {
+            Ok(()) => unsafe {
+                let pattern = SparsityPattern::from_offset_and_indices_unchecked(
```
Andlon, line 205: same as up above, cna probably move the unsafe to only encapsulate the unchecked call.

nalgebra-sparse/src/pattern.rs
```diff
@@ -184,6 +184,32 @@ impl SparsityPattern {
         })
     }
 
+    /// Try to construct a sparsity pattern from the given dimensions, major offsets
+    /// and minor indices.
+    ///
+    /// Returns an error if the data does not conform to the requirements.
```
Andlon, line 190: Docs here are not correct: it does not do any checking except panicking if the major offsets have the wrong size. Ideally this should be documented in a `# Panics` section.

nalgebra-sparse/tests/unit_tests/test_data_examples.rs
```diff
@@ -1,17 +1,44 @@
...
+        );
+
+        return Self {
+            valid_cs_data,
+            valid_unsorted_cs_data,
+        };
+    }
+}
+
+/// Examples of *invalid* raw CSR data `(offsets, indices, values)`.
```
Andlon, line 27: Maybe change this to `CS data`, since it's also used for CSC now.

Andlon (changes requested):
Great work here @aarsenij! Sorry that it's taken me so much time to get back to you, been pretty busy.

I think this is in pretty great shape already. My main concern right now is that we're currently allocating new vectors for each lane if sorting is required. See my comment for some suggestions to improve this.

Other than that just minor stuff.

nalgebra-sparse/src/csr.rs
aarsenij: Yes, then the buffer needs to be filled with some values, in this case with zeros.

nalgebra-sparse/src/cs.rs
aarsenij: Moving some logic from here to separate functions seems to work fine, I think it's a good idea.
I created _utils.rs_ for this purpose again.

nalgebra-sparse/src/cs.rs
```diff
@@ -630,56 +631,63 @@ where
...
             if nonmonotonic && sort {
-                minor_index_permutation.resize(range_end - range_start, 0);
-                for (index, value) in (range_start..range_end).enumerate() {
-                    minor_index_permutation[index] = value;
+                let range_size = range_end - range_start;
+                minor_index_permutation.resize(range_size, 0);
+                compute_sort_permutation(&mut minor_index_permutation, &minor_idx_in_lane);
+                minor_idx_buffer.resize(range_size, 0);
+                for (index, &value) in minor_idx_in_lane.iter().enumerate() {
+                    minor_idx_buffer[index] = value;
```
Andlon, line 672: Couldn't this be simplified to
```rust
minor_idx_buffer.clear();
minor_idx_buffer.clone_from_slice(&minor_idx_in_lane);
```
?

nalgebra-sparse/src/cs.rs
Andlon: It doesn't look to me like this was addressed yet?

nalgebra-sparse/src/cs.rs
```diff
@@ -630,56 +631,63 @@ where
...
-                    let v: T = unsorted_values[offset - range_start].clone();
-                    value_refs[index + range_start] = v;
+                apply_permutation(
+                    &mut minor_indices[range_start..range_end],
+                    &minor_idx_buffer,
+                    &minor_index_permutation,
+                );
+
+                // sort values if they exist
+                if !value_refs.is_empty() {
```
Andlon, line 681: This is pretty confusing, as it conflates whether or not the set of values is empty with whether or not they are present. Perhaps use `if let`?

```rust
if let Some(values) = values {
    let mut values_in_lane = ...;
    values_buffer.clear();
    values_buffer.extend_from_slice(&values_in_lane);
    apply_permutation(...)
}
```
I think this way we can also lift the `T: Zero` requirement since we no longer need to fill the buffer with arbitrary values.

Andlon (changes requested):
Hi @aarsenij: finally managed to get around to looking at your latest changes. A nice improvement! Though I still have some comments, see below. One of the more important ones seem to be that I think that we still don't error out if `nonmonotonic && !sort`: this is crucial for correctness!

I also *think* that my suggestions should allow us to remove the `T: Zero` bounds that we've added so far. Can you look into this? I'd like to avoid this breaking change if possible.

nalgebra-sparse/src/cs.rs
```diff
@@ -569,12 +569,10 @@ where
...
-            if sort && minor_indices.len() > 0 {
-                return Err(SparseFormatError::from_kind_and_msg(
-                    SparseFormatErrorKind::InvalidStructure,
-                    "No values provided for sorting.",
-                ));
-            }
+            assert!(
+                sort && minor_indices.len() > 0,
+                "No values provided for sorting."
+            );
```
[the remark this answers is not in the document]
aarsenij: Probably I misunderstood your last comment on these code lines when you said that it would be OK to have this restriction so far (see https://github.com/dimforge/nalgebra/pull/1015#discussion_r766490039). So I just replaced error response with a panic. 
So should we ignore a case when values are absent?

nalgebra-sparse/src/cs.rs
aarsenij: It should be addressed after adding following lines of code: 
https://github.com/dimforge/nalgebra/pull/1015/commits/9ff5d1a97929d4a9494383eb6dc5590f5243491f#diff-ad9fa9a189df6658e5fee89093cd2efba6292df75401887d4c3e336a6fca2e4bR650

nalgebra-sparse/src/cs.rs
aarsenij: It does not work directly as `clear()` does not change the size of the buffer which might be required on each loop iteration here. But one can use `reserve()` additionally and push required values to the buffer. 
And it looks like with this approach we can get rid of using T::Zero indeed.

nalgebra-sparse/src/cs.rs
aarsenij: I think I already tried this but there is an issue as the sort is applied only to a range of values and can be run in every iteration of the loop. So I get following error message if I try it: "_value moved here, in previous iteration of loop_".
Nevertheless, we can still lift T: Zero, as mentioned in my comment above.

Maybe we can find a better approach of checking values here.
But then another question comes up. How should handling of absent values differ from handling an empty list of values?

nalgebra-sparse/src/cs.rs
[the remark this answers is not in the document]
Andlon: Right. Yes, panicking in this case is fine, but I'm still confused about the assert condition, as it seems to do the opposite of the previous error, doesn't it? Also, I don't understand why we test for `minor_indices.len() > 0`?

Isn't something like this what you really want?

```rust
None => {
   if sort {
      panic!("Internal error: Currently don't support sorting if no values are present.");
   }
}
```
Perhaps I'm missing something though!

nalgebra-sparse/src/cs.rs
Andlon: Perhaps I misunderstand you, but if you use `clone_from_slice` as in my example you do not have to manage the resizing explicitly.

nalgebra-sparse/src/cs.rs
Andlon: Ah, right, I missed that. Do the tests currently exercise this branch?

nalgebra-sparse/src/cs.rs
Andlon: Ah, right, this is a bit tricky. The reason is that you would consume the `Option` at every iteration. I think this should work:
```rust
if let Some(values) = values.as_mut() {
   
}
```

The difference is mainly in semantics: The logic may *technically* be identical (not 100% sure), but when reading the code it's not quite clear what's going on. Basically, "values not provided" and "values provided but empty" are two different cases, and it can get a bit confusing if we batch them together.

Andlon (commented):
@aarsenij: I tried to respond to the points you brought up!

nalgebra-sparse/src/cs.rs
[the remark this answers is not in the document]
aarsenij: If `minor_indices.len() = 0` there would be nothing to sort, I suppose. So there would be no need to panic.
And I think in the previous error condition the assertion was the same (see [link](https://github.com/dimforge/nalgebra/pull/1015/commits/9ff5d1a97929d4a9494383eb6dc5590f5243491f#diff-ad9fa9a189df6658e5fee89093cd2efba6292df75401887d4c3e336a6fca2e4bL572) ).
But I see what you mean now, I used assert!() in a wrong way. Thank you.

nalgebra-sparse/src/cs.rs
aarsenij: I just tried to use exactly these two lines again and I get errors like `'unit_tests::csr::csr_matrix_valid_data_unsorted_column_indices' panicked at 'source slice length (3) does not match destination slice length (0)` when I run the unit tests.
Maybe I did something wrong but if we take a look into [documentation](https://doc.rust-lang.org/std/primitive.slice.html#method.clone_from_slice) , it looks like both slices have to be of same length.

nalgebra-sparse/src/cs.rs
aarsenij: Indeed, this works if we use a mutable variable for the values option check!

nalgebra-sparse/src/cs.rs
Andlon: Sorry, I meant `extend_from_slice`!

nalgebra-sparse/src/cs.rs
```diff
@@ -679,14 +666,22 @@ where
                 );
 
                 // sort values if they exist
-                if !value_refs.is_empty() {
+                if let Some(values) = values_option.as_mut() {
+                    if values.len() == 0 {
+                        panic!(
+                            "Internal error: Sorting currently not supported if values are empty."
+                        );
+                    }
```
Andlon, line 674: I don't understand this panic. Certainly we need to be able to handle this case, as it can easily happen based on user input. Also, what do we want to prevent here? It seems to me that the code should do the correct thing even in the case where `values` is empty?

nalgebra-sparse/src/cs.rs
```diff
@@ -679,14 +666,22 @@ where
...
+                    if minor_indices.len() != values.len() {
+                        panic!("Number of values and minor indices must be the same.");
+                    }
```
Andlon, line 677: This is an error condition, not a panic condition. In fact we used to check it [here](https://github.com/dimforge/nalgebra/pull/1015/commits/9af579f6623032b98910e03e048e5c26b33bac97#diff-ad9fa9a189df6658e5fee89093cd2efba6292df75401887d4c3e336a6fca2e4bL563).
I'm a bit puzzled by this change. Why remove the previous check and replace it with this panic? I think the previous check is preferable then.

nalgebra-sparse/src/cs.rs
```diff
@@ -679,14 +666,22 @@ where
...
                     values_buffer.clear();
                     values_buffer.reserve(range_size);
                     for index in range_start..range_end {
-                        values_buffer.push(value_refs[index].clone());
+                        values_buffer.push(values[index].clone());
                     }
```
Andlon, line 682: I think this can be simplified to
```rust
values_buffer.clear();
values_buffer.extend_from_slice(&values[range_start .. range_end]);
```

nalgebra-sparse/src/cs.rs
```diff
@@ -651,10 +639,9 @@ where
...
                     if prev > minor_idx {
                         if !sort {
-                            return Err(SparseFormatError::from_kind_and_msg(
-                                SparseFormatErrorKind::InvalidStructure,
-                                "Minor indices are not monotonically increasing within each lane.",
-                            ));
+                            panic!(
+                                "Minor indices are not monotonically increasing within each lane while sorting is not expected." 
+                            );
```
Andlon, line 644: Why was this turned into a panic?

Andlon (changes requested):
Sorry for taking so much time in getting back to you with a review, I've been busy lately.

I'm quite confused by some of the recent changes. Several error conditions were turned into panics. That doesn't make much sense to me, since we're trying to validate user input. As usual, please see comments.

nalgebra-sparse/src/cs.rs
aarsenij: Sorry. I thought I replied to your [question](https://github.com/dimforge/nalgebra/pull/1015#discussion_r780163049) regarding testing this branch but can't find my comment.
I was going to add a test for this case and came to the problem that it is not reachable for tests at the moment as this function i being called by other methods in `csr.rs`/`csc.rs`. So decided to turn the error into panic as it looks like to be an _internal error_ so far. Do you think, we should still return an error here instead?

nalgebra-sparse/src/cs.rs
aarsenij: I think you are right, this case could also be tested.

nalgebra-sparse/src/cs.rs
aarsenij: This could be an error indeed, but I am not sure if we should let the method do the correct thing regardless of empty `values` as I thought that we decided to have this restriction for now (see [comment](https://github.com/dimforge/nalgebra/pull/1015#discussion_r766490039)) .

 Otherwise we should also remove panic for the case [when `values` is None](https://github.com/dimforge/nalgebra/pull/1015/commits/bd864b84334a5496bdfc72a219570c5fb9c76831#diff-ad9fa9a189df6658e5fee89093cd2efba6292df75401887d4c3e336a6fca2e4bR563) .
 But for this case you said that it would be OK to panic, so I am also a bit confused when it is OK to panic and when not.
 
 I suppose we can chose between turning both cases into errors and test them or letting current method do the correct thing regardless of whether `values` is provided/non-empty or not?

nalgebra-sparse/src/cs.rs
aarsenij: And sorry for breaking your link. I rebased current branch and made the link unavailable unintentionally but I have a screenshot of it.
image

aarsenij:
@Andlon thank you for reviewing.
I added some further changes and I also was a bit confused regarding when it is OK to use panic and when not, please take a look on  my [comment](https://github.com/dimforge/nalgebra/pull/1015#discussion_r793997816).

Andlon:
@aarsenij:

Heh, I'm also a little confused at this point. I'll try to summarize the panic/error decisions at a high-level, I hope this helps to clarify things:

- A `panic` should only be used when:
  - we have internal assertions/sanity checks that should never be triggered unless there's a bug. This just helps us find bugs in our cod
  - we are checking API invariants that the user/programmer *must* adhere to.
- In particular, panics should never be triggered from *valid* user code.
- In the case of CSR/CSC validation from the `try_from_*` methods, panics should never be triggered due to invalid *data*, as in this case we need to return an error.

For the *internal* validation method that you are writing, I think the only case in which we can panic is if there's a path that we know can not be reached by any user-facing method. In fact, maybe we should even use `unreachable` instead of `panic` in that case to help clarify this fact.

Does that help clarify the difference between panic and errors?

nalgebra-sparse/src/cs.rs
Andlon: In addition to turning this back into an error, I think for clarity of code, it would be helpful if the check here only marked the lane as nonmontonic and the actual further processing of that logic happens together with the nonmonotonic logic.

See also separate comment about needing to check for duplicates *after* sorting.

That is, something like this (pseudo)

```rust
let mut monotonic = true;

for minor_idx in minor_idx_in_lane {
    // check out of bounds
    // ..
    if prev >= minor_idx {
        monotonic = false;
    }
}

if !monotonic {
    if sort {
       sort indices and possibly values
    } else {
       return error "minor indices not strictly monotonic"   
    }
    check for duplicates
}
```

nalgebra-sparse/src/cs.rs
```diff
@@ -543,3 +544,147 @@ pub fn convert_counts_to_offsets(counts: &mut [usize]) {
...
+                    &minor_idx_buffer,
+                    &minor_index_permutation,
+                );
+
+                // sort values if they exist
+                if let Some(values) = values_option.as_mut() {
+                    if values.len() == 0 {
+                        panic!(
+                            "Internal error: Sorting currently not supported if values are empty."
+                        );
```
Andlon, line 670: I don't immediately see any reason for this `panic`. Can it be directly removed?

Andlon (commented):
Especially after the force push I was having a hard time keeping everything in context, so I did a full pass from scratch. Please see comments: the main things remaining are changing some panics to errors and checking for duplicates *after* sorting (see comment on that, also includes some pseudo code).

aarsenij:
> unreach

Thank you, it think it makes distinguishing panic and error cases a bit clearer.

nalgebra-sparse/src/csc.rs
```diff
@@ -173,6 +174,50 @@ impl<T> CscMatrix<T> {
         Self::try_from_pattern_and_values(pattern, values)
     }
 
+    /// Try to construct a CSC matrix from raw CSC data with unsorted column indices.
+    ///
+    /// It is assumed that each row contains unique column indices that are in
+    /// bounds with respect to the number of columns in the matrix. If this is not the case,
+    /// an error is returned to indicate the failure.
```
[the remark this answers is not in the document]
aarsenij: Sorry for creating troubles after force push, I did not expect it then.

nalgebra-sparse/src/pattern.rs
aarsenij: I removed the # Errors section, as no errors are returned here.

nalgebra-sparse/src/cs.rs
aarsenij: Yes, I also think, it would be better to remove it.

nalgebra-sparse/src/cs.rs
```diff
@@ -543,3 +544,147 @@ pub fn convert_counts_to_offsets(counts: &mut [usize]) {
...
+                        SparseFormatErrorKind::IndexOutOfBounds,
+                        "A minor index is out of bounds.",
+                    ));
+                }
+
+                if let Some(prev) = prev {
+                    if prev == minor_idx {
+                        return Err(SparseFormatError::from_kind_and_msg(
+                            SparseFormatErrorKind::DuplicateEntry,
+                            "Input data contains duplicate entries.",
```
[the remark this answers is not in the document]
aarsenij: Done.

nalgebra-sparse/src/cs.rs
aarsenij: I turned it back into error, but I still think that this case is not reachable for user with current implementation.

Andlon (commented):
All right, we're basically there! I just made a couple of minor observations while going over the code a final time. They're not so important but I think it would improve both user experience and the code a little bit, so I think since we've made all this effort so far it would be nice to make those last couple of small improvements as well, if that's OK with you.

aarsenij:
I think the recently proposed changes make sense.

nalgebra-sparse/src/cs.rs
```diff
@@ -558,14 +558,21 @@ where
...
-        if sort {
+
+    if sort {
+        if let Some(values) = values_option.as_mut() {
+            if minor_indices.len() != values.len() {
+                return Err(SparseFormatError::from_kind_and_msg(
+                    SparseFormatErrorKind::InvalidStructure,
+                    "Number of values and minor indices must be the same.",
+                ));
+            }
```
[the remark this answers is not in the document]
aarsenij: I think, last commit did not really restrict the conditions for comparing the lengths of `values` and `minor_indices` but even extended them a bit. Previously, the check was only done if `sort == true` and `monotonic == false` (see previous [parent if-block](https://github.com/dimforge/nalgebra/blob/d7e7dcd9b5c09f194eff9ac6ae717fcf4400d457/nalgebra-sparse/src/cs.rs#L650)).
The problem is not clear to me. This check is only required if sorting is expected in current method. Otherwise it does not matter whether the lengths are same or not as no sparse matrix creation happens in this method.
Maybe something could break in another method, where `values` are needed and no corresponding check is performed but probably not here. 
Of course if someone modifies current method and uses `values` without extending the check conditions, it might lead to undesired behavior. But then a corresponding unit-test should be added as for every change anywhere else.

However, I do not have an overview on the whole package and the planned changes.
Do you think this check should be done independently from the value of `sort`?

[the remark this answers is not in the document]
Andlon: You're right, I didn't realize that the previous check was inside of several blocks (part of the motivation for making it explicit up front, by the way!).

Well, my understanding is that at *present*, the problematic code path that I outlined cannot be entered. However, I think the logic is a bit confusing. For example, it's **crucial** that if `values` are not `None`, then we must **never** invalidate the invariant that `values.len() == minor_indices.len()` otherwise we risk UB in other parts of the code that may rely on this invariant. This is really independent if `sort == true` or not. So I'd really prefer to have this switched around, e.g.
```rust
if let Some(values) = values_option.as_mut() {
    if minor_indices.len() != values.len() {
        return Err(SparseFormatError::from_kind_and_msg(
            SparseFormatErrorKind::InvalidStructure,
            "Number of values and minor indices must be the same.",
        ));
    }
} else if sort {
    unreachable!(
        "Internal error: Sorting currently not supported if no values are present."
    );
}
```
Not sure if I'm bringing my message across very well here: basically the first time I read the check I was confused because we start with checking `sort` - but this invariant in principle has nothing to do with sorting. I agree that the end result is at present pretty much the same, however.

Sorry for being so pedantic here, but I think it's very important to make the invariant checks here crystal clear for anyone interpreting the code later.

Andlon:
@aarsenij: thanks for sticking with me here! We've finally made it to the finishing line :) I apologize for being so pedantic, but interestingly this validation routine is one of the most important routines in the whole `nalgebra-sparse` library: if there's an issue in this routine it can lead to UB elsewhere. So it's important that we get it right. I'm pretty confident that things are correct at this point, in part due to your extensive work on setting up tests.

@sebcrozet: I assume the CI failure is unrelated? If so we're good to merge (might want to consider squashing for this one though? Not sure)

Andlon (approved):

sebcrozet:
Thank you both @aarsenij and @Andlon!

aarsenij:
> @aarsenij: thanks for sticking with me here! We've finally made it to the finishing line :) I apologize for being so pedantic, but interestingly this validation routine is one of the most important routines in the whole `nalgebra-sparse` library: if there's an issue in this routine it can lead to UB elsewhere. So it's important that we get it right. I'm pretty confident that things are correct at this point, in part due to your extensive work on setting up tests.
> 
> @sebcrozet: I assume the CI failure is unrelated? If so we're good to merge (might want to consider squashing for this one though? Not sure)

@Andlon I think it was useful to get some feedback, thank you for reviewing :)

Issue #1013, opened 2021-10-22
UnitDualQuaternion::sclerp() fails to interpolate equal rotations


Andlon:
Example:

```rust
let q1 = UnitDualQuaternion::from_rotation(UnitQuaternion::identity());
let q2 = UnitDualQuaternion::from_rotation(UnitQuaternion::identity());
q1.sclerp(&q2, 0.5);
```
panics with message:
```
DualQuaternion sclerp: ambiguous configuration.
```

Playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2d0788827c911f814132f41ce3e247d4

I realize that 180 degree rotation is ambiguous for interpolation, but 0 degrees ought to be unambiguous. I'd fix this myself but I'm not deep enough into the math to be confident about the fix (and I'm too involved with other things to get into it right now).

Maybe @tpdickso, @chinedufn or @ChristopherRabotin from #487 are able to chime in here.

sebcrozet:
Closing this as the PR fixing it is merged.

PR #1016, opened 2021-10-28
Don't panic ScLERPing `UnitDualQuaternion` with equal rotation

tpdickso:
Solves #1013.

Previously, when screw-linearly interpolating two unit dual quaternions
that had an identical orientation, `try_sclerp` would return `None`, as
the operation would introduce a division-by-zero.

This PR splits out the cases where two unit dual quaternions have an
identical orientation from the cases where they have opposite
orientations. In the case where they have identical orientations, the
operation is well-defined, but the exponential parameterization could
not handle it without introducing NaNs. Therefore, the function detects
this case and simply defaults to linearly interpolating the
translational components and using one of the two inputs' rotation
components.

The case where the inputs have opposite rotations is now detected
separately using the dot product of the real (rotation) parts, which was
already being computed anyway.

Also introduces proptests for these specific scenarios, to avoid any
regression.

### DIFF
--- a/src/geometry/dual_quaternion.rs
+++ b/src/geometry/dual_quaternion.rs
@@ -57,10 +57,7 @@ impl<T: Scalar> PartialEq for DualQuaternion<T> {
 
 impl<T: Scalar + Zero> Default for DualQuaternion<T> {
     fn default() -> Self {
-        Self {
-            real: Quaternion::default(),
-            dual: Quaternion::default(),
-        }
+        Self { real: Quaternion::default(), dual: Quaternion::default() }
     }
 }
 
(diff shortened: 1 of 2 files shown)


Andlon:
@tpdickso: looks like you need to run `cargo fmt` for CI to be happy. Is there otherwise anything else remaining to be done for this PR?

tpdickso:
Nope, nothing other than that to be done. Probably won't do it tonight (have not been feeling well) but I can do it in a bit!

tpdickso:
@Andlon should be good now!

sebcrozet:
Thanks!

PR #1017, opened 2021-11-02
added reading matrix market data to sparse coomatrix

losanc:
Hi, 
I have just implemented some part of io code about sparse matrix. Right now, it can read a matrix market data into a `nalgebra_sparse::CooMatrix` with error handling. It can be enabled by `--features io`. And also it contains several correct examples in unit test, which can be run by `cargo test --features proptest-support --features compare --features io test_mm`, also it contains several wrong examples in doc test, in `nalgebra_sparse::io::MMErrorKind`, which can be run by `cargo test --doc --features io MMErrorKind`

### DIFF
--- a/nalgebra-sparse/Cargo.toml
+++ b/nalgebra-sparse/Cargo.toml
@@ -15,6 +15,7 @@ license = "Apache-2.0"
 [features]
 proptest-support = ["proptest", "nalgebra/proptest-support"]
 compare = [ "matrixcompare-core" ]
+io      = [ "pest", "pest_derive" ]
 
 # Enable to enable running some tests that take a lot of time to run
 slow-tests = []
@@ -24,6 +25,8 @@ nalgebra = { version="0.29", path = "../" }
 num-traits = { version = "0.2", default-features = false }
 proptest = { version = "1.0", optional = true }
 matrixcompare-core = { version = "0.1.0", optional = true }
+pest           = { version = "2", optional = true }
+pest_derive    = { version = "2", optional = true }
 
 [dev-dependencies]
 itertools = "0.10"
--- a/nalgebra-sparse/src/io/matrix_market.rs
+++ b/nalgebra-sparse/src/io/matrix_market.rs
@@ -1072,17 +1072,6 @@ fn parse_dense_shape(
     let mut inner = shape_inner.into_inner();
     let r = inner.next().unwrap().as_str().parse::<usize>().unwrap();
     let c = inner.next().unwrap().as_str().parse::<usize>().unwrap();
-    // shape information can't use 0 as dimension
-    if r * c == 0 {
-        return Err(MatrixMarketError::from_kind_and_message(
-            MatrixMarketErrorKind::ZeroError,
-            String::from(
-                "
-        Matrix can't have 0 as shape dimensions.
-        ",
-            ),
-        ));
-    }
 
     // check for square matrix, when it's not a general matrix
     if *storagescheme != StorageScheme::General && r != c {
--- a/nalgebra-sparse/src/io/matrix_market.rs
+++ b/nalgebra-sparse/src/io/matrix_market.rs
@@ -720,7 +720,7 @@ pub fn load_coo_from_matrix_market_str<T>(data: &str) -> Result<CooMatrix<T>, Ma
 where
     T: MatrixMarketScalar,
 {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let file = MMParser::parse(Rule::Document, data)?.next().unwrap();
 
     let mut rows: Vec<usize> = Vec::new();
@@ -935,7 +935,7 @@ fn check_lower_triangle(r: usize, c: usize) -> Result<(), MatrixMarketError> {
 #[inline]
 /// Parse a pest structure to a Typecode of the matrix.
 fn parse_header(inner: &mut Pairs<'_, Rule>) -> Typecode {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     Typecode {
         sparsity: inner
             .next()
@@ -968,7 +968,7 @@ fn parse_sparse_shape(
     inner: &mut Pairs<'_, Rule>,
     storagescheme: &StorageScheme,
 ) -> Result<(usize, usize, usize), MatrixMarketError> {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let shape_inner = inner.next().unwrap();
     if shape_inner.as_rule() != Rule::SparseShape {
         return Err(MatrixMarketError::from_kind_and_message(MatrixMarketErrorKind::ParsingError,format!("
@@ -1007,7 +1007,7 @@ fn parse_dense_shape(
     inner: &mut Pairs<'_, Rule>,
     storagescheme: &StorageScheme,
 ) -> Result<(usize, usize, usize), MatrixMarketError> {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let shape_inner = inner.next().unwrap();
     if shape_inner.as_rule() != Rule::DenseShape {
         return Err(MatrixMarketError::from_kind_and_message(MatrixMarketErrorKind::ParsingError,format!("
@@ -1066,7 +1066,7 @@ fn parse_sparse_real<T>(inner: &mut Pairs<'_, Rule>) -> Result<(usize, usize, T)
 where
     T: MatrixMarketScalar,
 {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let entry_inner = inner.next().unwrap();
     if entry_inner.as_rule() != Rule::SparseReal {
         return Err(MatrixMarketError::from_kind_and_message(MatrixMarketErrorKind::ParsingError,format!("
@@ -1085,7 +1085,7 @@ fn parse_sparse_int<T>(inner: &mut Pairs<'_, Rule>) -> Result<(usize, usize, T),
 where
     T: MatrixMarketScalar,
 {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let entry_inner = inner.next().unwrap();
     // Because integer numbers can also be parsed as float numbers, it will be checked again in `parse::<i128>()?`
     if entry_inner.as_rule() != Rule::SparseReal {
@@ -1113,7 +1113,7 @@ fn parse_sparse_pattern<T>(
 where
     T: MatrixMarketScalar,
 {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let entry_inner = inner.next().unwrap();
     if entry_inner.as_rule() != Rule::SparsePattern {
         return Err(MatrixMarketError::from_kind_and_message(
@@ -1138,7 +1138,7 @@ fn parse_sparse_complex<T>(
 where
     T: MatrixMarketScalar,
 {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let entry_inner = inner.next().unwrap();
     if entry_inner.as_rule() != Rule::SparseComplex {
         return Err(MatrixMarketError::from_kind_and_message(MatrixMarketErrorKind::ParsingError,format!("
@@ -1158,7 +1158,7 @@ fn parse_dense_real<T>(inner: &mut Pairs<'_, Rule>) -> Result<T, MatrixMarketErr
 where
     T: MatrixMarketScalar,
 {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let entry_inner = inner.next().unwrap();
     if entry_inner.as_rule() != Rule::DenseReal {
         return Err(MatrixMarketError::from_kind_and_message(
@@ -1181,7 +1181,7 @@ fn parse_dense_int<T>(inner: &mut Pairs<'_, Rule>) -> Result<T, MatrixMarketErro
 where
     T: MatrixMarketScalar,
 {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let entry_inner = inner.next().unwrap();
     // Because integer numbers can also be parsed as float numbers, it will be checked again in `parse::<i128>()?`
     if entry_inner.as_rule() != Rule::DenseReal {
@@ -1206,7 +1206,7 @@ fn parse_dense_complex<T>(inner: &mut Pairs<'_, Rule>) -> Result<T, MatrixMarket
 where
     T: MatrixMarketScalar,
 {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let entry_inner = inner.next().unwrap();
     // Note: theoretically, 2 positive integers could also become the complex number,
     // but it would be parsed as SparsePattern, because SparsePattern has higher priority.
@@ -1236,7 +1236,7 @@ where
 fn parse_sparse_coordiante(
     inner: &mut Pairs<'_, Rule>,
 ) -> Result<(usize, usize), MatrixMarketError> {
-    // unwrap() in this function are guaranteed by pasing the data
+    // unwrap() in this function are guaranteed by parsing the data
     let r = inner.next().unwrap().as_str().parse::<usize>().unwrap();
     let c = inner.next().unwrap().as_str().parse::<usize>().unwrap();
     if r * c == 0 {
--- a/nalgebra-sparse/src/io/matrix_market.rs
+++ b/nalgebra-sparse/src/io/matrix_market.rs
@@ -1,48 +1,51 @@
-//! use mm(or MM) to represent martrix market
+//! Implementation of matrix market io code.
+//!
+//! See the [website](https://math.nist.gov/MatrixMarket/formats.html) or the [paper](https://www.researchgate.net/publication/2630533_The_Matrix_Market_Exchange_Formats_Initial_Design) for more details about matrix market.
 use crate::coo::CooMatrix;
 use crate::SparseFormatError;
 use crate::SparseFormatErrorKind;
+use nalgebra::base::Scalar;
 use nalgebra::Complex;
 use pest::iterators::Pairs;
 use pest::Parser;
 use std::cmp::PartialEq;
+use std::convert::Infallible;
+use std::convert::TryFrom;
 use std::fmt;
 use std::fmt::Formatter;
 use std::fs;
 use std::num::ParseIntError;
-use std::ops::Neg;
+use std::num::TryFromIntError;
 use std::path::Path;
 use std::str::FromStr;
 
 /// A description of the error that occurred during importing a matrix from a matrix market format data.
 #[derive(Debug)]
-pub struct MMError {
-    error_kind: MMErrorKind,
+pub struct MatrixMarketError {
+    error_kind: MatrixMarketErrorKind,
     message: String,
 }
 
 /// Errors produced by functions that expect well-formed matrix market format data.
-/// > _NOTE1:_ Since the matrix market design didn't mention if duplicate entries are allowed or not, so, it's allowed here.
-///
-/// > _NOTE2:_ Dense matrices are not supported here. For example, `%%matrixmarket matrix array real general` will give `MMErrorKind::ParsingError`.
+/// > _NOTE:_ Since the matrix market design didn't mention if multiple sparse entries with the same coordiantes are allowed or not, so, it's allowed here.
 #[non_exhaustive]
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub enum MMErrorKind {
+pub enum MatrixMarketErrorKind {
     /// Indicates that some word is not known to MM format
     ///
     /// Examples
     /// --------
     /// ```rust
-    /// # use nalgebra_sparse::io::load_coo_from_mm_str;
-    /// # use nalgebra_sparse::io::MMErrorKind;
+    /// # use nalgebra_sparse::io::load_coo_from_matrix_market_str;
+    /// # use nalgebra_sparse::io::MatrixMarketErrorKind;
     /// let str = r#"
     /// %%MatrixMarket whatever whatever whatever whatever
     /// 1 1 1
     /// 1 1 5
     /// "#;
-    /// let matrix_error = load_coo_from_mm_str::<f64>(str);
+    /// let matrix_error = load_coo_from_matrix_market_str::<f64>(str);
     /// assert_eq!(matrix_error.is_err(), true);
-    /// assert_eq!(*matrix_error.unwrap_err().kind(),MMErrorKind::ParsingError);
+    /// assert_eq!(matrix_error.unwrap_err().kind(),MatrixMarketErrorKind::ParsingError);
     /// ```
     ParsingError,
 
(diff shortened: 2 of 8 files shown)


nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+pub fn load_coo_from_mm_file<T, P: AsRef<Path>>(path: P) -> Result<CooMatrix<T>, MMError>
+where
+    T: MMType,
+{
+    let file = fs::read_to_string(path)?;
+    load_coo_from_mm_str(&file)
+}
+
+/// Parses a Matrix Market file described by the given string, and returns the corresponding as CooMatrix format.
+pub fn load_coo_from_mm_str<T>(data: &str) -> Result<CooMatrix<T>, MMError>
```
Andlon, line 714: Maybe `load_coo_from_matrix_market_str`?

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+                "Entry: row {} col {} should be put into lower triangle",
+                r, c
+            ),
+        ));
+    }
+    Ok(())
+}
+
+/// Parses a Matrix Market file at the given path, and returns the corresponding sparse matrix as CooMatrix format.
+pub fn load_coo_from_mm_file<T, P: AsRef<Path>>(path: P) -> Result<CooMatrix<T>, MMError>
```
Andlon, line 705: Maybe `load_coo_from_matrix_market_file`?

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+    fn from_kind_and_message(error_type: MMErrorKind, message: String) -> Self {
+        Self {
+            error_kind: error_type,
+            message,
+        }
+    }
+
+    /// The operation error kind.
+    #[must_use]
+    pub fn kind(&self) -> &MMErrorKind {
```
Andlon, line 222: Maybe return the type directly rather than a reference here, as it gives us a bit more freedom in the future

nalgebra-sparse/src/io/matrix_market.rs
Andlon: Maybe also add an `# Errors` section containing something like:

> # Errors
> See [`MatrixMarketErrorKind`] for a list of possible error conditions.

And the same to `_from_mm_str`?

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+    /// # use nalgebra_sparse::io::load_coo_from_mm_str;
+    /// # use nalgebra_sparse::io::MMErrorKind;
+    /// let str = r#"
+    /// %%matrixmarket matrix coordinate integer general
+    /// % it should be called by load_coo_from_mm_str::<i32>(str), or any other integer type;
+    /// 3 3 1
+    /// 2 2 2
+    /// 2 3 2
+    /// "#;
+    /// let matrix_error = load_coo_from_mm_str::<f64>(str);
```
Andlon, line 102: It seems to me that it wouldn't be unreasonable to load an integer matrix into a floating-point representation, provided we can fit all the integers in the floating-point type. Thoughts?

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+    /// When matrix is a Real matrix, it will convert a [f64] number to this type.
+    fn from_f64(f: f64) -> Result<Self, MMError>;
+    /// When matrix is a Complx matrix, it will convert a [Complex<f64>] number to this type.
+    fn from_c64(c: Complex<f64>) -> Result<Self, MMError>;
+    /// When matrix is a Pattern matrix, it will convert a unit type [unit] to this type.
+    fn from_pattern(p: ()) -> Result<Self, MMError>;
+    /// When matrix is a Skew-symmetric matrix, it will convert itself to its negative.
+    fn negative(self) -> Result<Self, MMError>;
+    /// When matrix is a Hermitian matrix, it will convert itself to its conjugate.
+    fn conjugate(self) -> Result<Self, MMError>;
```
Andlon, line 434: `from_pattern` is a little strange... I find myself wishing that the last three methods here were just internal implementation details, rather than exposed in the trait itself. But I think this is OK for now.

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+            fn conjugate(self) -> Result<Self, MMError> {
+                Err(MMError::from_kind_and_message(
+                    MMErrorKind::TypeUnmatched,
+                    format!("Int type has no conjugate"),
+                ))
+            }
+            #[inline]
+            fn negative(self) -> Result<Self, MMError>
+            where
+                Self: Neg<Output = Self>,
```
Andlon, line 475: since you're just doing this in a macro, you don't need the additional trait bound I think

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+            fn conjugate(self) -> Result<Self, MMError> {
+                Err(MMError::from_kind_and_message(
+                    MMErrorKind::TypeUnmatched,
+                    format!("Float type has no conjugate"),
+                ))
+            }
+            #[inline]
+            fn negative(self) -> Result<Self, MMError>
+            where
+                Self: Neg<Output = Self>,
```
Andlon, line 521: same as before: I don't think this trait bound is needed

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+                ))
+            }
+            #[inline]
+            fn conjugate(self) -> Result<Self, MMError> {
+                Ok(self.conj())
+            }
+            #[inline]
+            fn negative(self) -> Result<Self, MMError>
+            where
+                Self: Neg<Output = Self>,
```
Andlon, line 567: trait bound probably not necessary?

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+mm_complex_impl!(f64);
+
+mm_pattern_impl!(());
+
+#[derive(Parser)]
+#[grammar = "io/matrix_market.pest"]
+struct MMParser;
+
+/// Parsing a pest structure to a Typecode of the matrix.
+fn parsing_header(inner: &mut Pairs<'_, Rule>, header_type: &mut Typecode) -> Result<(), MMError> {
```
Andlon, line 641: should perhaps be `parse_header`, if I understand correctly?

Andlon: Also, why take `header_type` by `&mut` instead of just returning it from the function...?

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+    // once the data can be parsed, all unwrap() in this function here will not panic.
+    header_type.sp = inner
```
Andlon, line 643: can you use full names for this, like `sparsity` instead of `sp`, and `data_type` instead of `dt` and so on?

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+        .unwrap()
+        .as_str()
+        .to_ascii_lowercase()
+        .parse::<StorageScheme>()
+        .unwrap();
+    typecode_precheck(&header_type)
+}
+
+/// Parsing a pest structure to 3 int, which are number of rols, cols and non-zeros.
+fn parsing_shape(inner: &mut Pairs<'_, Rule>, shape: &mut (usize, usize, usize)) {
```
Andlon, line 668: maybe `parse_shape`? Also, why take `&mut` instead of returning...?

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+    //  Default typecode
+    let mut header_type = Typecode {
+        sp: Sparsity::Sparse,
+        dt: DataType::Real,
+        ss: StorageScheme::General,
+    };
+    let mut shape = (0, 0, 0);
+    let mut rows: Vec<usize> = Vec::new();
+    let mut cols: Vec<usize> = Vec::new();
+    let mut data: Vec<T> = Vec::new();
```
Andlon, line 729: perhaps we already know the exact expected amount of items from the header information? If so we could use `Vec::with_capacity` to prevent unnecessary reallocations

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+
+    let mut count = 0;
+    for line in file.into_inner() {
+        match line.as_rule() {
+            Rule::Header => {
```
Andlon, line 734: To me it's a little bit strange that we look for a header for each line in the whole file: wouldn't it be more natural to first look for header & shape, then parse only entries after that?

Andlon (commented):
Thank you very much for this contribution @losanc

Overall I think this is an amazing addition to the library, and it's already in excellent shape.

I have various comments and questions - please see the review comments.

One question: Is the parser robust to e.g. additional whitespace? For example, will it parse
```
          5   5      -24.5           
4 3 24.0
    10 3 1.0
```
correctly, effectively ignoring the (clumsy) additional whitespace? Would maybe be nice to have an additional test for this...?

nalgebra-sparse/src/io/matrix_market.rs
losanc: This is because I see that `nalgebra_sparse::ops::serial::OperationError.kind()` and `nalgebra_sparse::SparseFormatError.kind()` both return a reference. So this is why I returned it with a reference. But `std::io::Error` returns the `ErrorKind` not the reference. 

I can fix this, but I think it would be better also change the `nalgebra_sparse::ops::serial::OperationError.kind()` and `nalgebra_sparse::SparseFormatError.kind()`

nalgebra-sparse/src/io/matrix_market.rs
losanc: So here should be some kind of softer restrictions? 
for example, an integer matrix can also be read as `f64` and `Complex<f64>`, and real matrix can be read by `Complex<f64>` as well.
And only given this error, when trying to read a "big" matrix type with a "small" data type? For example, reading an `f64` matrix with `load_coo_from_mm_str::<i32>`

nalgebra-sparse/src/io/matrix_market.rs
losanc: This is because we don't know exact expected amount of items. If it's general matrix, then we know the size. But if it's symmetric, (suppose that line of shape gives r,c,n for rows, cols, and nnzs), then the matrix size it's between n and 2n. If there is no element on the diagonal, then it's 2n of nnzs. But if there are elements on diagonal, then size is unknown(but between n and 2n)

but I guess with_capacity could be also helpful, even if we don't know exact the size?

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+#[derive(Debug)]
+pub struct MMError {
+    error_kind: MMErrorKind,
+    message: String,
+}
+
+/// Errors produced by functions that expect well-formed matrix market format data.
+/// > _NOTE1:_ Since the matrix market design didn't mention if duplicate entries are allowed or not, so, it's allowed here.
+///
+/// > _NOTE2:_ Dense matrices are not supported here. For example, `%%matrixmarket matrix array real general` will give `MMErrorKind::ParsingError`.
```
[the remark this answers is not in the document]
losanc: It's just because I think reading dense matrix into a `nalgebra::Matrix` could be better? but if want to be read as a dense matrix, it can also be done.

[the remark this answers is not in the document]
Andlon: That's true - but it might be that the user doesn't know up front that the matrix market matrix is in fact stored as a dense matrix. For example, imagine that you have a large collection of MM files and you want to test some algorithm on it, then maybe some of those files have dense matrices. It would be nice if it "just works" in basically all cases, I think!

nalgebra-sparse/src/io/matrix_market.rs
Andlon: Ah, good point. I'm not sure why I returned the `kind` by reference there...

Let's maybe leave changing `OperationError` and `SparseFormatError` for a separate PR, since this is a breaking change (whereas the matrix market stuff is not).

nalgebra-sparse/src/io/matrix_market.rs
Andlon: Yeah, I think that's a good idea. Hmm, maybe we should actually allow even reading floating-point matrices as `i32` *if* all numbers actually fit. For example, a matrix might have all integers - then perhaps we should actually allow this?

nalgebra-sparse/src/io/matrix_market.rs
Andlon: Ah, good point. Yeah I don't know the best solution here. Maybe just leave it as it is for now, it's not likely to be a big performance bottleneck for anyone anyway.

nalgebra-sparse/src/io/matrix_market.rs
losanc: I don't think there is a easy way to allow reading floating-point matrices as `i32`. 

First, there is no simple way to detect all float numbers are all actually int numbers, unless read the whole data once, then read it again to construct the matrix. This can lead to a huge waste. 
Second another way to do this, using `as` in rust, though all float numbers can be converted into int numbers. But it will truncate the floating part, for example 3.12 -> 3, and I don't think this a good idea either.
And it will be more complicated, if even allow reading complex matrix as `i32` (if they are has 0 as imaginary part, and real part is also a int number), for the same idea.

Andlon: Ah, I thought e.g. `i32: TryFrom<f64>` existed with exactly the behavior I suggested. Apparently `TryFrom<float> for int` generally is not available, which makes the whole thing pretty difficult.

Unfortunately it seems there also is no implementation for `TryFrom<i32> for f64` or similar - I somehow thought this existed. All right, let's stick with the current behavior. `as` casts are too error-prone, I think.

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,861 @@
...
+    /// When matrix is a Hermitian matrix, it will convert itself to its conjugate.
+    fn conjugate(self) -> Result<Self, MMError>;
+}
+/// Implement MMType for primitive integer types.
+macro_rules! mm_int_impl {
+    ($T:ty) => {
+        impl MMType for $T {
+            #[inline]
+            fn from_i128(i: i128) -> Result<Self, MMError> {
+                Ok(i as Self)
```
Andlon, line 442: Here we should probably use `TryFrom` instead of `as`! Otherwise we might get *completely* wrong results if `i` is too big to fit in the integer.

losanc: I tried to use `TryFrom`, but it seems it doesn't fit in here.

`tryfrom` seems only mostly works  for integer types now. for example, it has `i8::try_from(i: i128)`, but it doesn't work with float numbers. For example, there is no `f32::try_from(f: f64)`, also no `f64::try_from(i: i128)`

losanc: Ahhh. forget about it. I read your comment in another block. github automatically fold it, so I didn't see it. I will use `as` for now

Andlon: Ah, damn, you're right. I thought int<->float comparison was supported by `TryFrom`. Disappointed to see that it's (currently) not.

In either case, `as` is absolutely not acceptable: it can lead to very, very subtle and almost undetectable bugs. So then we should rather be strict for now and e.g. only accept integer input for integers and floating-point input for floating-point matrices...

losanc: Also, there is no `f32::try_from(f: f64)`, and I only find `f64 as f32 ` to convert `f64` number to `f32`. Though, there is `f64::from(f:f32)`, but it's not needed here.
I also tried something about `as`, so when doing `f64 as f32`, if it's larger than `f32::MAX`, it will return `inf`, otherwise, it will do the cast and lose some accuray.
So `try_from` for int numbers, and `as` for float numbers?

Andlon: That sounds like a good solution! In particular, [for floats we have](https://doc.rust-lang.org/reference/expressions/operator-expr.html#semantics):

```
- Casting from an f32 to an f64 is perfect and lossless
- Casting from an f64 to an f32 will produce the closest possible f32 **
    if necessary, rounding is according to roundTiesToEven mode ***
    on overflow, infinity (of the same sign as the input) is produced
```

losanc: Should loading int matrix as real be allowed?  [It says](https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast)
> 
   +  if necessary, rounding is according to roundTiesToEven mode ***
   + on overflow, infinity (of the same sign as the input) is produced
   +  note: with the current set of numeric types, overflow can only happen on u128 as f32 for values greater or equal to f32::MAX + (0.5 ULP)

Since I am using i128, not u128 inside, so overflow won't happen, but surely it will lose accuray

Andlon: Yeah that's a good question. It would in general be nice - on the other hand, it's unfortunate if the file says one number, and you end up with a different number in your floating point representation :thinking: 

I'd say for now, let's go with whatever is the easiest for you. Then we can document the behavior clearly and/or create an issue for looking more into this in the future if necessary.

losanc:
I haven't find a way to comment this, so I comment it here
```
One question: Is the parser robust to e.g. additional whitespace? For example, will it parse

          5   5      -24.5           
4 3 24.0
    10 3 1.0

correctly, effectively ignoring the (clumsy) additional whitespace? Would maybe be nice to have an additional test for this...?
```
yes, it can parse this, the first unit test example shows this

nalgebra-sparse/Cargo.toml
```diff
@@ -15,6 +15,7 @@ license = "Apache-2.0"
 [features]
 proptest-support = ["proptest", "nalgebra/proptest-support"]
 compare = [ "matrixcompare-core" ]
+io      = [ "pest", "pest_derive" ]
```
Andlon, line 18: Should add `io` to the set of features used on `docs.rs` at the bottom of this file

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,1293 @@
...
+#[derive(Debug)]
+pub struct MatrixMarketError {
+    error_kind: MatrixMarketErrorKind,
+    message: String,
+}
+
+/// Errors produced by functions that expect well-formed matrix market format data.
+/// > _NOTE:_ Since the matrix market design didn't mention if multiple sparse entries with the same coordiantes are allowed or not, so, it's allowed here.
+#[non_exhaustive]
+#[derive(Copy, Clone, Debug, PartialEq)]
```
Andlon, line 32: I'd perhaps like to avoid the `Copy` impl for now, because I'm not sure if we really want to keep it `Copy` in the future.

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,1293 @@
...
+        )
+    }
+}
+
+impl From<TryFromIntError> for MatrixMarketError {
+    fn from(err: TryFromIntError) -> Self {
+        Self::from_kind_and_message(
+            MatrixMarketErrorKind::TypeUnmatched,
+            format!(
+                "Please consider using a larger integery type. Error message: {}",
```
Andlon, line 339: Typo: "integery"

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,1293 @@
...
+    }
+
+    let mut inner = shape_inner.into_inner();
+
+    let r = inner.next().unwrap().as_str().parse::<usize>().unwrap();
+    let c = inner.next().unwrap().as_str().parse::<usize>().unwrap();
+    let nnz = inner.next().unwrap().as_str().parse::<usize>().unwrap();
+
+    // shape information can't use 0 as dimension
+    if r * c * nnz == 0 {
```
Andlon, line 986: Does this mean we cannot load an empty matrix?

Also, should check the numbers individually in order to avoid possible overflow.

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -0,0 +1,1293 @@
...
+            MatrixMarketErrorKind::ZeroError,
+            String::from("The data has to be one-indixed"),
+        ));
+    }
+    // The coordiantes in matrix market is one-based, but in CooMatrix is zero-based.
+    Ok((r - 1, c - 1))
+}
+
+/// Calculate the next coordiantes used for dense matrix
+fn next_dense_coordiante(
```
Andlon, line 1253: typo: "coordinate"

nalgebra-sparse/tests/unit_tests/mod.rs
```diff
@@ -3,6 +3,8 @@ mod convert_serial;
 mod coo;
 mod csc;
 mod csr;
+#[cfg(feature = "io")]
```
Andlon, line 6: Should make sure to run the IO tests on CI!

Andlon (commented):
Thanks for addressing my concerns! I had a final pass and noticed a few details.

I'd like to go over the docs and make them a little bit more consistent with the other docs of the library - I think it would be much faster if I just do this quickly myself rather than trying to explain what I have in mind. Would you mind if I do this and send the commits as PRs to your branch? That way we can maintain your commits when we finally merge to the `nalgebra` repo.

If we do it this way I can also fix up the last remaining details that I noticed - maybe need a clarification from you on the question of whether `nnz = 0` should really be illegal. I think it's important to support loading empty matrices, unless there is a good reason that this is not feasible.

losanc:
>I'd like to go over the docs and make them a little bit more consistent with the other docs of the library - I think it would be much faster if I just do this quickly myself rather than trying to explain what I have in mind. Would you mind if I do this and send the commits as PRs to your branch? That way we can maintain your commits when we finally merge to the nalgebra repo.

sure, I don't mind that.

> whether nnz = 0 should really be illegal.

actually I didn't think very much of it when I write the code, but I think it can be allowed to have `nnz=0`

nalgebra-sparse/tests/unit_tests/mod.rs
losanc: should modify something inside `workflows/nalgebra-ci-build.yml` ?

because I never tried about github CI before, so don't know about it

nalgebra-sparse/src/io/matrix_market.rs
losanc: it's actually because both `OperationErrorKind` and `SparseFormatErrorKind` implement the copy trait, so I kind of copied that

losanc: actually, I tried remove the `Copy` impl, it gives compiling error ` type could implement `Copy`; consider adding `impl Copy``, because of `#![deny(missing_copy_implementations)]`

nalgebra-sparse/tests/unit_tests/mod.rs
Andlon: I can handle this part!

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -739,8 +739,10 @@ struct MatrixMarketParser;
 /// --------
 /// ```
 /// use nalgebra_sparse::io::load_coo_from_matrix_market_file;
-/// // Use e.g. `f64` for floating-point matrices
-/// let matrix = load_coo_from_matrix_market_file::<i32>("path/to/matrix.mtx")?;
+/// // Use e.g. `i32` for integer matrices
+/// let matrix = load_coo_from_matrix_market_file::<i32,_>("path/to/matrix.mtx");
+/// // extract the real matrix here by
+/// // let matrix = matrix.unwrap();
```
Andlon, line 745: Right, apparently I hadn't run the doc tests!

I'd recommend changing this to (can remove all the comments):
```rust
let matrix: CooMatrix<i32> = load_coo_from_matrix_market_file("path/to/matrix.mtx").unwrap();
```

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -767,12 +769,15 @@ where
...
 /// 5 4 2
 /// 1 1 10
 /// 2 3 5
 /// "#;
-/// let matrix = load_coo_from_matrix_market_str::<i32>(str)?;
+/// // Use e.g. `i32` for integer matrices
+/// let matrix = load_coo_from_matrix_market_str::<i32>(str);
+/// // extract the real matrix here by
+/// // let matrix = matrix.unwrap();
 /// ```
```
Andlon, line 781: Same comment as for `_file`

Andlon (changes requested):
OK, one last change and we're done!

Just think some of the doc examples got a bit messier than necessary (my fault), so I suggested a change that I think is cleaner. What do you think?

Andlon:
Ah, right: that example won't actually work because the file does not exist :thinking: Maybe it's ok to add `should_panic` to the code block.

losanc:
> should_panic tells rustdoc that the code should compile correctly but panic during execution. If the code doesn't panic, the test will fail.

> The no_run attribute will compile your code but not run it. This is important for examples such as "Here's how to retrieve a web page," which you would want to ensure compiles, but might be run in a test environment that has no network access.

So, I guess no_run is more suitable?

Andlon:
Ah, yes, that's correct! I was not aware that this exists.

Andlon:
:clap: :clap: :clap:

We're there! I'm very happy with hos this PR has turned out now. It's a really awesome and useful addition to the library, and it paves the ground for setting up useful benchmarks with real-world matrices that we can use to improve performance of our algorithms.

@sebcrozet: Ready to merge from my perspective.

sebcrozet:
Fantastic! Thank you both @losanc and @Andlon.

Issue #897, opened 2021-05-20
Consider sorting eigenvalues and singular values


Andlon:
Currently, there are no guarantees about ordering of eigenvalues and singular vectors returned by nalgebra's decompositions. A common convention throughout mathematical literature is to assume that eigenvalues and singular values are ordered in descending order. Many other mathematical libraries also do this. By following the same convention we would ease porting efforts from other languages into `nalgebra` code. Recently, a user had this exact issue when porting code from `numpy` (see #893).

The simplest way to achieve this is to compute a permutation of the eigenvalues/singular values and then apply the permutation also to the columns/rows of the accompanying orthogonal matrices (if computed). While this should have negligible overhead for larger matrices since the decomposition itself is `O(n^3)`, it is not clear to me if perhaps this has some non-trivial cost for small matrices. In this case, we should benchmark and perhaps consider providing means to opt **out** of this behavior if users are concerned about performance.

Andlon:
Very closely related: #349.

nestordemeure:
> While this should have negligible overhead for larger matrices since the decomposition itself is `O(n^3)`

It might have a large memory overhead if nalgebra doesn't have an in-place permutation operation.

Andlon:
> > While this should have negligible overhead for larger matrices since the decomposition itself is `O(n^3)`
> 
> It might have a large memory overhead if nalgebra doesn't have an in-place permutation operation.

I don't think so - computing and storing a permutation would only require `O(n)` memory, compared to `O(n^2)` for storingthe eigenvalue/singular value decomposition. Or perhaps I misunderstand you!

What we do need to take care with, however, is allocations. In particular we would need to make sure that computing/storing/applying this permutation does not require heap allocation for small matrices. That is however already similar/analogous to how we e.g. deal with permutations in `LU`.

timjb:
This issue should be closed because #1081 has been merged, right?

Andlon:
@timjb: singular values are now sorted, but IIRC, eigenvalues are not yet sorted (unless I missed a PR for this!).

Issue #349, opened 2018-06-05
SVD sorted


nilgoyette:
I'm porting an application from Python to Rust and I was surprised to see that I obtain the same results when I use `numpy.linalg.svd` and `nalgebra::DMatrix::SVD` but in a random order. Turns out that the order is not random at all, it's simply a [convention](https://math.stackexchange.com/questions/844289/why-diagonal-matrix-svd-sorted-from-largest-to-smallest-value#844297) to sort the `u` and `v` matrices by the singular values `s`.

I was about to code it myself because I only have 3 possible indices (using a serie of if, else if, etc.) but it's kind of an hassle. Is there some hidden code in nalgebra that does this sorting? I may have missed it.

sebcrozet:
Well to be honest I don't remember sorting the eigenvalues. So if they are, that may be a side-effect of the implementation which I did not know about.

nilgoyette:
No, sorry, I may have been unclear. They are NOT sorted in nalgebra, even if it's kind of a convention to do so. I'm asking if there's a tool in nalgebra to get `u` and `v` sorted by `s`. If not, maybe we should add it. I'm probably not the only one who want the results as NumPy gives them.

sebcrozet:
Ah, ok, I misunderstood. Yes, we should definitely add a way to sort the singular values (and singular vectors accordingly)!

nilgoyette:
Perfect. I'll try to add this feature. Meanwhile, do you (or anyone) have a better method to sort an array AND obtain the ordered indices? Anything better than that:

    // Add an index to all values and sort the tuples, so we know how it was sorted
    let mut s: Vec<(_, _)> = svd.singular_values.into_iter().enumerate().map(
        |(idx, &v)| (v, idx)
    ).collect();
    s.sort_unstable_by(|a, b| b.partial_cmp(a).unwrap());

    // Unzip the values and order
    let order: Vec<_> = s.iter().map(|t| t.1).collect();
    let s: wathever = s.iter().map(|t| t.0).collect();
    // Reorder u and v using order

I know it's working but I have the feeling that it's allocating too much.

michael-p:
I can confirm that for many algorithms at least in computer vision (like homography or fundamental matrix estimation) this is absolutely critical because one is not really interested in the full SVD decomposition but just in, for instance, the column of V corresponding to the smallest singular value (minimum direction problem) and hence the singular values must be sorted.

michael-p:
[...]

PR #1018, opened 2021-11-03
Allow descending sorting of estimated SVD

Christopher22:
Calculating an SVD with descending singular values is commonly required for different applications in machine learning and computer vision (#349, #897). This pull request adds a function `sort_by_singular_values` and corresponding tests to enforce such an ordering for these use cases. As this sorting is completely optional, the users must "pay" for this specific order only if it is really required.

### DIFF
--- a/src/linalg/decomposition.rs
+++ b/src/linalg/decomposition.rs
@@ -74,7 +74,31 @@ impl<T: ComplexField, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
     }
 
     /// Computes the Singular Value Decomposition using implicit shift.
+    /// The singular values are guaranteed to be sorted in descending order.
+    /// If this order is not required consider using `svd_unordered`.
     pub fn svd(self, compute_u: bool, compute_v: bool) -> SVD<T, R, C>
+    where
+        R: DimMin<C>,
+        DimMinimum<R, C>: DimSub<U1>, // for Bidiagonal.
+        DefaultAllocator: Allocator<T, R, C>
+            + Allocator<T, C>
+            + Allocator<T, R>
+            + Allocator<T, DimDiff<DimMinimum<R, C>, U1>>
+            + Allocator<T, DimMinimum<R, C>, C>
+            + Allocator<T, R, DimMinimum<R, C>>
+            + Allocator<T, DimMinimum<R, C>>
+            + Allocator<T::RealField, DimMinimum<R, C>>
+            + Allocator<T::RealField, DimDiff<DimMinimum<R, C>, U1>>
+            + Allocator<(usize, usize), DimMinimum<R, C>>
+            + Allocator<(T::RealField, usize), DimMinimum<R, C>>,
+    {
+        SVD::new(self.into_owned(), compute_u, compute_v)
+    }
+
+    /// Computes the Singular Value Decomposition using implicit shift.
+    /// The singular values are not guaranteed to be sorted in any particular order.
+    /// If a descending order is required, consider using `svd` instead.
+    pub fn svd_unordered(self, compute_u: bool, compute_v: bool) -> SVD<T, R, C>
     where
         R: DimMin<C>,
         DimMinimum<R, C>: DimSub<U1>, // for Bidiagonal.
@@ -88,10 +112,12 @@ impl<T: ComplexField, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
             + Allocator<T::RealField, DimMinimum<R, C>>
             + Allocator<T::RealField, DimDiff<DimMinimum<R, C>, U1>>,
     {
-        SVD::new(self.into_owned(), compute_u, compute_v)
+        SVD::new_unordered(self.into_owned(), compute_u, compute_v)
     }
 
     /// Attempts to compute the Singular Value Decomposition of `matrix` using implicit shift.
+    /// The singular values are guaranteed to be sorted in descending order.
+    /// If this order is not required consider using `try_svd_unordered`.
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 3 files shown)


sebcrozet:
Thank you for this PR!
The CI is currently failing when:
- Compiling `nalgebra` without its default features
- Compiling `nalgebra` on a no-std environment.

This is caused by the use of `Vec` and `Dynamic`, which require an allocator that may not be available for some platforms.
Ideally, we should make this work without heap-allocating at all when the decomposed matrix have known dimensions. This can be done by:
- Having `let mut singular_values = self.singular_values.map_with_location(|r, _, e| (r, e));` instead of using a `Vec`.
- For the permutation sequence, we could use some type-level operations on dimensions to express the fact that we want the length of `singular_values` divided by two. But I think that it is acceptable to just use a buffer with the same size as the s ingular values: `crate::PermutationSequence::::identity_generic(singular_values.data.shape().0)`.
- Making these modifications will require some additional allocator bounds. I think we need at least `DefaultAllocator: Allocator<(usize, usize), DimMinimum<R, C>> + Allocator<(usize, T::RealField), DimMinimum<R, C>>`.

Andlon:
May I also suggest that sorted singular values becomes the default behavior, with an option to opt *out* for rare cases where this could somehow make a difference (almost never)? We've already had a fair share of users running into problem because they implicitly assume that the singular values are sorted - for example when porting code from Python.

Christopher22:
Wow, thanks for the detailed instructions, @sebcrozet! Extending the code was just as simple as building a piece of furniture by a well-known Swedish company... I'm curious if it works now. Is there any flag for testing this non-std-cases offline?

Christopher22:
@Andlon I understand your considerations completely. However, the algorithm works quite well by now and calling a function explicitly should be simple enough for an "opt in" imho. I added documentation both to the SVD struct itself and the svd and try_svd methods regarding the (not specified) ordering. 

If we would like to be completely compatible with sci-kit, we would need to add functions for a deterministic output, too. Otherwise, the signs may vary between different calls. However, that might be a different "problem".

Andlon:
> @Andlon I understand your considerations completely. However, the algorithm works quite well by now and calling a function explicitly should be simple enough for an "opt in" imho.

The point I'm trying to make is that many users don't even know that they implicitly depend on the ordering of the singular values - hence why I think the right thing is "opt out" rather than "opt in".

sebcrozet:
Then let’s avoid any confusion and just add a new boolean parameter to select if we want the singular values sorted or not:

```rust
struct Matrix {
    pub fn svd(self, sort_eigv: bool, compute_u: bool, compute_v: bool) -> SVD<T, R, C> { ... }
}

struct SVD {
    pub fn new(matrix: OMatrix<T, R, C>, sort_eigv: bool, compute_u: bool, compute_v: bool) -> Self { ... }
}
```

> Is there any flag for testing this non-std-cases offline?

@Christopher22 For no-std cases, you can try to compile it locally without the default features (`cargo check --no-default-features`). However, tests won’t run (because they require `std`).

Christopher22:
Nice, I will add this! :)

sebcrozet:
Ok, adding `SVD::new_unordered` and `matrix.svd_unordered` look like good options too. Let’s do that for better visibility!

Christopher22:
It is done! :) Actually, I spotted a bug in my tracking of circles within the permutations. The resulting algotihm is a tiny bit more complex but now tested by all the existing unit tests regarding SVD.

src/linalg/decomposition.rs
```diff
@@ -86,12 +87,34 @@ impl<T: ComplexField, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
...
-            + Allocator<T::RealField, DimDiff<DimMinimum<R, C>, U1>>,
+            + Allocator<T::RealField, DimDiff<DimMinimum<R, C>, U1>>
+            + Allocator<(usize, usize), DimMinimum<R, C>>
+            + Allocator<(T::RealField, usize), DimMinimum<R, C>>,
     {
         SVD::new(self.into_owned(), compute_u, compute_v)
     }
 
+    /// Computes the Singular Value Decomposition using implicit shift.
+    /// nalgebra does not guarantee any particular order of the singular values by default.
```
sebcrozet, line 98: Let’s make the comment similar to `svd`:
```suggestion
    /// The singular values are not guaranteed to be sorted in any particular order.
    /// If a descending order is required, consider using `svd` instead.
```

src/linalg/decomposition.rs
```diff
@@ -74,6 +74,7 @@ impl<T: ComplexField, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
     }
 
     /// Computes the Singular Value Decomposition using implicit shift.
+    /// The singular values are guaranteed to be sorted in descending order. If this order is not required consider using `svd_unordered`.
```
sebcrozet, line 77:
```suggestion
    /// The singular values are guaranteed to be sorted in descending order.
    /// If this order is not required consider using `svd_unordered`.
```

src/linalg/decomposition.rs
```diff
@@ -86,12 +87,34 @@ impl<T: ComplexField, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
...
+            + Allocator<T, R, DimMinimum<R, C>>
+            + Allocator<T, DimMinimum<R, C>>
+            + Allocator<T::RealField, DimMinimum<R, C>>
+            + Allocator<T::RealField, DimDiff<DimMinimum<R, C>, U1>>,
+    {
+        SVD::new_unordered(self.into_owned(), compute_u, compute_v)
+    }
+
     /// Attempts to compute the Singular Value Decomposition of `matrix` using implicit shift.
+    /// The singular values are guaranteed to be sorted in descending order. If this order is not required consider using `try_svd_unordered`.
```
sebcrozet, line 117:
```suggestion
    /// The singular values are guaranteed to be sorted in descending order.
    /// If this order is not required consider using `try_svd_unordered`.
```

src/linalg/decomposition.rs
```diff
@@ -119,10 +142,46 @@ impl<T: ComplexField, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
...
-            + Allocator<T::RealField, DimDiff<DimMinimum<R, C>, U1>>,
+            + Allocator<T::RealField, DimDiff<DimMinimum<R, C>, U1>>
+            + Allocator<(usize, usize), DimMinimum<R, C>>
+            + Allocator<(T::RealField, usize), DimMinimum<R, C>>,
     {
         SVD::try_new(self.into_owned(), compute_u, compute_v, eps, max_niter)
     }
+
+    /// Attempts to compute the Singular Value Decomposition of `matrix` using implicit shift.
+    /// nalgebra does not guarantee any particular order of the singular values by default.
```
sebcrozet, line 153:
```suggestion
    /// The singular values are not guaranteed to be sorted in any particular order.
    /// If a descending order is required, consider using `try_svd` instead.
```

src/linalg/svd.rs
```diff
@@ -79,8 +80,9 @@ where
         + Allocator<T::RealField, DimDiff<DimMinimum<R, C>, U1>>,
 {
     /// Computes the Singular Value Decomposition of `matrix` using implicit shift.
-    pub fn new(matrix: OMatrix<T, R, C>, compute_u: bool, compute_v: bool) -> Self {
-        Self::try_new(
+    /// nalgebra does not guarantee any particular order of the singular values by default.
```
sebcrozet, line 83:
```suggestion
    /// The singular values are not guaranteed to be sorted in any particular order.
    /// If a descending order is required, consider using `SVD::new` instead.
```

src/linalg/svd.rs
```diff
@@ -91,6 +93,7 @@ where
     }
 
     /// Attempts to compute the Singular Value Decomposition of `matrix` using implicit shift.
+    /// nalgebra does not guarantee any particular order of the singular values by default.
```
sebcrozet, line 96:
```suggestion
    /// The singular values are not guaranteed to be sorted in any particular order.
    /// If a descending order is required, consider using `SVD::try_new` instead.
```

src/linalg/svd.rs
```diff
@@ -612,6 +615,112 @@ where
...
+        + Allocator<T, DimMinimum<R, C>, C>
+        + Allocator<T, R, DimMinimum<R, C>>
+        + Allocator<T, DimMinimum<R, C>>
+        + Allocator<T::RealField, DimMinimum<R, C>>
+        + Allocator<T::RealField, DimDiff<DimMinimum<R, C>, U1>>
+        + Allocator<(usize, usize), DimMinimum<R, C>> // for sorted singular values
+        + Allocator<(T::RealField, usize), DimMinimum<R, C>>, // for sorted singular values
+{
+    /// Computes the Singular Value Decomposition of `matrix` using implicit shift.
+    /// The singular values are guaranteed to be sorted in descending order. If this order is not required consider using `new_unordered`.
```
sebcrozet, line 634:
```suggestion
    /// The singular values are guaranteed to be sorted in descending order.
    /// If this order is not required consider using `new_unordered`.
```

src/linalg/svd.rs
```diff
@@ -612,6 +615,112 @@ where
...
+    pub fn new(matrix: OMatrix<T, R, C>, compute_u: bool, compute_v: bool) -> Self {
+        let mut svd = Self::new_unordered(matrix, compute_u, compute_v);
+        svd.sort_by_singular_values();
+        svd
+    }
+
+    /// Attempts to compute the Singular Value Decomposition of `matrix` using implicit shift.
+    /// The singular values are guaranteed to be sorted in descending order. If this order is not required consider using `new_unordered`.
```
sebcrozet, line 642:
```suggestion
    /// The singular values are guaranteed to be sorted in descending order.
    /// If this order is not required consider using `new_unordered`.
```

src/linalg/svd.rs
```diff
@@ -628,15 +737,15 @@ where
     /// Computes the singular values of this matrix.
     #[must_use]
     pub fn singular_values(&self) -> OVector<T::RealField, DimMinimum<R, C>> {
-        SVD::new(self.clone_owned(), false, false).singular_values
+        SVD::new_unordered(self.clone_owned(), false, false).singular_values
```
sebcrozet, line 740: Since `fn svd(...)` returns an ordered result, it would be more coherent to have `fn singular_values(...)` return an ordered result too. And add `fn singular_values_unordered(...)`.

sebcrozet (changes requested):
Thank you for taking the time to modify it! It’s almost perfect now.

src/linalg/svd.rs
Christopher22: FIne with me! The (possible breaking) additional constraints on the allocator are not a problem, are they?

Christopher22: Oh yes, I forgot that only the "_unordered" function have the same constraints... Never mind :)

sebcrozet (approved):

sebcrozet:
Perfect, thank you @Christopher22! And thank you @Andlon for your input.

PR #1019, opened 2021-11-09
Implement Serde Serialize/Deserialize for nalgebra-sparse types

w1th0utnam3:
This PR adds Serialize/Deserialize implementations for the following `nalgebra-sparse` types:
- `SparsityPattern`
- `CooMatrix`
- `CscMatrix`
- `CsrMatrix`

Internally I uses helper structs with a derived Serialized/Deserialize impl and then `try_from_*_data` to perform the fallible reconstruction of the type. Serialization is zero-copy because it serializes slices obtained from the types. Deserialization relies on the fact that `[T]` and `Vec<T>` serialize identically.

I decided to use a custom serialization helper type for each matrix instead of a nested sparsity pattern or similar because in my opinion this is more in line with what most people would expect when you serialize to a human readable format (simpler structure, clearer field names).

I tried to inline the helper structs into the trait impls as they are specific to these impls, but this currently causes rustc to crash with `SIGSEGV: invalid memory reference` (version 1.56.1, it compiles on the 1.57 beta channel though). I wasn't able to reproduce this in a smaller repo.

While writing tests for the implementations I noticed that some invalid data can actually cause a panic while constructing the sparsity pattern with `SparsityPattern::try_from_offsets_and_indices` instead of producing an error. The corresponding test cases are currently commented out (in the `csr_deserialize_invalid` and `csc_deserialize_invalid` tests).

EDIT: Error messages during deserialization are now propagated using `serde::de::Error::custom`. All `nalgebra-sparse` errors related to this are now updated to be all lowercase without punctuation.

Open questions:
- [x] Fix the panic related to invalid data in `try_from_offsets_and_indices`?
- [X] Should this PR already implement more specific error messages? IIRC Serde only supports `&'static str` as custom errors, so this would require a manual match with a string representation of each `SparseFormatErrorKind`
- [ ] Should this PR also implement Serialize/Deserialize for `CscSymbolicCholesky` and `CscCholesky`? I'm not sure what invariants have to be upheld for the factorization.

### DIFF
--- a/nalgebra-sparse/src/csc.rs
+++ b/nalgebra-sparse/src/csc.rs
@@ -7,12 +7,13 @@ use crate::cs::{CsLane, CsLaneIter, CsLaneIterMut, CsLaneMut, CsMatrix};
 use crate::csr::CsrMatrix;
 use crate::pattern::{SparsityPattern, SparsityPatternFormatError, SparsityPatternIter};
 use crate::{SparseEntry, SparseEntryMut, SparseFormatError, SparseFormatErrorKind};
-use std::borrow::Cow;
 
 use nalgebra::Scalar;
 use num_traits::One;
 #[cfg(feature = "serde-serialize")]
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+#[cfg(feature = "serde-serialize")]
+use std::borrow::Cow;
 use std::slice::{Iter, IterMut};
 
 /// A CSC representation of a sparse matrix.
--- a/nalgebra-sparse/src/csc.rs
+++ b/nalgebra-sparse/src/csc.rs
@@ -156,19 +156,15 @@ impl<T> CscMatrix<T> {
     /// An error is returned if the data given does not conform to the CSC storage format.
     /// See the documentation for [CscMatrix](struct.CscMatrix.html) for more information.
     pub fn try_from_csc_data(
-        num_rows: usize,
-        num_cols: usize,
+        nrows: usize,
+        ncols: usize,
         col_offsets: Vec<usize>,
         row_indices: Vec<usize>,
         values: Vec<T>,
     ) -> Result<Self, SparseFormatError> {
-        let pattern = SparsityPattern::try_from_offsets_and_indices(
-            num_cols,
-            num_rows,
-            col_offsets,
-            row_indices,
-        )
-        .map_err(pattern_format_error_to_csc_error)?;
+        let pattern =
+            SparsityPattern::try_from_offsets_and_indices(ncols, nrows, col_offsets, row_indices)
+                .map_err(pattern_format_error_to_csc_error)?;
         Self::try_from_pattern_and_values(pattern, values)
     }
 
--- a/nalgebra-sparse/src/csc.rs
+++ b/nalgebra-sparse/src/csc.rs
@@ -12,8 +12,6 @@ use nalgebra::Scalar;
 use num_traits::One;
 #[cfg(feature = "serde-serialize")]
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
-#[cfg(feature = "serde-serialize")]
-use std::borrow::Cow;
 use std::slice::{Iter, IterMut};
 
 /// A CSC representation of a sparse matrix.
@@ -526,12 +524,12 @@ impl<T> CscMatrix<T> {
 
 #[cfg(feature = "serde-serialize")]
 #[derive(Serialize, Deserialize)]
-struct CscMatrixSerializationData<'a, T: Clone> {
+struct CscMatrixSerializationData<Indices, Values> {
     nrows: usize,
     ncols: usize,
-    col_offsets: Cow<'a, [usize]>,
-    row_indices: Cow<'a, [usize]>,
-    values: Cow<'a, [T]>,
+    col_offsets: Indices,
+    row_indices: Indices,
+    values: Values,
 }
 
 #[cfg(feature = "serde-serialize")]
@@ -543,12 +541,12 @@ where
     where
         S: Serializer,
     {
-        CscMatrixSerializationData {
+        CscMatrixSerializationData::<&[usize], &[T]> {
             nrows: self.nrows(),
             ncols: self.ncols(),
-            col_offsets: Cow::Borrowed(self.col_offsets()),
-            row_indices: Cow::Borrowed(self.row_indices()),
-            values: Cow::Borrowed(self.values()),
+            col_offsets: self.col_offsets(),
+            row_indices: self.row_indices(),
+            values: self.values(),
         }
         .serialize(serializer)
     }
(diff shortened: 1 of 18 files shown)


w1th0utnam3:
Ok, I just fixed the panic with `SparsityPattern::try_from_offsets_and_indices` by using `.get(...)` instead of indexing for retrieving the minor indices of a lane and introducing a new error kind `SparsityPatternFormatError::MajorIndexOutOfBounds`. 

Previously: https://github.com/dimforge/nalgebra/blob/640ab4b12d7222bbddcdd72a44ab4daa9250034a/nalgebra-sparse/src/pattern.rs#L151-L156

Now: https://github.com/w1th0utnam3/nalgebra/blob/c8a12d6719b139bdb359113be1b64732cadd843f/nalgebra-sparse/src/pattern.rs#L154-L161

Andlon:
Awesome! Please see the discussion in #1020.

The only thing I can see that I'd perhaps consider changing is to move all the serde-specific stuff into separate files.

However, the panic fix (while great) would collide with @aarsenij's work in #1015, so perhaps we could leave this out of this PR? I think in general I'd anyway prefer to land #1015 first because any merge conflicts here might be harder to resolve for @aarsenij if we merge this first.

As for serializing/deserializing Cholesky, I think it makes sense to make it serializable, but I'm not sure of the invariants here, as I only ported the implementation. The difficulty here is that we don't have any constructors for constructing a cholesky factorization from the factors itself, which would probably be a requisite for implementing `Deserialize` (so that we could rely on the constructor for any invariant checking).

aarsenij:
> Awesome! Please see the discussion in #1020.
> 
> The only thing I can see that I'd perhaps consider changing is to move all the serde-specific stuff into separate files.
> 
> However, the panic fix (while great) would collide with @aarsenij's work in #1015, so perhaps we could leave this out of this PR? I think in general I'd anyway prefer to land #1015 first because any merge conflicts here might be harder to resolve for @aarsenij if we merge this first.
> 
> As for serializing/deserializing Cholesky, I think it makes sense to make it serializable, but I'm not sure of the invariants here, as I only ported the implementation. The difficulty here is that we don't have any constructors for constructing a cholesky factorization from the factors itself, which would probably be a requisite for implementing `Deserialize` (so that we could rely on the constructor for any invariant checking).

I think merging current PR would not create many problems for #1015 as the case of unsorted indices is handled in separate functions which will probably be replaced/moved anyway. So from my point of view I could handle the conflicts and there is no need to wait. :)

w1th0utnam3:
Initially I used a separate serialization helper struct with slices of the row/col/value data and a deserialization helper struct with `Vec`s for these values to avoid unnecessary copies when serializing. I realized that `Cow<'a, [T]>` also fulfills this purpose as it can be used to construct a `Vec` without cloning its content in the owned case and it serializes to the same representation.

The only downside is that both the `Serialize` and `Deserialize` impls now have ` T: Clone` bounds because this is required by `ToOwned` which is necessary to use `Cow`. In my opinion types that don't implement `Clone` are not interesting for `nalgebra-sparse`.

To simplify the code I would therefore keep this change.

I also updated the CI configuration to actually run `serde-serialize` tests and to compile `nalgebra-sparse` with different feature sets.

nalgebra-sparse/src/csr.rs
```diff
@@ -596,6 +591,60 @@ impl<T> CsrMatrix<T> {
     }
 }
 
+#[cfg(feature = "serde-serialize")]
+mod serde_serialize {
+    use super::CsrMatrix;
+    use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+
+    #[derive(Serialize, Deserialize)]
+    struct CsrMatrixSerializationData<Indices, Values> {
```
Andlon, line 600: Would maybe be nice to have a comment that explains the rationale for using the intermediate type? For the sake of future maintainers it would be good to document this!

Andlon (changes requested):
Just did a quick review here - I have yet to do a final thorough one.

From what I've seen this PR looks really good. I think we should split fixing up the formatting of the error messages into a separate PR, however, because it will *definitely* conflict with #879, which changes some of the `try_from*` methods.

So from my perspective, it would be best if we extract these error message formatting changes into a separate PR, so that we can merge the serialization stuff separately. What do you think?

nalgebra-sparse/src/csc.rs
```diff
@@ -524,6 +520,75 @@ impl<T> CscMatrix<T> {
     }
 }
 
+#[cfg(feature = "serde-serialize")]
+mod serde_serialize {
+    use super::CscMatrix;
```
Andlon, line 525: Would be nice to have this stuff in a separate file maybe? The same for other types.

nalgebra-sparse/src/pattern.rs
```diff
@@ -153,7 +153,9 @@ impl SparsityPattern {
                     return Err(NonmonotonicOffsets);
                 }
 
-                let minor_indices = &minor_indices[range_start..range_end];
+                let minor_indices = minor_indices
+                    .get(range_start..range_end)
+                    .ok_or(MajorIndexOutOfBounds)?;
```
Andlon, line 158: Is this a blocker for the serialization stuff? Because this will definitely collide with #1015...

nalgebra-sparse/src/pattern.rs
```diff
@@ -298,7 +362,7 @@ impl From<SparsityPatternFormatError> for SparseFormatError {
             | NonmonotonicMinorIndices => {
                 SparseFormatError::from_kind_and_error(InvalidStructure, Box::from(err))
             }
-            MinorIndexOutOfBounds => {
+            MajorIndexOutOfBounds | MinorIndexOutOfBounds => {
```
Andlon, line 365: Not sure if this collides with #1015 as well?

Andlon (changes requested):
Looks great to me! See comments.

The PR still contains some fixes for the "major index out of bounds" stuff though, which will cause conflicts with #1015. Could you maybe move these into a separate PR so that we don't lose track of it? That way we can make sure it really gets resolved either by or after the changes in #1015.

w1th0utnam3:
Ok, rebased to current dev branch. Serialization submodules are moved to separate files. I reverted the fix for the major index out of bounds panic.

nalgebra-sparse/src/csc.rs
```diff
@@ -154,19 +157,15 @@ impl<T> CscMatrix<T> {
...
-            num_cols,
-            num_rows,
-            col_offsets,
-            row_indices,
-        )
-        .map_err(pattern_format_error_to_csc_error)?;
+        let pattern =
+            SparsityPattern::try_from_offsets_and_indices(ncols, nrows, col_offsets, row_indices)
+                .map_err(pattern_format_error_to_csc_error)?;
         Self::try_from_pattern_and_values(pattern, values)
```
Andlon, line 169: Can you please take this out of the PR too? It will also conflict with #1015.

nalgebra-sparse/src/csr.rs
```diff
@@ -155,19 +158,15 @@ impl<T> CsrMatrix<T> {
...
-            num_rows,
-            num_cols,
-            row_offsets,
-            col_indices,
-        )
-        .map_err(pattern_format_error_to_csr_error)?;
+        let pattern =
+            SparsityPattern::try_from_offsets_and_indices(nrows, ncols, row_offsets, col_indices)
+                .map_err(pattern_format_error_to_csr_error)?;
         Self::try_from_pattern_and_values(pattern, values)
```
Andlon, line 170: Same here, will conflict with #1015 I think.

Andlon (changes requested):
Looks great, just a couple of unrelated changes that will end up conflicting with #1015 so I'd prefer to take them out.

w1th0utnam3:
Rebased and reverted the changes you mentioned.

Andlon:
Looks good to me! Thanks so much for implementing this.

@sebcrozet: this one's ready to be merged!

sebcrozet:
Thank you both!

Issue #876, opened 2021-04-29
Serde integration for nalgebra-sparse


Andlon:
`nalgebra-sparse` currently does not implement `Serialize` or `Deserialize`. Serialization should ideally be implemented for all major storage types (COO, CSR, CSC). These implementations should be hidden behind a feature flag, like `nalgebra`.

Note that deserialization requires some care, and simply slapping `derive(Deserialize)` is not sufficient. The reason is that the sparse matrix formats have some invariants that must be verified in order for subsequent use to remain sound.

PR #1020, opened 2021-11-09
WIP: Implement Serde Serialize/Deserialize for nalgebra-sparse types (Again)

CattleProdigy:
I'm opening this in response to seeing  https://github.com/dimforge/nalgebra/pull/1019

This PR is still a work in progress and hadn't yet nailed down a testing strategy.

Compared to #1019, my approach focuses on simplicity of code (e.g. it derives as much as possible), likely at the cost of simplicity and readability of the serialized output (which isn't important for my needs but might be for others). I don't have a strong preference for my approach over the one in #1019, but I figured I'd share and see what the maintainers think.

To get around using a helper struct, I'm using trick I saw here: https://github.com/serde-rs/serde/issues/1220

@w1th0utnam3 feel free to look and comment.

### DIFF
--- a/nalgebra-sparse/Cargo.toml
+++ b/nalgebra-sparse/Cargo.toml
@@ -13,8 +13,12 @@ keywords = [ "linear", "algebra", "matrix", "vector", "math" ]
 license = "Apache-2.0"
 
 [features]
+default = ["serde-serialize"]
 proptest-support = ["proptest", "nalgebra/proptest-support"]
 compare = [ "matrixcompare-core" ]
+# Sparse uses std::collection::Vec, so no-std is a non-starter for now
+# serde-serialize-no-std = [ "serde" ]
+serde-serialize        = [ "serde", "serde/std", "serde_test"]
 
 # Enable to enable running some tests that take a lot of time to run
 slow-tests = []
@@ -24,11 +28,14 @@ nalgebra = { version="0.29", path = "../" }
 num-traits = { version = "0.2", default-features = false }
 proptest = { version = "1.0", optional = true }
 matrixcompare-core = { version = "0.1.0", optional = true }
+serde = { version = "1.0", default-features = false, features = [ "derive" ], optional = true }
+serde_test = { version = "1.0", optional = true }
 
 [dev-dependencies]
 itertools = "0.10"
 matrixcompare = { version = "0.3.0", features = [ "proptest-support" ] }
 nalgebra = { version="0.29", path = "../", features = ["compare"] }
+serde_json = "1.0"
 
 [package.metadata.docs.rs]
 # Enable certain features when building docs for docs.rs
--- a/nalgebra-sparse/src/coo.rs
+++ b/nalgebra-sparse/src/coo.rs
@@ -2,6 +2,9 @@
 
 use crate::SparseFormatError;
 
+#[cfg(feature = "serde-serialize")]
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
 /// A COO representation of a sparse matrix.
 ///
 /// A COO matrix stores entries in coordinate-form, that is triplets `(i, j, v)`, where `i` and `j`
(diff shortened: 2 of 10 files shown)


Andlon:
Thanks @CattleProdigy, this is great. Hopefully we can soon land a nice serde impl by combining your effort and @w1th0utnam3's work in #1019.

I just took a brief look, and it's quite elegant. I'm a little confused how it actually works, though, maybe you can help me understand? For example, here (see comment):

```rust
 impl<'de, T: Deserialize<'de>> Deserialize<'de> for CsMatrix<T> {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        // Why doesn't this lead to infinite recursion?
        // Doesn't this call the same method as we're currently implementing?
        let unchecked = Self::deserialize(deserializer)?;
```

Andlon:
OK, so having looked at the two PRs, I think it would make sense to go with @w1th0utnam3's general approach: Although I think the output representation is not all that important, we actually do sometimes dump data to e.g. JSON, in which case it would be useful to have a more natural representation of the data, as it's easier to read and consequently debug. I also feel a little bit queasy about about constructing an "invalid" CSR/CSC matrix (even temporarily) with the `remote` approach - but this is perhaps unfounded and ultimately not a decision factor.

However, it would be nice to re-use the effort you have taken in writing tests. Perhaps we could port over the test cases @w1th0utnam3 ? @CattleProdigy: if you want to retain the ownership to those tests (by way of entries in the commit log), we can add those tests in a separate PR after merging #1019. What do you think?

CattleProdigy:
I'm back from vacation.

> Thanks @CattleProdigy, this is great. Hopefully we can soon land a nice serde impl by combining your effort and @w1th0utnam3's work in #1019.
> 
> I just took a brief look, and it's quite elegant. I'm a little confused how it actually works, though, maybe you can help me understand? For example, here (see comment):
> 
> ```rust
>  impl<'de, T: Deserialize<'de>> Deserialize<'de> for CsMatrix<T> {
>     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
>     where
>         D: Deserializer<'de>,
>     {
>         // Why doesn't this lead to infinite recursion?
>         // Doesn't this call the same method as we're currently implementing?
>         let unchecked = Self::deserialize(deserializer)?;
> ```

I don't actually know the answer to this just yet, but since we're not going with this approach. It'll remain a mystery for now.



> OK, so having looked at the two PRs, I think it would make sense to go with @w1th0utnam3's general approach: Although I think the output representation is not all that important, we actually do sometimes dump data to e.g. JSON, in which case it would be useful to have a more natural representation of the data, as it's easier to read and consequently debug. I also feel a little bit queasy about about constructing an "invalid" CSR/CSC matrix (even temporarily) with the `remote` approach - but this is perhaps unfounded and ultimately not a decision factor.
> 
> However, it would be nice to re-use the effort you have taken in writing tests. Perhaps we could port over the test cases @w1th0utnam3 ? @CattleProdigy: if you want to retain the ownership to those tests (by way of entries in the commit log), we can add those tests in a separate PR after merging #1019. What do you think?

That's fine by me. I don't share your concern about invalid construction, but the `remote` approach is exotic enough that perhaps it's worth avoiding. It doesn't save that much code anyways. 

R/e tests: I'm curious what aspects of my testing we'd like to retain. I felt my tests were a bit half-assed because I was trying to get the PR up in time for discussion. I'll look over @w1th0utnam3 's branch and see where more testing might be helpful. We have a few options logistically:

1. Separate PR after #1019 
2. I open a PR with more tests against @w1th0utnam3 's branch. He merges that and then we merge #1019.
3. I push commits with tests to a branch of my own and @w1th0utnam3 cherry picks them. In that case, he'd be listed as committer and I'd be the commit author on those commits, which is fine by me.

I'd vote for (2) over (1) since it provides just one PR which closes out https://github.com/dimforge/nalgebra/issues/876 and provides a single point for people to look back on. I think (3) is a worse version of (2).

Andlon:
@CattleProdigy: I agree that option (2) makes sense. @w1th0utnam3: any thoughts?

w1th0utnam3:
Yeah, (2) sounds good to me 👍

CattleProdigy:
I haven't had time to hack on this further so I'm closing. I don't have anything to contribute to the other branch right now.

@w1th0utnam3 if there's something specific you need help with, just let me know.

Andlon:
@CattleProdigy: appreciate all your input so far! We'll go ahead and get the other branch merged, so that we all can use serialization with `nalgebra-sparse` :clap:

PR #1022, opened 2021-11-16
Add extern crate declarations for lapack{-src}

CattleProdigy:
Without these declarations, `nalgebra-lapack` does not have runtime
linkage requirements for these libraries, meaning that binaries and
libraries using `nalgebra-lapack` have to link `lapack`/`lapack-src`
explicitly which shouldn't be necessary.


sebcrozet:
Thanks!

CattleProdigy:
Thank you!

Issue #878, opened 2021-04-29
CSR/CSC: Allow working with borrowed data + generalize index type


Andlon:
This issue describes two in principle disjoint issues, but a design that address either needs to take care to accommodate both. They are:

- Currently CSC/CSR (and COO) matrices only work with *owned* data. However, in many cases a user might have access to compatible, borrowed CSC/CSR data. In this case, it should be possible to work on the borrowed data with our data structures.
- Currently all indices in CSR/CSC matrices are stored as `usize`. For the vast majority of applications, 32-bit indices suffice, which can be expected to have a significant impact on performance due to reduced memory bandwidth requirements. We should therefore generalize the index type.

There are nuances to both these issues, and in the end they must remain compatible. Our *primary* goal for the design of both these features, however, should be to try to reduce the added complexity of a larger and increasingly generic API surface. The library should remain easy to use, the documentation should be easy to browse and the API easy to use.

To this end, one possibility would be to add another *Storage* generic parameter to `CsrMatrix/CscMatrix` and make it default to owned storage, e.g.:

```rust
struct CsrMatrix<T, Index=usize, Storage=OwnedCsStorage<T, Index>> { ... }
```
The existing `CsMatrix` struct, which is currently only an implementation detail used to reduce repetition between the CSR and CSC implementations, might be repurposed to fit the role of a `OwnedCsStorage`.

There's similarly a host of issues related to storing the indices. For example, we would like to be able to use *signed* integers as indices, because other software might use signed integers for this purpose. However, we need to make sure that we can soundly convert back and forth between `usize` and the index type on demand. That is, we need to ensure that all indices stored in a *valid* CSR/CSC matrix are actually convertible to `usize` without overflow/underflow.

ThatGeoGuy:
I'm fairly interested in this functionality. At present there's no way to slice sparse matrix types, and having borrowed storage makes sense for such functionality to exist.

I'm not sure that a generalized index type is needed for that (it seems like a distinct issue). 

Do you have recommendations for where to start if I wanted to add a custom storage generic parameter and start implementing the ability to slice csc / csr matrices?

ThatGeoGuy:
As for generalizing across index types, is the [`Dim` trait](https://docs.rs/nalgebra/latest/nalgebra/base/dimension/trait.Dim.html) insufficient? I feel like this is pretty much everything one would need on that front. It doesn't guarantee conversion to `usize`, but it does at least let you do `try_to_usize`.

[...]

PR #1023, opened 2021-11-16
WIP: [SPARSE] Support borrowed data / different index types in CsMatrix

ThatGeoGuy:
Alright, this is a work-in-progress PR surrounding issue #878. There's quite a big change in this refactor, so I wanted to go through each part bit by bit. The commit message so far is long, but I'm going to paste it here as well since it is easier for others to read as well.

# Goals

First off - my goals. I think it's important to understand where I'm coming from on this, since I wasn't super verbose in #878. My company ([@TangramVision](https://github.com/Tangram-Vision)) is in the middle of using nalgebra for some non-linear optimization work surrounding camera calibration. We are generating a Hessian matrix as part of this, which is mostly sparse. The sparse operations work well enough, however, we are interested in improvements in two key areas:

1. Being able to slice sparse matrices, so that we can compute the least squares estimate using the [Schur's complement](https://en.wikipedia.org/wiki/Schur_complement) method for solving very large non-linear systems. This doesn't have to be a free slice, but the method has some properties that are ideal for very large problem sizes.
2. Improving the speed of sparse-generalized-matrix-multiply (spGEMM). As of right now, we are seeing the following flamegraph when generating our Hessian:

Flamegraph of hessian generation

Looking at this, we find that `spmm_cs_prealloc` is where we're spending the most time. Notably I'm seeing that the `find` call in there is where we're spending most of the compute time. Taking a look at the code, it's as follows:

```rust
for i in 0..c.pattern().major_dim() {
    let a_lane_i = a.get_lane(i).unwrap();
    let mut c_lane_i = c.get_lane_mut(i).unwrap();
    for c_ij in c_lane_i.values_mut() {
        *c_ij = beta.clone() * c_ij.clone();
    }
                                                                                          
    for (&k, a_ik) in a_lane_i.minor_indices().iter().zip(a_lane_i.values()) {
        let b_lane_k = b.get_lane(k).unwrap();
        let (mut c_lane_i_cols, mut c_lane_i_values) = c_lane_i.indices_and_values_mut();
        let alpha_aik = alpha.clone() * a_ik.clone();
        for (j, b_kj) in b_lane_k.minor_indices().iter().zip(b_lane_k.values()) {
            // Determine the location in C to append the value
            let (c_local_idx, _) = c_lane_i_cols
                .iter()
                .enumerate()
                .find(|(_, c_col)| *c_col == j)
                .ok_or_else(spmm_cs_unexpected_entry)?;
                                                                                          
            c_lane_i_values[c_local_idx] += alpha_aik.clone() * b_kj.clone();
            c_lane_i_cols = &c_lane_i_cols[c_local_idx..];
            c_lane_i_values = &mut c_lane_i_values[c_local_idx..];
        }
    }
}
                                                                                          
Ok(())
```

The specific code isn't too important, but effectively the three for-loops plus the `find` get this closer to O(n^4) in the worst case (NOTE: it's a bit hand-wavy I'll admit, since "n" here depends on the number of non-zeros in each matrix, AND `c_lane_i_cols` gets smaller each iteration, but I think the point stands that we are running into issues due to repeated linear calls to `find`). I figured that there's gotta be a better way here, but I think the core of the problem is that spGEMM isn't taking advantage of the fact that one matrix will be CSR, and the other CSC. Instead it tries to be generic across all problem types, and suffers as a result. 

Anyways, on both these fronts I think we can make improvements that make spGEMM faster, as well as solve #878. I'm coming at this from the perspective that I want to solve #878, but keep the context in mind that ultimately we want a representation for `CsXMatrix` types that can be optimized better for e.g. spGEMM above. 

# Other projects

I took the time to look around at other projects (both in Rust and elsewhere) that deal with sparse matrices. The one that seemed the closest to what we were looking for is the [`sprs` crate](https://docs.rs/sprs/0.11.0/sprs/). However, this crate heavily leans on `ndarray` and the associated structures, so its a little far removed from what is desired here. Mostly though, it's the same, but their spGEMM dispatches to [one of four possible implementations](https://github.com/vbarrielle/sprs/blob/master/src/sparse/smmp.rs#L195-L234) for CSR / CSC product, CSR / CSR product, etc. The implementations are a bit more complex than above thanks to the multi-threading workspace aspect that they bring in, but overall the final implementations appear to be slightly smarter than O(n^4) in the worst case. 

In any case, I took a lot of direction from the `sprs` crate here, since they already have a generic implementation that would solve #878. They have proven that it works too - the crate links to the C++ suitesparse library for LDL / Cholesky decomposition and wraps around those types as well. There were some things that seemed amiss (e.g. `Deref<Target = [T]>` on their csmat type instead of `Borrow<[T]>`), but overall I tried to keep it very similar while also following the conventions set forth by `nalgebra`. 

# Current work (at time of posting)

This combs through the CSR / CSC matrix formats, and unifies them into a single type `CsMatrix`. While previously this type was private, this refactor unifies them both into a single public generic type, that can support either CSR or CSC formats with the same code.

The generic type introduces a few things that make a lot of code redundant, or at least fairly verbose. Some of the key changes here:

- SparsityPattern was removed
- Iterators were slightly renamed / moved around (two new iterators were introduced)
- CsrMatrix / CscMatrix are now aliases for the generic type
- Almost all functionality that would take `&mut self` of a CsXMatrix has been removed.
- CsMatrix stores "offsets" as counts per lane, not accumulated offsets. This may actually be wrong and in need of a future fixup, since it means that calling `get_lane` will have different performance when doing `get_lane(0)` vs. `get_lane(1000)`, as an example. However, this is on the radar and is up for discussion, because it is also advantageous in some scenarios (e.g. trying to slice along the major axis without copying any data).

Let's go through each of these big changes one-by-one:

## SparsityPattern

SparsityPattern was removed primarily because it adds nesting to the generic type. Instead, we now take in `MajorOffsets` and `MinorIndices` as generic parameters of the CsMatrix type. This is useful because nesting the type in means:

1. There needs to be interfaces for said type to use it
2. The code for it needs to be supported
3. We have to understand the abstraction

In pretty much every case, `Borrow<[Offset]>` and `Borrow<[Index]>` (which gives us `&[Offset]` and `&[Index]` respectively) solves the first point. In the past these were just thin wrappers around vectors, but since they are now made generic, we can get by without needing SparsityPattern purely by using the slice interface.

2 is handled trivially (the Rust std library is what supports `Borrow` and std::slice types), and 3 is something that you need to have a cogent understanding of in order to get started with Rust (else face the borrow checker). So overall, this just presents a cleaner way to offer the same functionality while reducing the overall code footprint. It wasn't a terrible abstraction, but nesting these types further into the SparsityPattern struct makes a lot of the remaining code more burdensome to write (because then we need to guarantee the first issue, an interface, via some trait or generic impl).

## Iterators

Two new iterators were introduced:

1. `AllElementsIter`
2. `CsLaneIter`

The first one is fairly self-explanatory - it iterates through all "entries" (including zeros) in the matrix. This isn't what you want most of the time but is a direct analog to `get_entry`. Mostly, it's for the sake of completion (otherwise you have to iterate through `get_entry` which requires a lot of option unpacking).

The second one isn't "new" per se, since a type with that name existed prior to the refactor. However, this refactor introduces iterators as:

- `CsMatrixIter`: iterate through each lane, as if calling `get_lane` successively.
- `CsLaneIter`: iterate through all the (Index, Value) pairs in a given lane.

The `CsLane` type has been removed because it was redundant. These iterators allow us to implement `triplet_iter` in a much more concise fashion. The new design along with these iterators allow for us to also remove the `CsMatrix::filter` function, since it can be replaced by:

    mat.triplet_iter().filter(/* ... */);

Using idiomatic iterator code.

## CsrMatrix / CscMatrix

These two are now aliases / specializations of the CsMatrix type. This is done to reduce nesting of the types, and is what enables the `transpose` function to be generic to index / offset types as well, since that function only needs to care about `shape` and `CompressionKind`.

Not sure what else to say here, but the lack of nesting means we can get rid of the source files csc.rs and csr.rs, since they mostly just wrapped the CsMatrix type anyways.

## The removal of `&mut self`

This one could be controversial. However, mutability in `CsMatrix` isn't actually that useful. Hear me out, please. In any case where you have mutability, you can really only have mutability on the non-zero values in the matrix, not the entire matrix itself. This is fine if you're adding a scalar to all non-zero values of the matrix, but doesn't make sense in a lot of other contexts. Stuff like `cs_data_mut`, `get_entry_mut`, etc. gave you a limited window into mutability when mutable access to the underlying data isn't necessarily safe to have in the compressed format.

Instead, it makes more sense to provide such functionality by consuming the type (i.e. taking ownership OR using `self` instead of `&mut self`). In that scenario, if the compressed layout needs to be recomputed because something in the structure changes, then you as a consumer will have `self` and can do that. You cannot change the underlying structure if you have `&mut self`. I don't think this actually prevents anything in the library, but rather just makes some of the clones / copies of sparse types more explicit.

I think overall if there's a desire to maintain the ability to mutate _just_ the non-zeros in the matrix, then a better way to do this would be to make the CsMatrix type implement Copy, and to let that be transparent where needed. However, I actually don't think we lose much other than "mutable interfaces for the sake of a complete API," which isn't necessarily semantically what you want.

## Offsets as counts

I did this because it was conceptually simpler, and because it avoids an edge case in the new iterators (specifically, CsMatrixIter). The iterator code is quite easy to understand IMO, and should be about as fast as we can hope for. The one sad bit of this is that `get_lane` will necessarily be slower when getting lanes at the end of the matrix rather than lanes near the beginning.

However, in most cases I would say one should not be getting lanes using `get_lane`. We can avoid this in CsMatrixIter by caching the previous total offset and adding each lane offset as we go. This makes it optimal, unlike `get_lane`. Pretty much every current operation will want to iterate through each lane sequentially, so I think from a computation standpoint it's fine if `get_lane` is slightly sub-optimal, since the natural way to write code will be to call `CsMatrix::iter` instead of `for i in 0..CsMatrix::nmajor` { CsMatrix::get_lane(i); } or whatever.

Again, it's up for discussion, but it simplifies a vast majority of lane-by-lane access while keeping it cache friendly :)

# To be done

- [ ] Fix format conversions and make them generic as well (I removed `cs::transpose_cs`, and should probably move that definition back into `nalgebra_sparse::convert`, rather than in `cs.rs`)
- [ ] Fix proptest generators (I assume using `usize` for `Offset` and `Index` is good enough to test?)
- [ ] Add tests for all iterators
- [ ] Ensure all constructors check for every invariant we care about
- [ ] Port examples from old code back into current rustdocs, get more / new examples for new interfaces
- [ ] Fix / update / complete definitions for all other `CsMatrix` operations that are not currently generic

Eventually I want to get to the point where we can think about optimizing spGEMM in the crate. However, as it stands, I think #878 is somewhat in the way of making that happen, especially since using smaller index types could contribute to better cache behaviour (although I am skeptical of that being the entire story). 

# Conclusion

I am sorry for the long, rambling post. Much of the work is yet undone, but I wanted to post a proof-of-concept to get early feedback on some of these points. I would appreciate any advice or feedback the maintainers have to offer. I'm doing this largely to get some speed improvements for my employer at large problem sizes, but we want to be good members of the community and contribute this back, so the ultimate aim is to get this upstreamed :)

Anyways, please have a look when you can and provided feedback. cc @Andlon since we were already talking about this in #878, although I've definitely gone a different direction from what I described in that issue.


ThatGeoGuy:
FWIW: I'm going to continue on the path as laid out in "To be done" above, but discussion can continue in the meantime. Worst case is I have to do some extra work after the fact, or re-factor / squash some changes, which I'm not worried about provided we get to an acceptable solution.

Andlon:
There's a lot to unpack here!

I'd like to support your use case, but I'm not convinced that such a total redesign of the library is the right way to go about addressing your concerns.

It seems to me that speeding up spmm is largely orthogonal to the big redesign you're proposing here. In the end it only comes down to managing some indices and floating-point numbers, and slices/views are not necessary to implement this. There's definitely a *lot* of room for improvement in speeding up spmm - the current implementation is as basic as it gets. But I see this largely as a separate issue that probably does not require sweeping changes to the library API.

With that out of the way, it seems to me that the majority of this design is primarily intended to support sparse matrix slices. As I mentioned in #878, I'd really advice you to reconsider if you *really* need sparse slices. In the number of years that I have been working with various sparse libraries, I have *never* seen a legitimate use case for this, so I hope that you will allow me to remain unconvinced pending further evidence. Invariably, almost any sparse matrix operation that you might want to perform on the slice is likely to be so expensive that you're probably just as well off simply copying out the data into a smaller separately stored matrix. That's also why I asked in #878 that perhaps a "selection API" might solve your particular use case.

Regardless of whether slices carry their weight, your design rolls back a number of very concious design decisions made during the initial design of `nalgebra-sparse`. The design is not set in stone, and I'm always open for suggestions for improvements, but the high-level design of `nalgebra-sparse` is very deliberate. I'll try to address some points where you have made significant changes below.

## Removal of SparsityPattern

The existence of `SparsityPattern` is one of the reasons I started working on `nalgebra-sparse` rather than use `sprs` (which at the time was somewhat less developed than it is now). Very often it is useful to be able to work with sparsity patterns separately from the floating-point data. The `SparsityPattern` concisely encodes only the graph of the problem. For example, for the Finite Element Method (my main motivating case for `nalgebra-sparse`), we can separately construct the `SparsityPattern` associated with a mesh, and for example construct a graph coloring which is useful for parallelization. In addition, we can assemble the sparsity pattern separately from the matrix entries. Often the sparsity pattern stays constant across iterations of an iterative algorithm, whereas the entries change. This way we can split up the computation of the pattern from the computation of the non-zero values.

## Combining CsrMatrix and CscMatrix into a single type

The second reason that motivated the design of `nalgebra-sparse` instead of using `sprs` is **precisely** the fact that `sprs` has a single sparse matrix type that encodes at run-time whether the matrix is CSR or CSC. This has very significant implications for users. For example, for every function

```rust
fn foo(matrix: &CsMatrix<T>) {}
```
the user must consider if the matrix is CSR or CSC, and either:
- make the method work for both.
- error at run-time.

However, users rarely want to deal with the complexity of managing two closely related - but ultimately different - sparse matrix formats. Usually a user wants *either* CSR or CSC.

Your suggestion is slightly better in this sense, because you encode at compile-time which type of matrix it is. However, the user still needs to work with a "generalized" type, which adds mental burden for the user.

The design of `nalgebra-sparse` is motivated by precisely this observation: a user is likely to want to work with **either** a CSR matrix, **or** a CSC matrix. Therefore a user should not need to know anything about a CSC matrix in order to effectively use a CSR matrix, or the other way around.

## Removing most of &mut self functionality

> However, mutability in CsMatrix isn't actually that useful. Hear me out, please. In any case where you have mutability, you can really only have mutability on the non-zero values in the matrix, not the entire matrix itself. This is fine if you're adding a scalar to all non-zero values of the matrix, but doesn't make sense in a lot of other contexts.

Unfortunately, this observation is completely wrong. Once again, the design of `nalgebra-sparse` is very deliberate in this regard: if the sparsity pattern does not change, then we can update only the non-zero values of the matrix by accessing only the non-zero values mutably. Taking the Finite Element Method as a motivating use case again, this is precisely how we can re-use our matrix from the previous iteration and directly re-assemble the system without needing to re-allocate any storage.

## Offsets as count

This would make the CSR/CSC formats incompatible with other sparse matrix libraries. Beyond that, this completely eradicates all the complexity guarantees of CSR/CSC matrices, which is based on its `O(1)` access to rows/cols. Moreover, your suggestion to caching a previous total offset only applies to single-threaded scenarios. In [fenris](https://github.com/InteractiveComputerGraphics/fenris) we are assembling CSR system matrices in parallel - in this context, having `O(1)` access to rows is a necessity.

## Generic interface

I was aware of the design of `sprs` when I started work on `nalgebra-sparse`, and the choice not to go down the same route was a deliberate one.

Consder the docs for [nalgebra-sparse::CsrMatrix](https://docs.rs/nalgebra-sparse/0.5.0/nalgebra_sparse/csr/struct.CsrMatrix.html) and [sprs::CsMat](https://docs.rs/sprs/0.11.0/sprs/type.CsMat.html), which is the closest equivalent.

In the case of `CsrMatrix`, we directly get a description of a `CsrMatrix`, which is a very simple type in that it only has a single generic scalar type `T` (just like `Vec<T>`). By comparison, `CsMat` is a type alias for `CsMatI<N, usize>`, which is a type alias for `CsMatBase<N, I, Vec<Iptr>, Vec<I>, Vec<N>, IPtr>`. There's a *lot* of generic complexity going on here. One of my goals with `nalgebra-sparse` has been to create an API that is more accessible and less complex. Of course, the comparison here is not fair: `CsrMatrix` always represents owned data, whereas `CsMat` has very flexible storage. However, in extending `CsrMatrix`/`CscMatrix` to types that can also work with borrowed data, I'd like to try to find a design whose generic complexity is more similar to the current `CsrMatrix` than `sprs::CsMat`. I have some ideas for how this might look like, and it is my hope that we can get the same or similar degree of flexibility while preserving the lower complexity of our docs and API.

## Conclusion

I appreciate your interest in `nalgebra-sparse` and the work you've invested here is commendable. However, your proposed design directly works against many of the guiding principles that I followed in the design of `nalgebra-sparse`. I hope I've managed to convince you that there are good reasons for these.

I suggest that we focus more on the specific issues that you stated as your motivation:

1. Do you **really** need slices/views, or would a "selection API" be sufficient? If not, can you elaborate on why not?
2. We can definitely improve the performance of spMM. But I don't think it's necessary to mix e.g. CSR * CSC to achieve this. In fact, I deliberately avoided mixing matrix types to avoid the combinatorial explosion of possible combinations. We just need a better algorithm.

Andlon:
Oh, and for the record, I have some concrete thoughts on how the SPMM algorithm can be improved - we can discuss this when we've decided on the path forward.

It should also be fairly straightforward to parallelize it, which might help with performance as well. We should also look at `sprs` for inspiration and comparison - I haven't done so yet.

ThatGeoGuy:
> It seems to me that speeding up spmm is largely orthogonal to the big redesign you're proposing here. In the end it only comes down to managing some indices and floating-point numbers, and slices/views are not necessary to implement this. There's definitely a lot of room for improvement in speeding up spmm - the current implementation is as basic as it gets. But I see this largely as a separate issue that probably does not require sweeping changes to the library API.

Admittedly, spGEMM is a separate issue. However, I think the case for solving #878 stands on its own. This work is concerned with that, so I will probably focus on that here. The big part of the design that relates to spGEMM though is the fact that there are two different types for CSC / CSR at compile time though -- this was a justification for something you already pointed out. Namely, that `sprs` defers the CSC / CSR distinction to runtime, which is confusing and can lead to slow code.

I want to respond to each bit in turn as well:

# Sparsity Pattern

The main reason for getting rid of it here was nesting. To solve #878, there was a need to make the index / offset types generic. e.g. if you want to store everything using `u32` indices instead of `usize`, to save space in memory. Consider if we leave sparsity pattern in, we would then have to define the `CsMatrix` type as follows:

```rust
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct CsMatrix<T, Offset, MajorOffsets, MinorIndices, Data, CompressionKind, Index = Offset>
where
    Offset: Add<usize, Output = usize> + Clone + Into<usize> + Unsigned + Ord,
    Index: Clone + Into<usize> + Unsigned + Ord,
    MajorOffsets: Borrow<[Offset]>,
    MinorIndices: Borrow<[Index]>,
    Data: Borrow<[T]>,
    CompressionKind: Compression,
{
    shape: (usize, usize),

    pattern: SparsityPattern<Offset, MajorOffsets, MinorIndices, Index>,

    data: Data,

    _phantom: PhantomData<(T, CompressionKind)>,
}
```

However, then we need to define `SparsityPattern` as follows:

```rust
pub struct SparsityPattern<Offset, MajorOffsets, MinorIndices, Index = Offset>
where
    Offset: Add<usize, Output = usize> + Clone + Into<usize> + Unsigned + Ord,
    Index: Clone + Into<usize> + Unsigned + Ord,
    MajorOffsets: Borrow<[Offset]>,
    MinorIndices: Borrow<[Index]>,
{
    major_offsets: MajorOffsets,
    minor_indices: MinorIndices,
    // this is probably redundant with `shape` in the CsMatrix type but lets ignore that
    minor_dim: usize,
    _phantom: PhantomData<(Offset, Index)>,
}
```

In order to make it possible to use borrowed data (the indices and compressed pattern are part of the data you may wish to borrow), you would have to nest the generics all the way down. This doesn't actually add anything other than API complexity, because now we're adding a bunch of trait restrictions everywhere. 

I generally agree that you want to handle the pattern separate from the data itself, especially since there's instances where the pattern may be constant but the data might update. However, keeping the type hierarchy flat allows this to happen already, without the extra overhead of managing trait bounds across many nested types.

Consider the `transpose` function as I've written it -- it returns a `CsMatrix<T, Offset, &[Offset], &[Index], &[T], CompressionKind::Transpose, Index>`. The fact that I can borrow the pattern part (i.e. `&[Offset]` and `&[Index]`) is kind of the whole point of what we want to do with #878. 

A similar use case arises with slicing. Let's say I have a CSC matrix and want to slice along columns (just columns). As it stands, I can do this for free by doing:

```rust
// lower and upper are column indices
let offsets = &self.offsets.borrow()[lower..upper];

// naive way to get lower / upper offsets but bear with me
let offset_lower = self.offsets.borrow()[0..lower].fold(0, |k, x| x + k);
let offset_upper = self.offsets.borrow()[0..upper].fold(0, |k, x| x + k);

let indices = &self.indices.borrow()[offset_lower..offset_upper];
let data = &self.data.borrow()[offset_lower..offset_upper];

CsMatrix {
    shape: (self.nrows, upper - lower),
    offsets,
    indices,
    data,
    _phantom: PhantomData,
}
```

It's a bit of a naive way to do it, but slices along the major axis should be free. One can still do this with sparsity pattern, but that's ~12-15 lines of code right there, and I can't imagine making it much simpler (of course, separate from this is the topic of counts vs offsets which _can_ make this a bit easier, but I'll get to that eventually). 

Note that in the above case the `SparsityPattern` type wouldn't really add much other than extra complexity. I understand the motivation for keeping it separate from the data itself, but I'm not sure the nested structure adds much of anything if you're moving towards generic index types.

I'm not sure this goes against a guiding principle here, since any implementation of making indices generic will have to deal with this in some capacity. Note that if you do keep sparsity pattern, then you would need some API to get:

* a new sparsity pattern
* the indices to slice the data / values

at the same time. The type nesting doesn't prevent this in the existing design, it just makes it a lot harder to read and make changes because you have to constantly add generics all the way down. It's not easy to solve #878 in a unified way without the generics, so I'm not sure what to do about that.

# Combining CsrMatrix and CscMatrix into a single type

> Your suggestion is slightly better in this sense, because you encode at compile-time which type of matrix it is. However, the user still needs to work with a "generalized" type, which adds mental burden for the user.

I mean, quite frankly there was already a `CsMatrix` type that unified these at a lower level. As for mental burden, this is `nalgebra-sparse`, and a lot of the interop is between the sparse type and `nalgebra::Matrix`. Users already have to deal with the complexity that is `Matrix`, `OMatrix`, `SMatrix`, `DMatrix`, etc. I would argue that this isn't a significant jump in understanding the underlying API and code. There seems to be two competing concerns here:

1. Having a generic storage backend / generic index type for different matrices
2. Having a single type that is plain and easy to understand

With regards to the `CsrMatrix` and `CscMatrix` type aliases though, you can just program with one or the other. Today's code would just be translated such that `CscMatrix<T>` becomes `CscMatrix<T, usize>` and `CsrMatrix<T>` becomes `CsrMatrix<T, usize>`. Users could make the remaining template bits generic if they desired, but I suspect only library maintainers would want to do so. In such a case, the need for generic offset / index types as outlined in #878 means that to some degree we still have to do this? 

I could see the argument that you want to provide something more concrete so that users don't have to add the trait bounds to all their functions. However, that would mean that using CsX matrix types that aren't using explicitly owned (i.e. `Vec<_>`) data with `usize` indices would be second class citizens in the library. Consider if you keep the type as I described it, would you then define:

```rust
pub struct OwnedCscMatrix<T> {
    mat: CscMatrix<T, usize>,
}
```

Certainly this allows one to move away from the trait bounds on anything that takes an `OwnedCscMatrix`, but do we then have to define non-owned variants as well? What if there are multiple kinds of non-owned variants? How much complexity do we save vs. how much do we gain then? I think that while it's important to try and keep the library manageable in terms of cognitive load, you can't also have custom index / offset types and expect complexity to not shift. If `nalgebra-sparse` doesn't actually want to solve this issue, then surely that's okay, but I'm not sure how you avoid the generic here otherwise.

Last note to hammer this point: I agree that we need to keep the types separate at compile time, not runtime. The algorithm for spGEMM can be changed depending on if you have `CSR * CSR` or `CSC * CSC` or some other combination. You can take advantage of the lane ordering for cache-locality, and change the underlying algorithm by dispatching to a different function. Rust even lets us get away with this, since I could make an (admittedly ugly) impl of the `Mul` trait something like:

```rust
impl Mul<CsrMatrix> for CscMatrix { ... } 
``` 

and uniquely define the multiplication for that specific matrix pairing such that the output is optimal for the layouts of the input without the user having to specify anything about the problem. To your point:

> The design of nalgebra-sparse is motivated by precisely this observation: a user is likely to want to work with either a CSR matrix, or a CSC matrix. Therefore a user should not need to know anything about a CSC matrix in order to effectively use a CSR matrix, or the other way around.

I think putting this in the type itself achieves that. But whether it's `pub struct CsrMatrix { ... }` or `pub type CsrMatrix<...> = CsMatrix<...>` is an implementation detail that users wouldn't need to worry about? Good examples in the docs would pretty much entirely ameliorate this.

# Removing `&mut self`

> Unfortunately, this observation is completely wrong. Once again, the design of nalgebra-sparse is very deliberate in this regard: if the sparsity pattern does not change, then we can update only the non-zero values of the matrix by accessing only the non-zero values mutably. Taking the Finite Element Method as a motivating use case again, this is precisely how we can re-use our matrix from the previous iteration and directly re-assemble the system without needing to re-allocate any storage.

This is worded rather strongly, but nonetheless: in such cases this assumes a lot about the underlying type that you have available, and the semantics therein. For _any_ generic type `Data` we cannot assume that mutable access will be safe, especially if we're interfacing into C / C++ somehow. There are times when this behaviour makes some stuff easy (e.g. your FEM example), but this is not a useful notion that many public users should necessarily have access to.

I think there's two notions here:

1. `&mut self` for functions that guarantee to only modify non-zero entries.
2. `&mut self` as part of a public API for everyone to use.

I don't think generally on the type (and I mean generally as in, for all generic bounds on the struct) that we should force `Data` to have to implement e.g. `BorrowMut<[T]>`. In such cases where `Data` does implement `BorrowMut<[T]>`, then we can choose to provide it in those specific cases. e.g. for `AddAssign`:

```rust
impl<T, Offset, Majors, Minors, Data, C, Index> AddAssign<T>
    for CsMatrix<T, Offset, Majors, Minors, Data, C, Index>
where
    T: Scalar,
    Offset: Add<usize, Output = usize> + Clone + Into<usize> + Unsigned + Ord,
    Index: Clone + Into<usize> + Unsigned + Ord,
    Majors: Borrow<[Offset]>,
    Minors: Borrow<[Index]>,
    Data: Borrow<[T]> + BorrowMut<[T]>,
    C: Compression,
{
    fn add_assign(&mut self, rhs: T) { ... }
}
```

Here we are adding a single scalar to all non-zero values, so we add `+ BorrowMut<[T]>` to the trait bounds for `Data`. But this isn't possible for every type. You might get a pointer + length in C for the `Data` type, which can be used as `Borrow<[T]>` fairly trivially. Mutating that memory may be unsafe and / or invalid (I mean, you don't know where that pointer has been), so we can't make it part of the struct definition. 

As for making this part of a public API -- should this be something that is public? If you really want to make it so, there is one out:

```rust
impl<T, Offset, Majors, Minors, Data, C, Index>
    CsMatrix<T, Offset, Majors, Minors, Data, C, Index>
where
    Offset: Add<usize, Output = usize> + Clone + Into<usize> + Unsigned + Ord,
    Index: Clone + Into<usize> + Unsigned + Ord,
    Majors: Borrow<[Offset]>,
    Minors: Borrow<[Index]>,
    Data: Borrow<[T]> + BorrowMut<[T]>,
    C: Compression,
{
    fn cs_data_mut(&mut self) -> (&[Offset], &[Index], &mut [T]) { ... }

    fn transpose_mut(&mut self) -> CsMatrix<T, Offset, &[Offset], &[Index], &mut [T], C::Transpose, Index> { ... }

    fn get_entry_mut(&mut self, major_index: usize, minor_index: usize) -> Option<SparseEntryMut<'_, T>> { ... }
}
```

and so on. But this is an extremely limited form of mutability, specifically for non-zeros, and specifically for data that can be borrowed as mutable. Generally speaking though not all data types that could be handled by `nalgebra-sparse` can be `BorrowMut<[T]>`. Many can, especially if `Data` is actually `Vec<T>`, but especially stuff from C or C++ might not necessarily be safe to mutate. C++ would be especially bad here if there was a raw pointer interface to an underlying const iterator of some kind.

# Offsets as count

This one I'm admittedly less married to as an idea. I did it for the simplicity of what I was showing off. I can revert the notation for `MajorOffsets` to be what it was before. It complicates the iterators, but ¯\_(ツ)_/¯

As for:

> Moreover, your suggestion to caching a previous total offset only applies to single-threaded scenarios. In fenris we are assembling CSR system matrices in parallel - in this context, having O(1) access to rows is a necessity.

Well, O(1) aside, Rust has a lot of great tools for parallelization. Using `pmap` from the `rayon` crate is an acceptable way to parallelize iterators. Given the way the iterators are currently written, O(1) lane access isn't really a worry and `rayon` would be able to parallelize any lane-wise algorithm pretty trivially. Again, I'm not married to the idea of using counts but I don't think that parallelization is really a concern.

> This would make the CSR/CSC formats incompatible with other sparse matrix libraries.

This is something that I've tried to accommodate with the generic parameters in `CsMatrix` somewhat. It's a good argument for switching to offsets. However, the same argument could be made in entirely the opposite direction (libraries using count will no longer be compatible with `nalgebra-sparse`), so I guess we pick our poison?

# Conclusion

> I suggest that we focus more on the specific issues that you stated as your motivation:
>  1. Do you really need slices/views, or would a "selection API" be sufficient? If not, can you elaborate on why not?
>  2. We can definitely improve the performance of spMM. But I don't think it's necessary to mix e.g. CSR * CSC to achieve this. In fact, I deliberately avoided mixing matrix types to avoid the combinatorial explosion of possible combinations. We just need a better algorithm.

I appreciate the thoughtful response. However, at least as far as slicing goes, users expect to slice along the major dimension at the very least. There also seems to be some desire to solve #878, and as I've outlined above, I'm not sure how different other solutions would look like. There are some ways to make things simpler (e.g. force `Index = Offset` so we have two fewer types), but ultimately there are only so many ways to generalize index types and allow for borrowed data storage in the CsMatrix type.

Let's put my goals aside for the moment. I added them there to contextualize how I was approaching this and why I was interested in contributing in the first place. We should definitely take any discussion about spGEMM somewhere else. I brought it into this discussion because it was directly related to the choice to make `CompressionKind` one of the generic parameters (so that we know CSR / CSC at compile time, not runtime). Nonetheless, I still think its necessary to solve both #878, and then afterwards see how that might affect the strategy for spGEMM. The iterators I've presented could be a good starting point for how this could be done, and if the final algorithm can be completed functionally (e.g. using `filter_map` or `map`) then parallelization can be trivial through the use of e.g. the rayon crate. Lots to think about, but it's getting into a digression.

Narrowing to the relevant point then:

> 1. Do you really need slices/views, or would a "selection API" be sufficient? If not, can you elaborate on why not?

Yes. I do not want to call `filter` across many triplets (which is what I'm thinking you mean by a selection API). Assume you have a Hessian that is 2000 x 2000. This would not be an abnormally large matrix for the kinds of problems we are tackling. For a dense matrix, that's 4M elements. For a sparse matrix, that's significantly fewer than 4M, but Hessian's may be mostly sparse while still containing a good number of non-zero elements. If we even have 1M non-zeros, then a full slice has to iterate through 1M non-zeros. But consider a block format like the following: 

Screenshot from 2021-11-08 14-44-56

The upper left portion is extremely dense, as are the off diagonal bits. However, the bottom right block is extremely sparse. Schur's complement works by exploiting this relationship. We can do better by being able to generate the Hessian using spGEMM, slicing out the upper portion, and then solving the whole matrix in two steps. In fact, because it's symmetric, we don't even have to really do a full slice along rows and columns, but can simply slice out the left blocks together, and the bottom right block on its own to exploit the same structure & technique. In column major format you can do this purely using just column slicing, as I presented when talking about `SparsityPattern`, and then use `.filter(...)` on the lane iterator to remove entries that are outside the row I care for. This isn't a perfect "slice" in the sense of how dense nalgebra does it, but it is significantly less work. 

In order to do this roughly "for free" we need to be able to treat offsets/indices/data as `Borrow<[ ]>` for some type, which is more or less the description of #878 and the problem therein. I see this as fundamentally the same problem because it lets us borrow storage without potentially selecting across millions of elements -- I just need to know size of the slice and I can construct an iterator that goes through all the elements in X-major order using the code I have now written in this PR:

```rust
let sliced_major = mat.slice_major(lower, upper);

let slice_minor_iter = sliced_major.iter().map(|lane| {
    lane.filter(|(index, value)| {
        index >= index_lower && index < index_upper
    })
});
```

And while this isn't a "free" slice, it does mean as long as I write algorithms to the lane iterator API that I have to do the absolute minimum work to "select" the appropriate elements. But this selection is only optimal because I started by slicing along the major axis.

ThatGeoGuy:
One suggestion for `SparsityPattern`: Rather than using this as an abstraction at a lower level and having to nest the traits / generics / etc, I propose the following.

I assume that one use for `SparsityPattern` is to pass around the pattern itself, so that you're not making a bunch of functions that need to handle lifetimes / types generically. If so, then one option is to present `SparsityPattern` as:

```rust
pub struct SparsityPattern<'a, Offset, Index>
where
    Offset: Add<usize, Output = usize> + Clone + Into<usize> + Unsigned + Ord,
    Index: Clone + Into<usize> + Unsigned + Ord,
{
    offsets: &'a [Offset],
    indices: &'a [Index],
    minor_dim: usize,
}
``` 

We could still trivially have `CsMatrix` output this type and associated API (for compatibility or whatever you want to do with it) without the extra trait overhead. The offset / index part is still included, but at least with this we can avoid a lot of the repetition and still be able to do `CsMatrix::pattern() -> SparsityPattern<'_, O, I>`, assuming that there are functions that consume sparsity patterns and need this abstraction. The pattern type itself adds nothing to the internals of `CsMatrix` aside from nesting (as explained previously), but since `SparsityPattern` already doesn't have any functions on it that take `&mut self`, the above structure should stand in for it while allowing for generic index types.

Thoughts? I still think that the `CsMatrix` type should be flat for pragmatic reasons related to access, as well as avoiding complicated nesting of the types, but this seems like a good middle ground to allow for either mode of access. The only thing that _can't_ be done with this is to pass in `SparsityPattern<'_, O, I>` to a `CsMatrix` constructor and get `MajorOffsets` / `MinorIndices` as `Vec<_>`. This is probably fine though, since `CsMatrix<T, usize, &[usize], &[usize], Vec<T>, CSC, usize>` as an example would still be a valid type (although the lifetime of the type would be tricky, but we could implement an `into_owned(self)` on the type to make it easier.

Anyways, I just wanted to share this as it came to me as I was working on matrix conversions.

Andlon:
Unfortunately I'm very pressed on time lately and have limited time to devote to this, therefore I must be a little blunt: I presented a number of reasons for the very deliberate design choices made in `nalgebra-sparse`. You've subsequently downplayed the importance of these design choices in order to support the introduction of a very complicated API in order to support slicing - which is still not clear how it actually works in detail.

I am not at all convinced that the proposed changes carry their weight. Some we can anyway immediately discard for the reasons I already set out:

- Using counts instead of offsets: this would make us incompatible with all other CSR formats in existence and would destroy O(1) computational complexity for row-lookup. 
- Removing SparsityPattern entirely.
- Removing &mut access to values

The remaining suggestions are made on the assumption that supporting slicing - even at great cost to API complexity - is worthwhile. As I've already stated, I still don't believe that slicing CSR/CSC matrices is worth this complexity. Furthermore, there are two related but ultimately separate concerns here, so let's clarify:

- Construct CSR/CSC/SparsityPattern from borrowed data. This was the motivation behind #878, **not** slicing (which was never mentioned in the original issue).
- Construct "slices" into existing CSR/CSC data.

The first is very important for being able to inter-op with other libraries/applications with low overhead. I still maintain that the latter is not very useful.

> However, at least as far as slicing goes, users expect to slice along the major dimension at the very least.

I think for virtually all users, this use case can be supported by efficient *selection*: to be able to *select* a set of rows/cols and copy it over to a new matrix or even store in parts of an existing matrix. It might look like this (example from #878):

```rust
// Allocate new matrix containing the select rows/cols
let selection = matrix.select(rows, cols).clone_owned();
// Store selection into existing matrix (might need to re-allocate matrix storage completely if non-zero entries are not already present)
output.select_mut(output_rows, output_cols).assign(&matrix.select(rows, cols));
```

This can be implemented fairly efficiently. Since CSR has O(1) major dimension lookup, we can quickly look up a row. Once we have a row, we can also efficiently pick out the correct columns without necessarily visiting them all, using a combination of linear and binary/exponential search. The expected complexity for selection would be `O(n_sel + nnz_sel)`, where `n_sel` is the number of rows in the selection and `nnz_sel` is the number of *selected* non-zeros, not the number of non-zeros in the whole matrix. There is no need to `.filter` over all triplets/non-zeros in the matrix.

Would a selection API like this not satisfy your use case for working with blocks of a matrix?

## Borrowed data

There still remains the use case of abstracting over owned/borrowing data. This is important and I want to support this. But I don't think the design you're proposing is the way to go, as it exposes too many generics throughout the API. I think it's feasible to hide much of the complexity behind `CsStorage` and `CsStorageMut` traits. I don't have more time right now to go into detail on this, unfortunately.

Andlon:
Based on the previous discussion, I will close this PR, which has lingered for a long time by now. Thanks for the interest in making `nalgebra-sparse` better!

PR #1039, opened 2021-12-01
Remove redundant proptest patch in `nalgebra-sparse`

Andlon:
Now that we're using `proptest` 1.0, the issue that the patch was designed to tackle has been fixed, so we can remove our patched code for `Shuffle` and rely on `.proptest_shuffle()` from `proptest` directly.


Andlon:
@sebcrozet: tests pass, so this should be safe to merge. We merely remove some internal things that are no longer necessary, without any changes to the API.

sebcrozet:
Sounds good, thanks!

PR #1043, opened 2021-12-09
Fix the special-case for 3x3 Real SVD

sebcrozet:
I stumbled upon a matrix that failed the special-case for 3x3 SVD on real matrices that I introduced in #1034.
It turns out that I skipped an important step: sorting the singular values before running the QR decomposition.

In this PR, I:
- add a sorting step in the the 3x3 SVD. This doesn’t appear to change the performances on either CPU or GPU.
- add a test to skip the `sort_by_singular_values` called by `matrix.svd()` on 2x2 and 3x3 real matrices (because the special SVD implementation always sort the singular values anyway).
- add to most of our SVD tests an assertion  checking that the singular values are actually sorted.
- add the problematic 3x3 matrix I found to the tests.

I’m actually surprised that our proptest tests didn’t catch this edge-case earlier (we even already had a specific test for 3x3 matrices).

### DIFF
--- a/tests/linalg/svd.rs
+++ b/tests/linalg/svd.rs
@@ -1,3 +1,4 @@
+use crate::utils::is_sorted_descending;
 use na::{DMatrix, Matrix6};
 
 #[cfg(feature = "proptest-support")]
@@ -14,6 +15,7 @@ mod proptest_tests {
                 use crate::core::helper::{RandScalar, RandComplex};
                 use crate::proptest::*;
                 use proptest::{prop_assert, proptest};
+                use crate::utils::is_sorted_descending;
 
                 proptest! {
                     #[test]
@@ -26,7 +28,7 @@ mod proptest_tests {
                         prop_assert!(s.iter().all(|e| *e >= 0.0));
                         prop_assert!(relative_eq!(&u * ds * &v_t, recomp_m, epsilon = 1.0e-5));
                         prop_assert!(relative_eq!(m, recomp_m, epsilon = 1.0e-5));
-                        prop_assert!(s.as_slice().windows(2).all(|elts| elts[0] >= elts[1]));
+                        prop_assert!(is_sorted_descending(s.as_slice()));
                     }
 
                     #[test]
@@ -39,7 +41,7 @@ mod proptest_tests {
                         prop_assert!(relative_eq!(m, &u * ds * &v_t, epsilon = 1.0e-5));
                         prop_assert!(u.is_orthogonal(1.0e-5));
                         prop_assert!(v_t.is_orthogonal(1.0e-5));
-                        prop_assert!(s.as_slice().windows(2).all(|elts| elts[0] >= elts[1]));
+                        prop_assert!(is_sorted_descending(s.as_slice()));
                     }
 
                     #[test]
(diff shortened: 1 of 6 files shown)


tests/linalg/svd.rs
```diff
@@ -350,6 +361,26 @@ fn svd_fail() {
     assert_relative_eq!(m, recomp, epsilon = 1.0e-5);
 }
 
+#[test]
+#[rustfmt::skip]
+fn svd3_fail() {
+    let m = nalgebra::matrix![
```
Andlon, line 367: Might be a good idea to add a comment that this is a real case that actually failed at some point. Otherwise in a few years' time it might look like a very arbitrary test

tests/linalg/svd.rs
```diff
@@ -38,6 +39,7 @@ mod proptest_tests {
                         prop_assert!(relative_eq!(m, &u * ds * &v_t, epsilon = 1.0e-5));
                         prop_assert!(u.is_orthogonal(1.0e-5));
                         prop_assert!(v_t.is_orthogonal(1.0e-5));
+                        prop_assert!(s.as_slice().windows(2).all(|elts| elts[0] >= elts[1]));
```
Andlon, line 42: might be nice to have a utility method `is_sorted(slice)`..

Andlon (commented):
Just happened to look at it because I was interested in the `proptest` failures and just had a couple of thoughts (not important to address, just thought it might be nice).

Anyway, I'm curious about that `proptest` didn't capture them: are you saying that *any* decomposition in which the singular values happened to be unsorted would give a wrong result with this method? If so I would hope a case like this would be captured with `proptest`. On the other hand, if the set of matrices which fail is very small, then I guess maybe not...

Might be worthwhile to investigate the distribution of matrices generated by our proptest generators to look for potential issues :thinking:

Andlon:
By the way, did you try rerunning the proptests but with `PROPTEST_CASES=1000000` or something similarly high? (maybe run with `--release`) Curious if we just didn't test *enough* matrices?

tests/linalg/svd.rs
sebcrozet: Comment added.

tests/linalg/svd.rs
sebcrozet: Done.
Actually, `slice.is_sorted` already exists: https://doc.rust-lang.org/std/primitive.slice.html#method.is_sorted
But it hasn’t been stabilized yet. Also, it checks that it’s sorted in ascending order (though we could use `is_sorted_by` instead).

sebcrozet:
@Andlon 
> Anyway, I'm curious about that proptest didn't capture them: are you saying that any decomposition in which the singular values happened to be unsorted would give a wrong result with this method?

According to the paper this is based on, this happens sometimes when the matrix has at least one 0 singular values, and that singular value isn’t located at the end  of the vector of singular values. So I’m guessing this happens somewhat rarely.

> By the way, did you try rerunning the proptests but with PROPTEST_CASES=1000000 or something similarly high? (maybe run with --release) Curious if we just didn't test enough matrices?

Ha! You are right, using `PROPTEST_CASES=1000000` detects the bug.

Andlon:
That's very interesting! Hmm, maybe this is a case for occasionally running tests in release mode with `PROPTEST_CASES` very high? Maybe whenever something gets merged to `dev`? Not sure.

sebcrozet:
> That's very interesting! Hmm, maybe this is a case for occasionally running tests in release mode with `PROPTEST_CASES` very high? Maybe whenever something gets merged to `dev`? Not sure.

Maybe. We should think about it. Perhaps when merging something to `master` (i.e. before a release)?

Andlon:
> > That's very interesting! Hmm, maybe this is a case for occasionally running tests in release mode with `PROPTEST_CASES` very high? Maybe whenever something gets merged to `dev`? Not sure.
> 
> Maybe. We should think about it. Perhaps when merging something to `master` (i.e. before a release)?

Ah that's a good idea. Yeah, just run as many tests as possible upon merging to `master` might be a decent strategy (can we run things like `miri` as well? :thinking: )

sebcrozet:
> can we run things like miri as well?

I guess we could, yes.

Issue #1046, opened 2021-12-13
Matrix::select_rows and Matrix::select_columns requires T: Zero


haibane-tenshi:
The fact is somewhat surprising when trying to use matrix types with non-numerical data.
Internal implementation doesn't seem to rely on the trait, can the restriction be lifted?

CattleProdigy:
@haibane-tenshi Looks like `select_rows()` and `select_columns()` got grouped (not unreasonably) with `lower_triangular()` and `upper_triangular()` into one `impl` block with `T:Zero`. The triangular methods use `T::zero()` to fill out the rest of extracted matrix. You should be able to move the select methods into their own block without  `T:Zero` in `src/base/edition.rs` and submit a PR.

haibane-tenshi:
Sure, I'll make a PR when I get my hands on it. The question mostly was if it is ok to lift it or not :)

haibane-tenshi:
PR got merged, closing issue.

PR #1048, opened 2021-12-16
Relax T: Zero bound on Matrix::select_rows() and Matrix::select_columns()

haibane-tenshi:
Original issue:  #1046.

Should I add  "Triangular matrix extraction" bullet to Matrix documentation, now as it is decoupled from select functions?


sebcrozet:
Thanks! I took the liberty of adding the bullet to the Matrix documentation as you suggested.

PR #656, opened 2019-09-24
Polar decomposition

LucasCampos:
Hello. As detailed in [this thread](https://discourse.nphysics.org/t/matrix-polar-decomposition/409) in the user forum, I am implementing the [matricial polar decomposition](https://en.wikipedia.org/wiki/Polar_decomposition) in nalgebra.

The bulk of the code is ready. However, I am having some issues with the traits needed to properly compile the code. 

Cleanup and further implementations are still in the 'to do' list.

Any pointers on the specific type traits needed would be appreciated.

The compile errors are

```
error[E0277]: cannot multiply `base::matrix::Matrix<<N as alga::general::ComplexField>::RealField, base::dimension::Dynamic, base::dimension::Dynamic, base::vec_storage::VecStorage<<N as alga::general::ComplexField>::RealField, base::dimension::Dynamic, base::dimension::Dynamic>>` to `base::matrix::Matrix<N, base::dimension::Dynamic, base::dimension::Dynamic, base::vec_storage::VecStorage<N, base::dimension::Dynamic, base::dimension::Dynamic>>`
  --> src/linalg/polar.rs:76:39
   |
76 |                     Some(v_t.adjoint()*DMatrix::from_diagonal(&svd.singular_values)*v_t)
   |                                       ^ no implementation for `base::matrix::Matrix<N, base::dimension::Dynamic, base::dimension::Dynamic, base::vec_storage::VecStorage<N, base::dimension::Dynamic, base::dimension::Dynamic>> * base::matrix::Matrix<<N as alga::general::ComplexField>::RealField, base::dimension::Dynamic, base::dimension::Dynamic, base::vec_storage::VecStorage<<N as alga::general::ComplexField>::RealField, base::dimension::Dynamic, base::dimension::Dynamic>>`
   |
   = help: the trait `std::ops::Mul<base::matrix::Matrix<<N as alga::general::ComplexField>::RealField, base::dimension::Dynamic, base::dimension::Dynamic, base::vec_storage::VecStorage<<N as alga::general::ComplexField>::RealField, base::dimension::Dynamic, base::dimension::Dynamic>>>` is not implemented for `base::matrix::Matrix<N, base::dimension::Dynamic, base::dimension::Dynamic, base::vec_storage::VecStorage<N, base::dimension::Dynamic, base::dimension::Dynamic>>`
[...]

PR #1050, opened 2021-12-22
Take-2 of polar-decomposition

metric-space:
Code inspired by this PR/thread: https://github.com/dimforge/nalgebra/pull/656

Main person behind this is @LucasCampos

This PR  implements  [matrix polar decomposition](https://en.wikipedia.org/wiki/Polar_decomposition)

### DIFF
--- a/src/linalg/decomposition.rs
+++ b/src/linalg/decomposition.rs
@@ -188,7 +188,7 @@ impl<T: ComplexField, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
         SVD::try_new_unordered(self.into_owned(), compute_u, compute_v, eps, max_niter)
     }
 
-    /// Attempts to compute the Polar Decomposition of  a `matrix
+    /// Attempts to compute the Polar Decomposition of  a `matrix` (indirectly uses SVD)
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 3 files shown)


sebcrozet:
Thanks! I took the liberty of making a few small changes:
- I renamed `Matrix::polar` to `Matrix::try_polar` and added `Matrix::polar` which returns the result instead of `Option` (similar to the other methods, e.g., `Matrix::svd` and `Matrix::try_svd`).
- I made your test use `proptest`  instead of `DMatrix::new_random`. This will cover more cases.

The new proptest test is failing right now. I haven’t investigated in details yet to determine if this is a bug, or if the test are just failing because we used an epsilon that is too restrictive.

metric-space:
@sebcrozet Thank you for the review and the additions. Not sure if you plan to look into it further but I will take look into the bug with the prop test and return back with a fix/solution

metric-space:
I zeroed in on the causes. 

1. The orthogonal test is a proxy for a unitary/semi-unitary check. It's a unitary check for square matrices and supposed to be a semi-unitary check for rectangular matrices which is a bit more involved because it tests whether either the rows are orthogonal or the columns are orthogonal. It depends on the dimensions of the rectangular matrix

2. In the main test, the test svd  generated  is ordered while the method based `polar` calls the unordered SVD for further processing, which gives different decompositions

Fixes applied 
Ball's back in your court @sebcrozet

sebcrozet:
Perfect, thanks!

metric-space:
Thank you

PR #1069, opened 2022-01-24
add support for matrix market export

losanc:
Hi,
I have added support for exporting sparse matrix into matrix market. The difference between import and export is that, export using `SparseAccess` trait, which can export all three formats of sparse matrix.  Also added two more methods for `SupportedMatrixMarketScalar` trait, which are  `typename()` and `to_matrixmarket_string`

### DIFF
--- a/nalgebra-sparse/src/io/matrix_market.rs
+++ b/nalgebra-sparse/src/io/matrix_market.rs
@@ -1,9 +1,10 @@
 //! Implementation of matrix market io code.
 //!
 //! See the [website](https://math.nist.gov/MatrixMarket/formats.html) or the [paper](https://www.researchgate.net/publication/2630533_The_Matrix_Market_Exchange_Formats_Initial_Design) for more details about matrix market.
-use crate::coo::CooMatrix;
+use crate::CooMatrix;
 use crate::SparseFormatError;
 use crate::SparseFormatErrorKind;
+use matrixcompare_core::SparseAccess;
 use nalgebra::Complex;
 use pest::iterators::Pairs;
 use pest::Parser;
@@ -12,7 +13,8 @@ use std::convert::Infallible;
 use std::convert::TryFrom;
 use std::fmt;
 use std::fmt::Formatter;
-use std::fs;
+use std::fs::{self, File};
+use std::io::Write;
 use std::num::ParseIntError;
 use std::num::TryFromIntError;
 use std::path::Path;
@@ -267,7 +269,7 @@ impl fmt::Display for MatrixMarketError {
                 write!(f, "InvalidHeader,")?;
             }
             MatrixMarketErrorKind::EntryMismatch => {
-                write!(f, "EntryNumUnmatched,")?;
+                write!(f, "EntryMismatch,")?;
             }
             MatrixMarketErrorKind::TypeMismatch => {
                 write!(f, "TypeMismatch,")?;
@@ -288,7 +290,7 @@ impl fmt::Display for MatrixMarketError {
                 write!(f, "NotLowerTriangle,")?;
             }
             MatrixMarketErrorKind::NonSquare => {
-                write!(f, "NotSquareMatrix,")?;
+                write!(f, "NonSquare,")?;
             }
         }
         write!(f, " message: {}", self.message)
@@ -506,6 +508,10 @@ mod internal {
         fn negative(self) -> Result<Self, MatrixMarketError>;
         /// When matrix is a Hermitian matrix, it will convert itself to its conjugate.
         fn conjugate(self) -> Result<Self, MatrixMarketError>;
+        /// Returns the name of SupportedMatrixMarketScalar, used when write the matrix
+        fn typename() -> &'static str;
+        /// Convert the data to string
+        fn to_matrixmarket_string(&self) -> String;
     }
 }
 
@@ -557,6 +563,14 @@ macro_rules! mm_int_impl {
             fn negative(self) -> Result<Self, MatrixMarketError> {
                 Ok(-self)
             }
+            #[inline]
+            fn typename() -> &'static str {
+                "integer"
+            }
+            #[inline]
+            fn to_matrixmarket_string(&self) -> String {
+                self.to_string()
+            }
         }
     };
 }
@@ -601,6 +615,14 @@ macro_rules! mm_real_impl {
             #[inline]
             fn negative(self) -> Result<Self, MatrixMarketError> {
                 Ok(-self)
+            }
+            #[inline]
+            fn typename() -> &'static str {
+                "real"
+            }
+            #[inline]
+            fn to_matrixmarket_string(&self) -> String {
+                self.to_string()
             }
         }
     };
@@ -647,6 +669,14 @@ macro_rules! mm_complex_impl {
             #[inline]
             fn negative(self) -> Result<Self, MatrixMarketError> {
                 Ok(-self)
+            }
+            #[inline]
+            fn typename() -> &'static str {
+                "complex"
+            }
+            #[inline]
+            fn to_matrixmarket_string(&self) -> String {
+                self.re.to_string() + " " + &self.im.to_string()
             }
         }
     };
@@ -696,6 +726,15 @@ macro_rules! mm_pattern_impl {
                     MatrixMarketErrorKind::TypeMismatch,
                     format!("Pattern type has no negative"),
                 ))
+            }
+            #[inline]
+            fn typename() -> &'static str {
+                "pattern"
+            }
+            #[inline]
+            fn to_matrixmarket_string(&self) -> String {
+                // pattern type will return an empty string
+                String::new()
             }
         }
     };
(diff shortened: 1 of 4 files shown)


Andlon (commented):
This is great @losanc!

I'll conduct a detailed review later in the process. Here are some high-level things to discuss first:

- Extending the internal `MatrixMarketScalar` trait like this is a great way to implement this.
- I think `matrixcompare::SparseAccess` is not the right abstraction to use here. It's really only intended for use with `matrixcompare`. For one, it's geared towards small-ish examples used in tests, not for exporting potentially big matrices to file.
  - Maybe creating a sealed/internal trait `MatrixMarketExport<T>`  implemented for the various matrix types could be useful? All matrices do actually have some kind of triplet iterator, so we could leverage this. The tricky part here is that `CooMatrix` only returns `impl Iterator<...>` which we cannot really use as an associated type. I'm not completely sure what the best way forward here is. Do you have some thoughts?
- Building the string in memory first is not quite ideal, since a big matrix might require a large amount of memory when encoded as a string.
  - Can we make it work with the `Write` trait and build on top of this, perhaps?

Andlon:
I just took a brief look: the changes with the new internal trait look good. Did you think about leveraging the `std::io::Write` trait? I think you could basically just use your impl for `write_to_matrix_market_file` with a `w: Write` parameter, and then forward to this impl from the file export method. You could also implement your `_str` method by passing in a `Vec<u8>` and converting it to a `String` afterwards.

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -1329,3 +1417,110 @@ fn next_dense_coordinate(
...
+/// assert_matrix_eq!(matrix,generated_matrixmarket);
+/// ```
+pub fn write_to_matrix_market_str<T: MatrixMarketScalar, S: MatrixMarketExport<T>>(
+    sparse_matrix: &S,
+) -> String {
+    let mut bytes = Vec::<u8>::new();
+    // This will call impl<A: Allocator> Write for Vec<u8, A>
+    // The vector will grow as needed.
+    // So, unwrap here won't cause any issue.
+    write_to_matrix_market(&mut bytes, sparse_matrix).unwrap();
```
Andlon, line 1450: This comment makes me wonder if perhaps `write_to_matrix_market` shouldn't even return a `MatrixMarketError`, because the only thing that can go wrong are IO errors? Or is there something else? I would suppose that we should *always* be able to *export* matrix market data without errors, right?

Andlon (changes requested):
Think we're on a good track here! I think the biggest remaining issue is the fact that we're allocating a `Vec<u8>` every time we write any kind of scalar. This means one heap allocation per scalar, which is a pretty massive potential overhead. I made some comments for suggestions for how we can improve this - maybe you have some thoughts on this as well.

In addition some minor comments.

Also, a tiny nitpick: I think following Rust's naming conventions, we should write `matrix_market` rather than `matrixmarket`

nalgebra-sparse/src/io/matrix_market.rs
losanc: yes, I think the only thing could go wrong is IO error. Since I wrap up the error when loading the matrix by `load_from_file`, so I just thought to make it kind of consistent, also give an MatrixMarketError when `write_to_file`

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -1329,3 +1417,110 @@ fn next_dense_coordinate(
...
+    // safety issue is because 'from_utf8_unchecked' does not check that the bytes passed to it are valid UTF-8.
+    // Since 'bytes' created here is valid UTF-8 data, so it will not cause any problem.
+    unsafe { String::from_utf8_unchecked(bytes) }
```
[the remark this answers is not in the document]
losanc: Because everything, which will be written, is controlled by the code, not the user. And from the code side, we only write header, one line of comment and data, which  is only implemented for basic data types, e.g. int, float. So that's why I think it should be safe to use `from_utf8_unchecked`, 

however, there is another 2 safe operations, one is [from_utf8](https://doc.rust-lang.org/stable/std/string/struct.String.html#method.from_utf8), which gives an error when non-utf8 data encountered, another one is [from_utf8_lossy](https://doc.rust-lang.org/stable/std/string/struct.String.html#method.from_utf8_lossy), which converts non-utf8 character to `std::char::REPLACEMENT_CHARACTER`, it looks like this �

so I don't know which one you think would be better

losanc:
> Also, a tiny nitpick: I think following Rust's naming conventions, we should write matrix_market rather than matrixmarket

So, `matrix_market` when using as lowercase, e.g. function signature etc. and  `MatrixMarket` when using as uppercase, e.g. struct name, etc,  like this?

nalgebra-sparse/src/io/matrix_market.rs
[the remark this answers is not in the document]
Andlon: I agree that with the current implementation, it's probably fine. Basically as long as we're using `write!` everywhere we should be in the clear. But we'd also like code to be robust to later changes, and this is a very implicit invariant that might not be obvious for someone working on the code base later on.

I think `from_utf8(bytes).expect("..")` is probably the way to go: after all, our code *should* be unicode-compatible, so if it's not then there must be a bug in our program.

nalgebra-sparse/src/io/matrix_market.rs
Andlon: Maybe we should rename `MatrixMarketError` to `MatrixMarketParseError` and simply return only an IO error for export?

Andlon:
Yep, that's right!

nalgebra-sparse/src/io/matrix_market.rs
```diff
@@ -1329,3 +1429,112 @@ fn next_dense_coordinate(
...
+/// %%matrixmarket matrix coordinate integer general
+/// 5 4 2
+/// 1 1 10
+/// 2 3 5
+/// "#;
+/// let matrix = load_coo_from_matrix_market_str::<i32>(&str).unwrap();
+/// let res = write_to_matrix_market_file(&matrix,"path/to/matrix.mtx");
+/// if res.is_err(){
+/// // do something
+/// }
```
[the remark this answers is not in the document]
losanc: I think `?` can not be used in doc, because it requires the function signature `fn fun() -> Result<>`, but in the doc, we don't write the function signature, and it's controlled by `cargo test` .
Unless we use `ignore` attribute for this example, otherwise this example can't compile

Andlon:
@losanc: I've lost a little bit overview. Is this ready for review? Also, please don't force push your branch, because now I cannot track the history and easily see what was addressed...

losanc:
Sorry for the late reply,  yes, I think it's ready for the review

nalgebra-sparse/tests/unit_tests/matrix_market.rs
```diff
@@ -336,10 +344,123 @@ fn test_matrixmarket_dense_complex_general() {
...
+        let generated_matrix = load_coo_from_matrix_market_str(&generated_matrixmarket_string).unwrap();
+        assert_matrix_eq!(generated_matrix, coo);
+    }
+}
+
+proptest! {
+    #[test]
+    fn coo_matrix_market_roundtrip_file(coo in coo_no_duplicates(-10 ..= 10, 0 ..= 10, 0..= 10, 100)) {
+        let mut tempdir = std::env::temp_dir();
+        tempdir.push("temp.mtx");
```
Andlon, line 461: Please use `tempfile::TempDir` instead, which will create a randomly named directory to make tests independent. It will also delete the directory after use.

Andlon:
@losanc: I'm sorry, but due to the force push we lost history and I cannot easily see if my comments were addressed. Could you perhaps go over the comments again and let me know if any of them were not addressed? I cannot answer to the comments anymore because they are listed as outdated.

I see I didn't answer this comment though:
> I think ? can not be used in doc, because it requires the function signature fn fun() -> Result<>, but in the doc, we don't write the function signature, and it's controlled by cargo test .
Unless we use ignore attribute for this example, otherwise this example can't compile

You can change the function signature of the "main" function in the examples. For example:

```rust
/// Function docs.
///
/// The code behind `#` will be invisible to the example, but understood by the compiler/`rustdoc`.
/// ```rust
/// # fn main() -> Result<(), Box<dyn Error>> {
///  example_code()?;
/// # }
```

Andlon (commented):
Just one comment here, I'll make another review once you get back to me about the resolution of the comments.

losanc:
Hi, 
I will answer all the comments above,
1. typo `coordinate`, fixed
2. About the unit test, now I structured it into 4 parts, 
    1. read matrix from string, which contains most of the cases, (not all cases, which is too much), then compare the matrix with dense matrix
    2. write matrix (`coomatrix` only) to string, then compare the string
    3. roundtrip test, generate random matrix, then write it to string, then read it back as the matrix, compares the original matrix with new read matrix
    4.  roundtrip test, same as 3, but write matrix into a file, and read it from the file
3. When writing to a file, [BufWriter](https://doc.rust-lang.org/std/io/struct.BufWriter.html) is used
4. `fn write_to_matrix_market<T: MatrixMarketScalar, S: MatrixMarketExport<T>, W: Write>(` is now public
5. `write_to_matrix_market_file` doc test, I use the `no_run`, and `write_to_matrix_market_file(&matrix,"path/to/matrix.mtx").unwrap();`
    + From what I have tested, even with this code
```rust
/// # fn main() -> Result<(), Box<dyn Error>> {
///  example_code()?;
/// # }
```
 because ` write_to_matrix_market_file(&matrix,"path/to/matrix.mtx")?` will always return the error, and as a result, the doc test would still fail, because the `main` function in this doc test didn't return success
6. `tempfile::TempDir` has been used, instead of `std::env::temp_dir`
7. slightly change to cargo.toml, to use `tempfile` 


I think this would be all the changed I made

Andlon:
@losanc: awesome, thanks for all your hard work here! I took the liberty to do some final polish, mainly for the docs. I also decided to change `write_to_matrix_market...` to `save_to_matrix_market` so that it's consistent with our existing terminology (`save` is the "opposite" of `load`).

> because write_to_matrix_market_file(&matrix,"path/to/matrix.mtx")? will always return the error, and as a result, the doc test would still fail, because the main function in this doc test didn't return success

The key here is to use `# Ok(()) }` to end the `main` function! See how I did this in one of the commits I pushed.

Can you please look over my changes and see if you agree to them?

Once we're in agreement I'll fix the conflict and merge :)

losanc:
I think these changes are fine, I have no other comments

Andlon:
@losanc: All right, then this PR is merged :-) Thanks for all your hard work, excited to have this functionality in `nalgebra-sparse`!

Issue #1011, opened 2021-10-17
Violation of stacked borrows in abomonation tests


saethlin:
None of the abomonation tests in this crate pass miri because they all invoke UB by producing improperly-aligned references and doing improperly-aligned reads. That level of UB is par for the course with abomonation, but this one also violates stacked borrows:
```
$ cargo miri test --all-features --no-fail-fast abomonate_point3
    Finished test [unoptimized + debuginfo] target(s) in 0.04s
     Running unittests (target/miri/x86_64-unknown-linux-gnu/debug/deps/nalgebra-8bb505112d667bf7)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 20 filtered out

     Running tests/lib.rs (target/miri/x86_64-unknown-linux-gnu/debug/deps/lib-17c301152c55f7ae)

running 1 test
test core::abomonation::abomonate_point3 ... error: Undefined Behavior: trying to reborrow for Unique at alloc644077+0x1, but parent tag <1628945> does not have an appropriate item in the borrow stack
   --> /home/ben/.cargo/registry/src/github.com-1ecc6299db9ec823/abomonation-0.7.3/src/lib.rs:127:30
    |
127 |         let result: &mut T = mem::transmute(split1.get_unchecked_mut(0));
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ trying to reborrow for Unique at alloc644077+0x1, but parent tag <1628945> does not have an appropriate item in the borrow stack
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information

    = note: inside `abomonation::decode::<na::OPoint<f32, na::Const<3_usize>>>` at /home/ben/.cargo/registry/src/github.com-1ecc6299db9ec823/abomonation-0.7.3/src/lib.rs:127:30

```

sebcrozet:
Thank you for pointing this out. I thin we should simply remove the support for `abomonation` completely.

It is inherently unsafe, doesn’t seem to be actively maintained anymore, and doesn’t really appears to be very popular. I feel like it made sense to support this in the 4 years ago, when the serialization scene wasn’t super well developed in Rust. Now it seems like there are better and safer alternative.

[...]

PR #1074, opened 2022-02-05
Remove abomonation support

saethlin:
Abomonation has numerous soundness problems which have been well-documented in its issue tracker for over 2 years. Some of them could be fixed, but some are fundamental to its design. If a user wants super-fast ser/de, they should use rkyv. See https://rustsec.org/advisories/RUSTSEC-2021-0120.html.

Closes https://github.com/dimforge/nalgebra/issues/1011


sebcrozet:
Thanks! I think this is the right move.
Unfortunately the CI doesn’t want to run of this PR (likely because the workflow has been modified), but I’m confident that the changes are straightforward enough not to cause any issue.

saethlin:
> likely because the workflow has been modified

Have you seen this happen before? Most of my experience is with Travis, which occasionally just seems to miss a commit. I'm going to push an empty commit here and see if that causes the CI to wake up.

saethlin:
Ah-ha! Now I see "1 workflow awaiting approval"

sebcrozet:
> Have you seen this happen before?

I thought I saw this before, but I may be misremembering. Looks like your additional commit did the trick, thanks!

PR #1081, opened 2022-02-23
Make sparse-times-sparse faster

smr97:
This PR changes the sparse times sparse matrix computation kernel, such that the algorithmic cost reduces.
The earlier implementation was searching in the result matrix for the correct column position to update a non-zero value.
This implementation uses a dense array for accumulation. This is also called the scatter-vector accumulation, and has been well studied in research on sparse algebra.

There is an SPMM example that I wrote to compare performance on a SuiteSparse matrix, and it seems to have reduced from 13 seconds to 4 seconds on my machine (11th Gen Intel i7-11700F, 16M L3, 4M L2 cache).

### DIFF
--- a/nalgebra-sparse/Cargo.toml
+++ b/nalgebra-sparse/Cargo.toml
@@ -48,3 +48,9 @@ path = "example/spmm.rs"
 [package.metadata.docs.rs]
 # Enable certain features when building docs for docs.rs
 features = [ "proptest-support", "compare" ]
+
+[profile.release]
+opt-level = 3 
+lto = "fat"
+codegen-units = 1
+panic = "abort"
--- a/nalgebra-sparse/Cargo.toml
+++ b/nalgebra-sparse/Cargo.toml
@@ -24,6 +24,7 @@ io      = [ "pest", "pest_derive" ]
 slow-tests = []
 
 [dependencies]
+criterion = { version = "0.3", features = ["html_reports"] }
 nalgebra = { version="0.30", path = "../" }
 num-traits = { version = "0.2", default-features = false }
 proptest = { version = "1.0", optional = true }
@@ -31,12 +32,18 @@ matrixcompare-core = { version = "0.1.0", optional = true }
 pest           = { version = "2", optional = true }
 pest_derive    = { version = "2", optional = true }
 serde = { version = "1.0", default-features = false, features = [ "derive" ], optional = true }
+itertools = "0.10"
 
 [dev-dependencies]
 itertools = "0.10"
 matrixcompare = { version = "0.3.0", features = [ "proptest-support" ] }
 nalgebra = { version="0.30", path = "../", features = ["compare"] }
 serde_json = "1.0"
+
+[[example]]
+name = "spmm"
+required-features = ["io"]
+path = "example/spmm.rs"
 
 [package.metadata.docs.rs]
 # Enable certain features when building docs for docs.rs
--- a/nalgebra-sparse/src/ops/serial/cs.rs
+++ b/nalgebra-sparse/src/ops/serial/cs.rs
@@ -1,16 +1,19 @@
+use std::collections::HashSet;
+
 use crate::cs::CsMatrix;
 use crate::ops::serial::{OperationError, OperationErrorKind};
 use crate::ops::Op;
 use crate::SparseEntryMut;
+use itertools::Itertools;
 use nalgebra::{ClosedAdd, ClosedMul, DMatrixSlice, DMatrixSliceMut, Scalar};
 use num_traits::{One, Zero};
 
-fn spmm_cs_unexpected_entry() -> OperationError {
-    OperationError::from_kind_and_message(
-        OperationErrorKind::InvalidPattern,
-        String::from("Found unexpected entry that is not present in `c`."),
-    )
-}
+//fn spmm_cs_unexpected_entry() -> OperationError {
+//    OperationError::from_kind_and_message(
+//        OperationErrorKind::InvalidPattern,
+//        String::from("Found unexpected entry that is not present in `c`."),
+//    )
+//}
 
 /// Helper functionality for implementing CSR/CSC SPMM.
 ///
@@ -32,28 +35,54 @@ where
 {
     for i in 0..c.pattern().major_dim() {
         let a_lane_i = a.get_lane(i).unwrap();
+
+        let some_val = Zero::zero();
+        let mut scratchpad_values: Vec<T> = vec![some_val; b.pattern().minor_dim()];
+        let mut scratchpad_indices: HashSet<usize> = HashSet::new();
+
         let mut c_lane_i = c.get_lane_mut(i).unwrap();
-        for c_ij in c_lane_i.values_mut() {
-            *c_ij = beta.clone() * c_ij.clone();
-        }
+        //let (indices, values) = c_lane_i.indices_and_values_mut();
+        //indices
+        //    .iter()
+        //    .zip(values.iter())
+        //    .for_each(|(id, val)| scratchpad_values[*id] = beta.clone() * val.clone());
+
+        //for (index, c_ij) in c_lane_i.indices_and_values_mut() {
+        //    *c_ij = beta.clone() * c_ij.clone();
+        //}
 
         for (&k, a_ik) in a_lane_i.minor_indices().iter().zip(a_lane_i.values()) {
             let b_lane_k = b.get_lane(k).unwrap();
-            let (mut c_lane_i_cols, mut c_lane_i_values) = c_lane_i.indices_and_values_mut();
+            //let (mut c_lane_i_cols, mut c_lane_i_values) = c_lane_i.indices_and_values_mut();
             let alpha_aik = alpha.clone() * a_ik.clone();
             for (j, b_kj) in b_lane_k.minor_indices().iter().zip(b_lane_k.values()) {
                 // Determine the location in C to append the value
-                let (c_local_idx, _) = c_lane_i_cols
-                    .iter()
-                    .enumerate()
-                    .find(|(_, c_col)| *c_col == j)
-                    .ok_or_else(spmm_cs_unexpected_entry)?;
+                // TODO make a scratchpad and defer the accumulation into C after processing one
+                // full row of A.
+                scratchpad_values[*j] += alpha_aik.clone() * b_kj.clone();
+                scratchpad_indices.insert(*j);
+                //let (c_local_idx, _) = c_lane_i_cols
+                //    .iter()
+                //    .enumerate()
+                //    .find(|(_, c_col)| *c_col == j)
+                //    .ok_or_else(spmm_cs_unexpected_entry)?;
 
-                c_lane_i_values[c_local_idx] += alpha_aik.clone() * b_kj.clone();
-                c_lane_i_cols = &c_lane_i_cols[c_local_idx..];
-                c_lane_i_values = &mut c_lane_i_values[c_local_idx..];
+                //c_lane_i_values[c_local_idx] += alpha_aik.clone() * b_kj.clone();
+                //c_lane_i_cols = &c_lane_i_cols[c_local_idx..];
+                //c_lane_i_values = &mut c_lane_i_values[c_local_idx..];
             }
         }
+        // sort the indices, and then access the relevant indices (in sorted order) from values
+        // into C.
+        let sorted_indices: Vec<usize> =
+            Itertools::sorted(scratchpad_indices.into_iter()).collect();
+        c_lane_i
+            .values_mut()
+            .iter_mut()
+            .zip(sorted_indices.into_iter())
+            .for_each(|(output_ref, index)| {
+                *output_ref = beta.clone() * output_ref.clone() + scratchpad_values[index].clone()
+            });
     }
 
     Ok(())
--- a/nalgebra-sparse/src/ops/serial/cs.rs
+++ b/nalgebra-sparse/src/ops/serial/cs.rs
@@ -1,10 +1,9 @@
-use std::collections::HashSet;
+//use std::collections::HashSet;
 
 use crate::cs::CsMatrix;
 use crate::ops::serial::{OperationError, OperationErrorKind};
 use crate::ops::Op;
 use crate::SparseEntryMut;
-use itertools::Itertools;
 use nalgebra::{ClosedAdd, ClosedMul, DMatrixSlice, DMatrixSliceMut, Scalar};
 use num_traits::{One, Zero};
 
@@ -33,56 +32,42 @@ pub fn spmm_cs_prealloc<T>(
 where
     T: Scalar + ClosedAdd + ClosedMul + Zero + One,
 {
+    let some_val = Zero::zero();
+    let mut scratchpad_values: Vec<T> = vec![some_val; b.pattern().minor_dim()];
+    let mut scratchpad_indices: Vec<usize> = vec![0; b.pattern().minor_dim()];
+    let mut scratchpad_used: Vec<bool> = vec![false; b.pattern().minor_dim()];
+    let mut right_end = 0usize;
     for i in 0..c.pattern().major_dim() {
         let a_lane_i = a.get_lane(i).unwrap();
 
-        let some_val = Zero::zero();
-        let mut scratchpad_values: Vec<T> = vec![some_val; b.pattern().minor_dim()];
-        let mut scratchpad_indices: HashSet<usize> = HashSet::new();
-
         let mut c_lane_i = c.get_lane_mut(i).unwrap();
-        //let (indices, values) = c_lane_i.indices_and_values_mut();
-        //indices
-        //    .iter()
-        //    .zip(values.iter())
-        //    .for_each(|(id, val)| scratchpad_values[*id] = beta.clone() * val.clone());
-
-        //for (index, c_ij) in c_lane_i.indices_and_values_mut() {
-        //    *c_ij = beta.clone() * c_ij.clone();
-        //}
 
         for (&k, a_ik) in a_lane_i.minor_indices().iter().zip(a_lane_i.values()) {
             let b_lane_k = b.get_lane(k).unwrap();
-            //let (mut c_lane_i_cols, mut c_lane_i_values) = c_lane_i.indices_and_values_mut();
             let alpha_aik = alpha.clone() * a_ik.clone();
             for (j, b_kj) in b_lane_k.minor_indices().iter().zip(b_lane_k.values()) {
                 // Determine the location in C to append the value
-                // TODO make a scratchpad and defer the accumulation into C after processing one
-                // full row of A.
                 scratchpad_values[*j] += alpha_aik.clone() * b_kj.clone();
-                scratchpad_indices.insert(*j);
-                //let (c_local_idx, _) = c_lane_i_cols
-                //    .iter()
-                //    .enumerate()
-                //    .find(|(_, c_col)| *c_col == j)
-                //    .ok_or_else(spmm_cs_unexpected_entry)?;
-
-                //c_lane_i_values[c_local_idx] += alpha_aik.clone() * b_kj.clone();
-                //c_lane_i_cols = &c_lane_i_cols[c_local_idx..];
-                //c_lane_i_values = &mut c_lane_i_values[c_local_idx..];
+                if !scratchpad_used[*j] {
+                    scratchpad_indices[right_end] = *j;
+                    right_end += 1;
+                    scratchpad_used[*j] = true;
+                }
             }
         }
         // sort the indices, and then access the relevant indices (in sorted order) from values
         // into C.
-        let sorted_indices: Vec<usize> =
-            Itertools::sorted(scratchpad_indices.into_iter()).collect();
+        scratchpad_indices[0..right_end].sort_unstable();
         c_lane_i
             .values_mut()
             .iter_mut()
-            .zip(sorted_indices.into_iter())
+            .zip(scratchpad_indices[0..right_end].iter())
             .for_each(|(output_ref, index)| {
-                *output_ref = beta.clone() * output_ref.clone() + scratchpad_values[index].clone()
+                *output_ref = beta.clone() * output_ref.clone() + scratchpad_values[*index].clone();
+                scratchpad_used[*index] = false;
+                scratchpad_values[*index] = Zero::zero();
             });
+        right_end = 0usize;
     }
 
     Ok(())


Andlon:
Hi @smr97, this looks great! The matrix algebra implementations are indeed very simplistic, so it's great to see someone spending some time on improving them.

I'm pressed for time at the moment and unfortunately won't have time to look at this before next week. Will try to get back to you then!

Andlon:
@sebcrozet: maybe you can approve the workflow in the meantime? Thanks!

smr97:
hi @Andlon, thanks for your comments! I have added the missing example file, so this should pass the build (at least) now.

I am doing research in accelerating sparse tensor algebra in general, and in that context I find this library very useful to try my ideas out in Rust. In the coming weeks, I plan to give you a parallel SPGEMM implementation as well. This is kind of my side thread, so I won't be too active, but I'll try my best.

Good to know that you folks are looking to keep improving this performance-wise as well. Do let me know if this PR has some unintended side-effects, and we can then revisit some implementation details if required.

nalgebra-sparse/example/spmm.rs
```diff
@@ -0,0 +1,25 @@
...
+use nalgebra_sparse::CsrMatrix;
+use std::time::{Duration, Instant};
+
+#[cfg(feature = "io")]
+use nalgebra_sparse::io::load_coo_from_matrix_market_file;
+fn main() {
+    #[cfg(feature = "io")]
+    {
+        let sparse_input_matrix =
+            load_coo_from_matrix_market_file::<f64, _>("./data/crankseg_1.mtx").unwrap();
```
smr97, line 11: This assumes you create a directory named `data` in the nalgebra-sparse directory, and then put the `crankseg_1` matrix in that directory. This matrix is a part of the open source SuiteSparse matrix dataset that is commonly used for performance benchmarking.

losanc:
Hi, @smr97 ,

just a few comments, obviously only testing one matrix is not enough, and I would suggest [theses matrices](https://sparse.tamu.edu/Williams/), and there are commonly used as sparse matrix benchmarks.  

Also, using a benchmark framework, e.g. criterion, should be better than manually writing an example code. I am working on the benchmark framework now, I should be able to create a PR next week.

and here is a python3 code to download the matrices, hope it might be helpful for you. just copy the code to the directory you want and run it, and it will download the matrices. (note: dense matrix is not included here)
```python
import urllib.request
import tarfile
import shutil
import os
import argparse
import time

matrix_list = {
    "pdb1HYS": "https://suitesparse-collection-website.herokuapp.com/MM/Williams/pdb1HYS.tar.gz",
    "consph": "https://suitesparse-collection-website.herokuapp.com/MM/Williams/consph.tar.gz",
    "cant": "https://suitesparse-collection-website.herokuapp.com/MM/Williams/cant.tar.gz",
    "pwtk": "https://suitesparse-collection-website.herokuapp.com/MM/Boeing/pwtk.tar.gz",
    "rma10": "https://suitesparse-collection-website.herokuapp.com/MM/Bova/rma10.tar.gz",
    "conf5_4-8x8-05": "https://suitesparse-collection-website.herokuapp.com/MM/QCD/conf5_4-8x8-05.tar.gz",
    "shipsec1": "https://suitesparse-collection-website.herokuapp.com/MM/DNVS/shipsec1.tar.gz",
    "mac_econ_fwd500": "https://suitesparse-collection-website.herokuapp.com/MM/Williams/mac_econ_fwd500.tar.gz",
    "mc2depi": "https://suitesparse-collection-website.herokuapp.com/MM/Williams/mc2depi.tar.gz",
    "cop20k_A": "https://suitesparse-collection-website.herokuapp.com/MM/Williams/cop20k_A.tar.gz",
    "scircuit": "https://suitesparse-collection-website.herokuapp.com/MM/Hamm/scircuit.tar.gz",
    "webbase-1M": "https://suitesparse-collection-website.herokuapp.com/MM/Williams/webbase-1M.tar.gz",
    "rail4284": "https://suitesparse-collection-website.herokuapp.com/MM/Mittelmann/rail4284.tar.gz"
}
tar_postfix = ".tar.gz"
matrix_postfix = '.mtx'


def download():
    for name, url in matrix_list.items():

        # Download the matrix
        print("Downloading: ", name)
        with urllib.request.urlopen(url) as response, open(name + tar_postfix, 'wb') as matrix_tarfile:
            data = response.read()
            matrix_tarfile.write(data)
        print("Finished Downloading: ", name)

        #  Extract the matrix
        print("Extracting: ", name)
        with tarfile.open(name + tar_postfix, "r:gz") as tar:
            tar.extractall()
        print("Finished Extracting: ", name)

        # Move the matrix & remove unnecessary files
        print("Cleaning Up: ", name)
        matrix_name = name + matrix_postfix
        old_path = os.path.join('.', name, matrix_name)
        new_path = os.path.join('.', matrix_name)
        shutil.move(old_path, new_path)
        shutil.rmtree(name)
        os.remove(name + tar_postfix)
        print("Finished Cleaning Up: ", name)

        # avoid being blocked from the website
        time.sleep(0.2)


def clean():
    for name in matrix_list.keys():
        filename = name + matrix_postfix
        if not os.path.isfile(filename):
            continue
        os.remove(filename)
        print("Removed: " + filename)


help_message = "you can use \'python data.py download\' or \'python data.py\' to download the matrices, \
                or \'python data.py clean\' to remove all downloaded matrices."

if __name__ == "__main__":

    parser = argparse.ArgumentParser(description='Process some integers.')
    parser.add_argument("option", nargs='?', help=help_message)
    args = parser.parse_args()
    if not args.option or args.option == "download":
        download()
    elif args.option == "clean":
        clean()
    else:
        print("unknown option")

```

smr97:
Thanks @losanc, I used your script to download all the matrices, modified the example to iterate over them and finally ran both the branches.
Here are the results (all times are in milliseconds):

```
Old performance (Added compiler optimisation flags to have a fair comparison)
Benchmark file Ok(DirEntry("./data/cant.mtx"))
SGEMM time was 1049
sum of product is 1739133575.2574668
Benchmark file Ok(DirEntry("./data/pdb1HYS.mtx"))
SGEMM time was 2454
sum of product is 0.0000000003055902197957039
Benchmark file Ok(DirEntry("./data/mac_econ_fwd500.mtx"))
SGEMM time was 166
sum of product is 1467042346453.5232
Benchmark file Ok(DirEntry("./data/scircuit.mtx"))
SGEMM time was 159
sum of product is 1495.4854200405027
Benchmark file Ok(DirEntry("./data/rma10.mtx"))
SGEMM time was 633
sum of product is 6140451084475891
Benchmark file Ok(DirEntry("./data/consph.mtx"))
SGEMM time was 1703
sum of product is 48105124791.05674
Benchmark file Ok(DirEntry("./data/pwtk.mtx"))
SGEMM time was 2146
sum of product is 37661761312690216
Benchmark file Ok(DirEntry("./data/cop20k_A.mtx"))
SGEMM time was 997
sum of product is 51734283.15793739
Benchmark file Ok(DirEntry("./data/shipsec1.mtx"))
SGEMM time was 1730
sum of product is 4607378142188466000000000000
Benchmark file Ok(DirEntry("./data/crankseg_1.mtx"))
SGEMM time was 13588
sum of product is 2407099595647278400000
Benchmark file Ok(DirEntry("./data/webbase-1M.mtx"))
SGEMM time was 735
sum of product is 1000005.0000038443


New performance numbers
Benchmark file Ok(DirEntry("./data/cant.mtx"))
SGEMM time was 653
sum of product is 1739133575.2574668
Benchmark file Ok(DirEntry("./data/pdb1HYS.mtx"))
SGEMM time was 1153
sum of product is 0.0000000003055902197957039
Benchmark file Ok(DirEntry("./data/mac_econ_fwd500.mtx"))
SGEMM time was 171
sum of product is 1467042346453.5232
Benchmark file Ok(DirEntry("./data/scircuit.mtx"))
SGEMM time was 163
sum of product is 1495.4854200405027
Benchmark file Ok(DirEntry("./data/rma10.mtx"))
SGEMM time was 360
sum of product is 6140451084475891
Benchmark file Ok(DirEntry("./data/consph.mtx"))
SGEMM time was 1008
sum of product is 48105124791.05674
Benchmark file Ok(DirEntry("./data/pwtk.mtx"))
SGEMM time was 1235
sum of product is 37661761312690216
Benchmark file Ok(DirEntry("./data/cop20k_A.mtx"))
SGEMM time was 781
sum of product is 51734283.15793739
Benchmark file Ok(DirEntry("./data/shipsec1.mtx"))
SGEMM time was 1016
sum of product is 4607378142188466000000000000
Benchmark file Ok(DirEntry("./data/crankseg_1.mtx"))
SGEMM time was 4827
sum of product is 2407099595647278400000
Benchmark file Ok(DirEntry("./data/webbase-1M.mtx"))
SGEMM time was 558
sum of product is 1000005.0000038443
```

I agree that we need a criterion bench, but this quick and dirty example is basically just to motivate whether or not the optimisations are "worth it". These are not "publishable results" by any means. It will certainly be helpful if you could give us the benchmark as a PR. In the meanwhile, I will use this to quickly check whether I'm on the right track with the changes that I'm making.

Thanks!

losanc:
Hi @smr97 ,
I just created a draft PR here, https://github.com/dimforge/nalgebra/pull/1087.  I think it should be able to test your code. 
So you need to move matrices into `nalgebra-sparse/data` folder, and run it with `cargo bench --features io spmm_prealloc`
A small difference is that I implemented as bench the function `AA^T`, I think your code is testing `AA`, though most of the matrices are square matrices. 

A small note: it does run very very slow.  And `rail4284`  is 4K x 1M matrix, and you might get out of memory issue if your result matrix is a 1M x 1M matrix, (the current implementation, the result is 4K x 4K matrix), so just be careful with that.

Andlon (commented):
The performance results are definitely encouraging! I think we can get this merged without the full benchmark suite that @losanc is working on, since I'm quite convinced that the approach taken here is a fairly obvious improvement over the current implementation. However, in the future the benchmark suite will be invaluable for analyzing and improving our performance further!

However, the behavior seems to be drastically different in the case that the pattern of `C` does not entirely match the pattern of `A * B`. See the comments for more, but the gist of it is that the old/current implementation always produces the correct result (if `pattern(C)` is a superset of `pattern(A * B)`) or returns an error (if `pattern(A * B)` has entries not present in `pattern(C)`). Please correct me if I'm wrong, but I believe your proposed implementation would produce erronous results in both these cases? I suppose ideally our test suite would have caught this - so it suggests we need to expand it with some "negative" test cases.

It seems to me that a slightly simpler version of your implementation is possible (please let me know if this is incorrect!):
```text
Compute C <- α A * B + β C, where C is an m x n matrix.

Let c = 0 be an n-dimensional vector.

for i in 1, ..., m:
    for k s.t. pattern(A)_ik != 0:
         for j s.t. pattern(B)_kj != 0:
             c[j] += A_ik * B_kj
    for j s.t. pattern(C)_ij != 0:
         C_ij <- α c[j] + β C_ij
         c[j] <- 0
```
Here we don't need to sort anything and we don't need to keep track of indices and whether we have visited an entry or not. It also works if `pattern(C)` is a superset of `pattern(A * B)`. However, we don't detect the case where there are entries in `pattern(A * B)` which are *not* in the `pattern(C)`. So if that happens we once again get silently erronous results.

How can we fix this? Here's one lightweight variation of the above algorithm:
```text
Compute C <- α A * B + β C, where C is an m x n matrix.

Let c = 0 be an n-dimensional vector.
let v = [false, false, ...] be an n-dimensional vector.

for i in 1, ..., m:
    invalid_entry = false
	
    // First mark all entries in pattern(C) that we might encounter
    // Hence if we encounter v[j] = false below, we know that
    // pattern(A * B) is *not* a subset of pattern(C)
    for j s.t. pattern(C)_ij != 0:
         v[j] = true
		
    for k s.t. pattern(A)_ik != 0:
         for j s.t. pattern(B)_kj != 0:
             c[j] += A_ik * B_kj
             // Branchless detection of errors
             invalid_entry = invalid_entry || !v[j]

    for j s.t. pattern(C)_ij != 0:
        C_ij <- α c[j] + β C_ij
        c[j] <- 0
        v[j] <- false
    
    if invalid_entry:
        return error "unexpected entry"
```

Instead of the "branchless" error detection we could also have an `if` check and an return error as soon as we run into an entry, but I suspect this might have slightly worse performance (depending on how well everything else gets optimized).

Unfortunately detecting errors here is not zero cost: it should be relatively cheap, but in the end we'll have random access to the `v` vector, which might add latency that perhaps cannot be hidden.

Curious to hear your thoughts on this, @smr97!

smr97:
Hi @Andlon, thanks for the note. I am somewhat confused by the situation you describe. While it is clear that the implementation fails in case C's pattern is not the exact same as A or B, I don't see how that is even possible. As I understand it, we want to multiply two matrices and put the result in the third, is it not? So, the shape of the product is by definition the shape of the result?!

If this is not the case, please help me understand what this operation exactly is, because I think it is not exactly an SpGEMM then. Following that, I can propose some solutions depending on whether the shape of A x B is the shape you want, or the shape of C is the shape that you really want.

Interestingly enough, the sequential case is easier to handle. Your point might get trickier for my (WIP) parallel implementation, where I was just assuming that the shapes are the same (or rather HAVE to be), and I am just constructing the pattern on the fly while co-iterating through A and B. So, the kernel that does the symbolic computation would just not be required. This makes it much easier to write a parallel iterator-based implementation.

Finally, if this is hard to describe in text, we can set up some time to talk about this. You can send a meeting invite at saurabh.raje@utah.edu.

nalgebra-sparse/src/ops/serial/cs.rs
```diff
@@ -30,30 +32,42 @@ pub fn spmm_cs_prealloc<T>(
 where
     T: Scalar + ClosedAdd + ClosedMul + Zero + One,
 {
+    let some_val = Zero::zero();
+    let mut scratchpad_values: Vec<T> = vec![some_val; b.pattern().minor_dim()];
+    let mut scratchpad_indices: Vec<usize> = vec![0; b.pattern().minor_dim()];
```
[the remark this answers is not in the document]
smr97: I agree, this is not really Rust-ic. I was more or less copying from C code (to try to match performance with it), and was somewhat paranoid about Vector implementation in Rust. Didn't want to have any performance artifacts of pushing into the vector. Since then, I read the documentation of the vector and it seems the with_capacity should avoid any variable costs.

Will change this up.

Andlon:
@smr97: it's possible because a user might make a mistake :) For example, let's say we have the following case:

- a user wants to compute N products `C_i = A_i * B_i`. The entries of `A_i` and `B_i` vary for each `i`, but the pattern is always the same.
- Therefore, they precompute the pattern with `spmm_csr_pattern`.
- then they compute, in a loop, each product `C_i`, but reusing the precomputed pattern.

Now, say that the program is somewhat complex, and somehow someone makes a mistake by accidentally giving in the wrong sparsity pattern for `C_i`. The current implementation would then give an error when computing `C_i <- A_i * B_i`, since `C_i` does not have the correct pattern.

It is however debatable whether this is something we should do. An alternative could also be to provide a separate `_unchecked` method which performs no checks in this direction (certainly this would be useful for the `Mul` implementation).

smr97:
Okay, now I get it. The unchecked idea sounds good! Overall, the cost of checking should actually be the same as cost of creating the pattern in the first place. We could have the check while running the kernel, but I think those branches should slow down the code especially for smaller matrices.
As for the two directions of error, in this context it seems like if `pattern(C)` is a superset of `pattern(A*B)`, it should be safe? There will just be a superfluous set of computations, but the results should be mathematically correct.

Let me know what you prefer. Should I try to make a different path for unchecked mul and have this kernel there?

Andlon:
@smr97: Regardless of the error checking question, what do you think of my [first proposal](https://github.com/dimforge/nalgebra/pull/1081#pullrequestreview-898898727)? At first glance, it seems both simpler and possibly even a bit faster than your initial implementation (no sort needed, for example) - but perhaps I've missed a crucial detail that would become evident if one were to try to implement it? Would be curious to hear what you think about it!

smr97:
@Andlon, I thought about it, and your suggestion makes perfect sense. I am still operating in the assumption that all the patterns match perfectly. Therefore, the indices I was storing in the previous implementation were already stored in `C.pattern(i)`. It made no sense to compute (store+sort) them again.

Thanks a ton for pointing this out. Here are the new performance numbers (TL;DR it got faster!):


```
Benchmark file "data/cant.mtx"                                                                                                                          
SGEMM time was 481                                                                                                                                      
sum of product is 1739133575.2574668                                                                                                                    
Benchmark file "data/pdb1HYS.mtx"                                                                                                                       
SGEMM time was 821                                                                                                                                      
sum of product is 0.0000000003055902197957039                                                                                                           
Benchmark file "data/mac_econ_fwd500.mtx"                                                                                                               
SGEMM time was 116                                                                                                                                      
sum of product is 1467042346453.5232                                                                                                                    
Benchmark file "data/scircuit.mtx"                                                                                                                      
SGEMM time was 110                                                                                                                                      
sum of product is 1495.4854200405027                                                                                                                    
Benchmark file "data/rma10.mtx"                                                                                                                         
SGEMM time was 262                                                                                                                                      
sum of product is 6140451084475891                                                                                                                      
Benchmark file "data/consph.mtx"                                                                                                                        
SGEMM time was 753                                                                                                                                      
sum of product is 48105124791.05674                                                                                                                     
Benchmark file "data/pwtk.mtx"                                                                                                                          
SGEMM time was 952                                                                                                                                      
sum of product is 37661761312690216                                                                                                                     
Benchmark file "data/cop20k_A.mtx"                                                                                                                      
SGEMM time was 533                                                                                                                                      
sum of product is 51734283.15793739                                                                                                                     
Benchmark file "data/shipsec1.mtx"                                                                                                                      
SGEMM time was 744                                                                                                                                      
sum of product is 4607378142188466000000000000                                                                                                          
Benchmark file "data/crankseg_1.mtx"                                                                                                                    
SGEMM time was 3501                                                                                                                                     
sum of product is 2407099595647278400000                                                                                                                
Benchmark file "data/webbase-1M.mtx"                                                                                                                    
SGEMM time was 471                                                                                                                                      
sum of product is 1000005.0000038443
```

Let me know how you prefer to handle the previous issue of incorrect patterns for `A * B` and `C`

Andlon:
@smr97: that's great!

First, let me give some background.

My experience with scientific computing libraries is that they tend to be brittle: they have often subtle invariants that can easily be broken by accident. I've personally spent weeks debugging bugs arising from issues like this.

Part of my design philosphy for `nalgebra-sparse` is that it should be easy to use by default. This means that the APIs generally try to protect you from subtle mistakes that could give you incorrect results, possibly at the cost of some performance. I think the present issue is an excellent example of this: without checking for an erronous pattern, a user might get completely wrong results without at all realizing. This is unfortunate and I'd like to prevent this as long as it can be reasonably done.

At the same time, another aspect of the design philosophy is that you should be able to *opt out* of runtime checks where the additional checks cause significant performance overhead.

With this in mind, I would suggest the following course of action:

1. Duplicate the current fastest implementation that you have as `spmm_cs_prealloc_unchecked`.
2. Add the checks [that I proposed](https://github.com/dimforge/nalgebra/pull/1081#pullrequestreview-898898727) to `spmm_cs_prealloc`. Perhaps you also have some ideas for further improvement?
	1. Maybe consider using [bitvec](https://docs.rs/bitvec/latest/bitvec/) instead of `Vec<bool>` as this may perhaps even further reduce the cost of error checking? Would be interesting to first try with `Vec<bool>` and then later swap with `bitvec` to see if it makes a difference.
3. Benchmark both implementations: is the overhead of checking for an erronous pattern signifcant? If so, perhaps we need to keep the `_unsafe` method, otherwise maybe we can just simply always check for an erronous pattern.

I realize this might take a little bit of effort though. What do you think?

Andlon:
@smr97: that sounds like a good path forward! In fact it's what I had in mind (but failed to communicate) when I proposed the `_unchecked` variant. The `Mul` implementation is an excellent candidate for this since we directly construct the pattern directly before calling the kernel, as you point out.

Note: as far as I can tell, it `unchecked` variant is not *unsafe*, i.e. I don't think it can break any of Rust's safety invariants.

The only downside I can see to adding a separate `_unchecked` function is the need for additional tests - but I think we can get away with a single `proptest` that tests that the results of the checked and unchecked variants are the same (exactly so for integer matrices) for arbitrary matrices of compatible dimensions. I can help you with this if you're not familiar with `proptest`.

As for parallelization: assuming the pattern is already available, it seems to me that parallelization would be relatively "simple" if we had parallel mutable iterators over CSR matrix rows, since then we could use the same strategy as for the single-threaded case, since we're already handling one row at a time. Each thread would needs its own scratchpad, however, which means that memory use increases rather significantly with number of threads... (I wonder if a `FxHashMap` could be fast enough to serve as a replacement scratchpad).

At least to me it seems easier to separately parallelize pattern construction and arithmetic, but it seems you have a different strategy (which is OK! We can offer several types of computational kernels in the API).

Also, please feel free to join us on #nalgebra on the [Dimforge Discord server](https://discord.gg/vt9DJSW) if you want to chat about the implementation/algorithm (I'm very busy these days so I prefer async communication).

Andlon:
Oh, and one more thing: it might be worthwhile to check if using unsafe unchecked indexing for the random accesses in your current implementation might speed things up further. Since Rust cannot know that the indices we're looking up are in bounds we are probably incurring bounds checks on all accesses, which may or may not incur a significant performance penalty.

Andlon:
It just struck me that *if* `FxHashMap` is fast enough to use as a scratchpad, it would also neatly solve our invalid pattern problem. Consider the following modification of the present algorithm:

```text
Compute C <- α A * B + β C, where C is an m x n matrix.

Let c: usize -> T be an empty hash map.

for i in 1, ..., m:
    for k s.t. pattern(A)_ik != 0:
        for j s.t. pattern(B)_kj != 0:
             // Of course, need to initialize if it does not yet exist
             c[j] += A_ik * B_kj
    for j s.t. pattern(C)_ij != 0:
        C_ij <- α c.remove(j).unwrap_or(0) + β C_ij
        c[j] <- 0
    if !c.is_empty():
        // If the hash map is not empty, then it means that
        // pattern(C) is *not* a superset of pattern(A * B)
        return error
```
Inserting/looking up in a hash map is generally far more expensive than modifying/looking up in an array. *However*, the hash map is generally "small" and is basically almost guaranteed to be in cache (unless the row has an enormous amount of entries), whereas the big array scratchpad relies on random memory access in a potentially big array where entries are not all in cache for big matrices.

Hash maps also have one more advantage over scratchpad arrays: If your CSR matrix is `mxn` with `n >> m`, then the scratchpad array might require an enormous amount of memory, much more so than the matrix itself, if each row only has a relatively small number of entries.

Worth thinking about maybe?

smr97:
@Andlon , as we discussed, the Hashmap somehow was significantly slower. Thanks for the pointer to unchecked indexing, I have implemented it in the kernel. The performance did get better by about 10%.
Also, in my previous post, I used `unsafe` incorrectly. I meant to say `unchecked`. The faster kernel now however, is also `unsafe` due to the indexing :laughing: .
I did also refactor some transpose calls which would be common across the checked and unchecked kernels.
Finally, I linked the macros implementing the multiply trait to call the unchecked variant always.

Let me know what you think? Hopefully this is good to merge now?

nalgebra-sparse/src/ops/serial/cs.rs
```diff
@@ -20,7 +20,50 @@ fn spmm_cs_unexpected_entry() -> OperationError {
...
+                *output_ref = beta.clone() * output_ref.clone()
+                    + scratchpad_values.get_unchecked(*index).clone();
+                *scratchpad_values.get_unchecked_mut(*index) = Zero::zero();
+            });
+    }
+
+    Ok(())
+}
+
+pub fn spmm_cs_prealloc_checked<T>(
```
Andlon, line 66: Let's keep the original name here. "checked" is implied by not being "unchecked", and we don't want a breaking change on top!

nalgebra-sparse/src/ops/serial/csc.rs
```diff
@@ -83,35 +85,81 @@ where
...
+    match (&a, &b) {
+        (NoOp(ref a), NoOp(ref b)) => {
+            // Note: We have to reverse the order for CSC matrices
+            spmm_cs_prealloc_unchecked(beta, &mut c.cs, alpha, &b.cs, &a.cs)
         }
+        _ => do_transposes(beta, c, alpha, a, b, spmm_csc_prealloc_unchecked),
     }
 }
 
+fn do_transposes<T, F>(
```
Andlon, line 127: Maybe rename to something like `spmm_csc_transposed` so that it's a bit clearer what this function does

Same for the CSR variant.

nalgebra-sparse/src/ops/serial/csc.rs
```diff
@@ -83,35 +85,81 @@ where
...
+    beta: T,
+    c: &mut CscMatrix<T>,
+    alpha: T,
+    a: Op<&CscMatrix<T>>,
+    b: Op<&CscMatrix<T>>,
+    caller: F,
```
Andlon, line 133: Maybe rename `caller` to something like `spmm_kernel` to make it clearer what it is

nalgebra-sparse/Cargo.toml
```diff
@@ -24,20 +24,33 @@ io      = [ "pest", "pest_derive" ]
...
 [dependencies]
+criterion = { version = "0.3", features = ["html_reports"] }
 nalgebra = { version="0.30", path = "../" }
 num-traits = { version = "0.2", default-features = false }
 proptest = { version = "1.0", optional = true }
 matrixcompare-core = { version = "0.1.0", optional = true }
 pest           = { version = "2", optional = true }
 pest_derive    = { version = "2", optional = true }
 serde = { version = "1.0", default-features = false, features = [ "derive" ], optional = true }
+itertools = "0.10"
```
Andlon, line 35: Is this used anywhere? I can't find any new import of `itertools`.

nalgebra-sparse/example/spmm.rs
```diff
@@ -0,0 +1,36 @@
+extern crate nalgebra_sparse;
```
Andlon, line 1: I think we should remove this whole example before merging, because it's not an actual *example* that shows how to use the library, it's a benchmark. And I guess since @losanc is working on benchmarks we'll eventually get a nice benchmark suite that will fill this role.

nalgebra-sparse/Cargo.toml
```diff
@@ -24,20 +24,33 @@ io      = [ "pest", "pest_derive" ]
...
+
 [package.metadata.docs.rs]
 # Enable certain features when building docs for docs.rs
 features = [ "proptest-support", "compare" ]
+
+[profile.release]
+opt-level = 3 
+lto = "fat"
+codegen-units = 1
+panic = "abort"
```
Andlon, line 56: I think we should remove the optimization flags here (and *especially* `panic = abort`!).

nalgebra-sparse/Cargo.toml
```diff
@@ -24,20 +24,33 @@ io      = [ "pest", "pest_derive" ]
 slow-tests = []
 
 [dependencies]
+criterion = { version = "0.3", features = ["html_reports"] }
```
Andlon, line 27: Can remove `criterion` if we're removing the example using it.

Andlon (changes requested):
Thanks @smr97, that's great! Other than some minor things I think we should address the following:

- remove the example and the extra dependencies (see code comments for reasoning)
- we should still test `spmm_cs*_prealloc_unchecked`, even though it's (partially) tested

smr97:
@Andlon I have made changes to address all review comments above. 

>  You can use the same setup, but verify the results against `spmm_cs*_prealloc` instead of against a dense matrix multiplication like those existing tests do (IIRC). What do you think?

Here I feel it is easier and more robust to just keep comparing against the Dense version? The test cases are generated using the same function in any case. Do you think there is a specific reason for not comparing against the dense matrix and comparing the two implementations?

Andlon:
> @Andlon I have made changes to address all review comments above.
> 
> > You can use the same setup, but verify the results against `spmm_cs*_prealloc` instead of against a dense matrix multiplication like those existing tests do (IIRC). What do you think?
> 
> Here I feel it is easier and more robust to just keep comparing against the Dense version? The test cases are generated using the same function in any case. Do you think there is a specific reason for not comparing against the dense matrix and comparing the two implementations?

Sorry, I thought I had responded to this - I agree that just copying the tests and comparing against the dense versions is just as reasonable.

nalgebra-sparse/src/ops/serial/cs.rs
```diff
@@ -20,6 +20,49 @@ fn spmm_cs_unexpected_entry() -> OperationError {
...
+        //Get indices from C pattern and gather from the dense scratchpad_values
+        let (indices, values) = c_lane_i.indices_and_values_mut();
+        values
+            .iter_mut()
+            .zip(indices)
+            .for_each(|(output_ref, index)| unsafe {
+                *output_ref = beta.clone() * output_ref.clone()
+                    + scratchpad_values.get_unchecked(*index).clone();
+                *scratchpad_values.get_unchecked_mut(*index) = Zero::zero();
+            });
```
Andlon, line 60: As far as I can tell, this will lead to undefined behavior if `c`, `a` and `b` do not have the right dimensions. For example, if `c` has `10` columns, but `b` only has `5`, then we might try to index into element `8` in the scratchpad vector (which will have dimension `5`.

The dimensions are currently checked in the caller methods `spmm_csr/csc_prealloc`, but since we're using `unsafe` here we should additionally check the invariants here in this method so that the method is generally sound to call. In other words: since the method is not marked `unsafe`, we can not rely on the method being called with valid arguments, since unsafe code cannot rely on safe code being correct.

Andlon (changes requested):
Sorry for taking so long with the review @smr97. Looks great to me, excited to get the performance boost in the library! I only noticed one minor but very important detail concerning `unsafe` and soundness, please see my comment.

nalgebra-sparse/src/ops/serial/cs.rs
smr97: Makes sense, I think the latest commit should handle this.

Andlon:
Hi @smr97, the PR looks good to me now, but I see that some of the CI tests are failing. I only took a brief look and I couldn't quite tell why - do you have some thoughts? I'm not able to dig into it right now, so if you already have an idea what it might be that would be great!

smr97:
> Hi @smr97, the PR looks good to me now, but I see that some of the CI tests are failing. I only took a brief look and I couldn't quite tell why - do you have some thoughts? I'm not able to dig into it right now, so if you already have an idea what it might be that would be great!

@Andlon 
I think it was just a dependency in Cargo.toml that I didn't have. I merged with master, and re-ran the workflow on my fork using your yaml file. Seems to be passing now.

Andlon:
@smr97: it looks like there's a superfluous CI config file `.github/workflows/rust.yaml`? I assume this should not be part of the PR...?

smr97:
@Andlon, yes, I just wanted to see if the tests pass so I added that for my workflow. Removed it now.

Andlon (approved):
Sorry for the time it took to approve this @smr97, I've had some extraordinarily busy weeks. I will try to be more responsive in the future.

Thank you so much for your work here - it's a very welcome performance boost to matrix multiplications!

smr97:
Thanks @Andlon, I am happy that this useful library is now a bit faster!

Issue #983, opened 2021-09-07
Umpredictable unstable (buggy?) SVD evaluation


Nekrolm:
Hi! I have faced with strange behavior of `Matrix::svd` method:

Largest singular value somehow significantly depends on `compute_u`/`compute_v` params. And transposition can hide/expose this effect.

There is an example (nalgebra = 0.28, also reproducible for 0.29)
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d96e388cd6e8aa7f437451d44960b9fe

```
    let svd1 = m.svd(true, true);
    println!("{:?}", svd1.singular_values);
    let svd2 = m.svd(false, false);
    println!("{:?}", svd2.singular_values);
    let svd3 = m.svd(false, true);
    println!("{:?}", svd3.singular_values);
    let svd4 = m.svd(true, false);
    println!("{:?}", svd4.singular_values);
```
prints different values for same matrix:
```
Matrix { data: [[0.0000000000000034441735216135435, 46.56197615440326, 22.38119784648299]] }
Matrix { data: [[0.000000000000002338832028974828, 31.618802153516995, 22.38119784648299]] }
Matrix { data: [[0.0000000000000020737891983235553, 28.035673172576868, 22.38119784648299]] }
Matrix { data: [[0.0000000000000034441735216135435, 46.56197615440326, 22.38119784648299]] }
```

It does not look like numerical error.

Nekrolm:
result from [`numpy`](https://colab.research.google.com/drive/1TjX1NQcSfGCimmOmlNmgliVlTOe4Cl8b?usp=sharing)

```
A = np.array([[10.74785316637712, -5.994983325167452, -6.064492921857296], 
              [-4.149751381521569, 20.654504205822462, -4.470436210703133], 
              [-22.772715014220207, -1.4554372570788008, 18.108113992170573]])
U, S, Vt = np.linalg.svd(A)
print(S)
```
```
[3.16188022e+01 2.23811978e+01 3.31763547e-15]
```

cfunky:
The particular case of 3x3 statically allocated matrices seems to have been fixed, probably with release 0.30.0. For dynamically allocated matrices the problem still persists.

Andlon:
Fixed by #1089.

Issue #1072, opened 2022-01-28
SVD produces wrong singular values


cfunky:
The SVD with the default tolerance parameter `eps` can produce significantly wrong singular values. This behavior occurs in nalgebra version 0.30.1, potentially also in other versions (I did not try this). It looks like the problem occurs when, as in the example above, an singular value is close to zero and the tolerance parameter `eps` of `Matrix::try_svd(...)` or `Matrix::try_svd_unordered(...)` is set to a very low value, as is done internally, e.g., in `Matrix::svd(...)`. Here's a minimal example reproducing this issue:

```rust
use nalgebra::{dmatrix, dvector};

fn main() {
    let x = dmatrix![-6.206610118536945f64, -3.67612186839874; -1.2755730783423473, 6.047238193479124];
    // let mut x_svd = x.try_svd(true, true, 1e-8, 0).unwrap();// This gives almost the same result as the line below.
    let mut x_svd = x.svd(true, true); // This gives almost the same result as the line above.
    println!("{}", x_svd.singular_values);
    x_svd.singular_values = dvector![1.0, 0.0];
    println!("{}", x_svd.singular_values);
    let y = x_svd.recompose().unwrap();
    let y_svd = y.svd(true, true); // This produces wrong singular values!
    // let y_svd = y.try_svd(true, true, 1e-8, 0).unwrap(); // This produces correct singular values! For small enough `eps` the wrong behavior occurs.
    println!("{}", y_svd.singular_values);
}
```
[Link to playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5d8c493a911dfeb9c2990cdb253d8e33)

The above code prints
```
  ┌                   ┐
  │  7.81117060812432 │
  │ 5.405337300528416 │
  └                   ┘



  ┌   ┐
  │ 1 │
  │ 0 │
  └   ┘



  ┌                                    ┐
  │                 1.1503016006828526 │
  │ 0.00000000000000015415627596860353 │
  └                                    ┘
```

Andlon:
Thanks for reporting! Ugh, that's pretty bad. Here's a modified version which also compiles for older versions of nalgebra:

```rust
use nalgebra::{DMatrix, DVector};

fn main() {
    let x = DMatrix::from_row_slice(2, 2, &[-6.206610118536945f64, -3.67612186839874, -1.2755730783423473, 6.047238193479124]);
[...]

PR #1089, opened 2022-03-09
Fix SVD instability bug

YuhanLiin:
Fixes the SVD bug in #983 and introduces a new test covering it.

### DIFF
--- a/src/linalg/svd.rs
+++ b/src/linalg/svd.rs
@@ -2,7 +2,6 @@
 use serde::{Deserialize, Serialize};
 use std::any::TypeId;
 
-use approx::AbsDiffEq;
 use num::{One, Zero};
 
 use crate::allocator::Allocator;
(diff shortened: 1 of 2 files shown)


tests/linalg/svd.rs
```diff
@@ -460,3 +460,27 @@ fn svd_sorted() {
         epsilon = 1.0e-5
     );
 }
+
+#[test]
+// Exercises bug reported in issue #983 of nalgebra
+fn svd_consistent() {
```
Andlon, line 466: maybe rename to `svd_regression_issue_983` for clarity? Adding a link to the issue might also be nice :)

Andlon (commented):
Thanks for looking into this! At first glance this looks great, added a few minor comments.

I think your change of epsilon might fix the example in #1072 as well. Would you be willing to add a regression test for the example in that issue as well? That would be great!

Andlon:
@sebcrozet: would be great if you could approve the CI :-)

Andlon:
@sebcrozet: this looks good to me, but I'm not intimately familiar with the SVD algorithm, so would be great if you could take a closer look as well when you have time.

src/linalg/svd.rs
```diff
@@ -888,13 +888,13 @@ fn compute_2x2_uptrig_svd<T: RealField>(
...
-            let (csu, sgn_u) = GivensRotation::new(cu, su);
+        let cu = (m11.scale(csv.c()) + m12 * csv.s()) / v1.clone();
+        let su = (m22 * csv.s()) / v1.clone();
+        let (csu, sgn_u) = GivensRotation::new(cu, su);
+        v1 *= sgn_u.clone();
+        v2 *= sgn_u;
 
-            v1 *= sgn_u.clone();
-            v2 *= sgn_u;
+        if compute_u {
```
sebcrozet, line 897: Why was this change needed?

sebcrozet:
Looks good to me, thanks!

PR #1095, opened 2022-03-22
improve rkyv support

zyansheep:
upgrade rkyv to 0.7
update resolve functions
add bytecheck impls if not #[no-std]


sebcrozet:
Thank you for this PR! I took the liberty of simplifying the implementations by relying on derives instead of manual impls.

PR #1096, opened 2022-03-22
Cleanup examples and doc links

sterlingjensen:
Close example code fences and normalize containing head line in touched files.
Remove stale reference to `slice_assume_init` (commit 8c6ebf27), fix long dead internal links in deprecation notices.


sebcrozet:
Thank you for this PR, that looks great!
I think you just need to run a `cargo fmt` and this will be good to merge.

PR #1094, opened 2022-03-21
fix for `UnitComplex::slerp()` #1093

jsmith628:
Potential fix for #1093

Issue #1093, opened 2022-03-21
Error in `slerp()` formula for `UnitComplex`?


jsmith628:
So I was poking through the code for rotational interpolation, and I saw the definition for `slerp()` for `UnitComplex` was
```rust
pub fn slerp(&self, other: &Self, t: T) -> Self {
        Self::new(self.angle() * (T::one() - t.clone()) + other.angle() * t)
}
```
However, this causes some really weird behavior with certain pairs of complex numbers where they will rotate the long way around.

For instance, if you `slerp()` from e^(-3pi/4) to e^(3pi/4), this code will interpolate the angle *counter clockwise*  from -135° to 135° even though going clockwise is the shorter path.
```rust

    use std::f64::consts::PI;

    let c1 = UnitComplex::new( -135.0f64.to_radians() );
    let c2 = UnitComplex::new( 135.0f64.to_radians() );

    // left  = Complex { re: 1.0, im: 0.0 }
    // right = Complex { re: -1.0, im: 1.2246467991473532e-16 }
    assert_relative_eq!(c1.slerp(&c2, 0.5), UnitComplex::new(PI), epsilon=2e-10);
```

Of course, do say so if this is intended, but I would imagine going clockwise here would be the more natural choice?

Issue #908, opened 2021-06-09
Slerp / lerp between 4d rotations?


iMplode-nZ:
I would like to do some sort of interpolation between 4d rotations, I don't really care what type it is though.

Issue #657, opened 2019-09-25
Slerp special cases


Diggsey:
## Colinear, opposite direction inputs

> Returns `None` if the two vectors are almost collinear and with opposite direction
    /// (in this case, there is an infinity of possible results).

This is only true for the 3D slerp, for 2D slerp there are only two possibilities. In either case, it might be more useful to return *some* result, rather than `None`. At the moment, the non-try variant will just return the original rotation which is quite surprising. IMO it would be better to return any rotation which is a valid solution rather than the original input.

You could have a slerp variation which returns `Result<Rotation, RotationSet>` where `RotationSet` would be a pair of rotations in 2D, and an implicitly defined "circle" of rotations in 3D, from which the user could pull out a single rotation by specifying a preferred solution in some way. The normal "slerp" function could call this function and then use an arbitrary method of selection.

## Colinear, very close

At the moment, there is a special case for when they are exactly aligned, but there's no special case for when they are just very close - the Wikipedia article on SLERP suggests falling back to linear interpolation when the inputs are very close together. Do you think this is worthwhile?

PR #1100, opened 2022-03-30
Generalized `slerp()` and `powf()` for `Rotation<T,N>`

jsmith628:
Implements `Rotation::slerp()` and `Rotation::powf()` for all static dimensions.

Solves #908. Potentially interacts with #657, #1093, and #1094

The algorithm I used for `powf()` works by first computing the Real Schur form of the backing matrix to get it in a block-diagonal form, and then interpolating each 2x2 block as if it were a 2D rotation. The `slerp()` implementation then interpolates using `(R2/R1)^t * R1`. If the dimension is 2 or 3 though, it falls back on the old implementation using complex numbers and Quaternions.

My first consideration was to instead compute `powf()` by first taking the logarithm, multiplying by the exponent, and then taking the exponential. The issue with that was that in order to compute the log, we'd most likely have to either take a sequence of square roots using newton's method (which can get quite unstable when starting near a 180degree rotation) or do a Schur based algorithm. So since we'd be computing the form anyway, I opted to just do both the exponential and logarithm steps in one go with a Schur decomposition.

In order to do all of this, I obviously had to extend the API a bit, but _for the most part_, it should just be drop in place for most users. The implementation does add a fair number of extra trait bounds for the dimension so it could use the Schur decomposition, but in the 2D and 3D cases, those should just resolve down and not affect any code using the old versions. **However**, I _did_ have to force `T` to be a `RealField`, which is an added restriction from the old 2D case, which previously only required `SimdRealField`. Unless it's decided the generalized pow and slerp should be separate from the old implementations, I don't really know how else to unify them without this problem happening.

Lastly, from my experience working with Geometric algebra, I *do* know of an additional algorithm that works in the special case of 4D and 5D. I haven't implemented it here yet, but if there's enough desire for it, I could implement it and run some benchmarks to see if it should be added as a specialization for sake of optimization.

### DIFF
--- a/src/geometry/rotation_interpolation.rs
+++ b/src/geometry/rotation_interpolation.rs
@@ -140,7 +140,8 @@ impl<T:RealField, const D: usize> Rotation<T,D> where
                 transmute::<&Rotation3<T>,&Self>(&self3d.slerp_3d(other3d, t)).clone()
             },
 
-            _ => self * (self/other).powf(t)
+            //the multiplication order matters here
+            _ => (other/self).powf(t) * self
         }
 
     }
--- a/src/geometry/rotation_interpolation.rs
+++ b/src/geometry/rotation_interpolation.rs
@@ -118,7 +118,8 @@ impl<T:RealField, const D: usize> Rotation<T,D> where
     pub fn general_pow(self, t:T) -> Self {
         if D<=1 { return self; }
 
-        // println!("r:{}", self);
+        println!("r:{}", self);
+        println!("{}", self.clone().into_inner().hessenberg().unpack_h());
 
         //taking the (real) schur form is guaranteed to produce a block-diagonal matrix
         //where each block is either a 1 (if there's no rotation in that axis) or a 2x2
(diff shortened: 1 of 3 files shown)


src/geometry/rotation_interpolation.rs
```diff
@@ -79,3 +82,67 @@ impl<T: SimdRealField> Rotation3<T> {
...
+
+            //FIXME: this doesn't really work in 1D since we can't interp between -1 and 1
+            1 => self.clone(),
+
+            //NOTE: Not pretty, but without refactoring the API, this is the best we can do
+            //NOTE: This is safe because we directly check the dimension first
+            2 => unsafe {
+                let (self2d, other2d) = (
+                    transmute::<&Self, &Rotation2<T>>(self),
+                    transmute::<&Self, &Rotation2<T>>(other),
```
tpdickso, line 132: Is this sound? It seems like a reasonable operation in theory, just imagining the references as pointers-to-chunks-of-data, but I'm not confident enough in Rust's type aliasing rules to know if this violates some sort of compiler invariant.

Ralith: Yes, this is sound. Because this branch can only be dynamically reached when `Self` is `Rotaiton2<T>`, this is really just an identity operation. Leaving the lifetime unspecified is sketchy (this is probably transmuting to `&'static` implicitly), but still strictly sound since no dynamic access outlives the underlying borrow.

For easier auditing and more obvious correctness, extract the transmute into a safe helper function like:

```rs
fn identity_cast<T, D1, D2>(x: &Rotation<T, D1>) -> &Rotation<T, D2> {
    assert_eq!(D1, D2);
    mem::transmute(x)
}
```

jsmith628: Well, I was _about_ to say that I tweaked the function to use `fixed_resize()` instead in order to remove the `transmute()`s, but I see what you're saying @Ralith

I'm good with either variant honestly. If we want to keep the code safe, I feel like the compiler _should_ be able to inline all of the copies. Otherwise, I'm down to revert back to the transmutes and add this signposting function if we want the performance guaranteed.

src/geometry/rotation_interpolation.rs
```diff
@@ -79,3 +82,67 @@ impl<T: SimdRealField> Rotation3<T> {
...
+    ///
+    /// let q1 = Rotation3::from_euler_angles(std::f32::consts::FRAC_PI_4, 0.0, 0.0);
+    /// let q2 = Rotation3::from_euler_angles(-std::f32::consts::PI, 0.0, 0.0);
+    ///
+    /// let q = q1.slerp(&q2, 1.0 / 3.0);
+    ///
+    /// assert_eq!(q.euler_angles(), (std::f32::consts::FRAC_PI_2, 0.0, 0.0));
+    /// ```
+    ///
+    //FIXME: merging slerp for Rotation2 and Rotation3 into this raises the trait bounds
```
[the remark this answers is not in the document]
jsmith628: That is correct. This was basically me leaving the final API choice to later; it's not great to change that without input from everyone else.

Obviously, it'd be nice to just have one `slerp()` function that gets specialized for all of them, but if changing the bounds is going to send some people into generic hell, we can also just accept it and change the names.

I don't know how often people try to use `Complex` with a SIMD scalar though? I can't imagine it's really that often, but I suppose the real issue is that this will propagate to all of the dependent trait bounds.

src/geometry/rotation_interpolation.rs
Ralith: I think a purely safe version is preferable unless a specific performance problem has been identified.

Issue #1078, opened 2022-02-15
`from_matrix` returns the identity on a rotation matrix


cpg314:
Consider the following simple program ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bd92af4961f4c660020c2958e9151045)), creating a rotation matrix, and creating a rotation from it.
```rust
use nalgebra; // 0.30.1

fn main() {
    let mat = nalgebra::Matrix3::<f64>::new(0.0, 0.0, 1.0, 0.0, -1.0, 0.0, 1.0, 0.0, 0.0);
    println!("{}", mat);

    // A. Returns the rotation matrix
    let mat2 = nalgebra::Rotation3::from_matrix_unchecked(mat.clone());
    println!("{}", mat2);

    // B. Returns identity
    let mat2 = nalgebra::Rotation3::from_matrix(&mat);
    println!("{}", mat2);

    // C. from_matrix_eps, starting at the identity matrix (equivalent to B.)
    // Returns identity
    let mat2 =
        nalgebra::Rotation3::from_matrix_eps(&mat, f64::EPSILON, 0, nalgebra::Rotation3::identity());
    println!("{}", mat2);

    // D. from_matrix_eps, starting away from the identity matrix
    // Returns the rotation matrix
    let rot = nalgebra::UnitQuaternion::<f64>::from_euler_angles(0.1, 0.2, 0.3);
    let mat2 =
        nalgebra::Rotation3::from_matrix_eps(&mat, f64::EPSILON, 0, rot.to_rotation_matrix());
    println!("{}", mat2);
}
```
- The use of `from_matrix` (calling `from_matrix_eps` behind the scenes) return the identity matrix instead of the original rotation matrix (B, C)
- The problem disappears when starting `from_matrix_eps` from a non-identity matrix (A, D).

Looking at the [implementation](https://github.com/dimforge/nalgebra/blob/319597e137f4a9a0f82f6570ec2964aa15c5c80d/src/geometry/rotation_specialization.rs#L728) of `SimdRealField::from_matrix_eps`, it exits after one iteration: the `axis` is the zero vector, `axisangle` is zero. Indeed, the three components of the sum
```rust
let axis = rot.column(0).cross(&m.column(0))
                + rot.column(1).cross(&m.column(1))
                + rot.column(2).cross(&m.column(2));
```
are (0,-1,0), 0 and (0,1,0).

Is this a bug or the expected behavior?

See also #661 #627

[...]

Issue #627, opened 2019-08-05
Behavior of Rotation3::from_matrix(&M) when M is a diagonal matrix 


krzesi-mir:
Hello,

While tracking a bug in our code I noticed that `Rotation3::from_matrix` does not behave the way I expected. In particular, for any 3x3 diagonal matrix `M`, `Rotation3::from_matrix(&M)` returns the identity rotation. For example if `M` is
```
[-1,  0, 0,
  0, -1, 0,
  0,  0, 1],
```
which already is a rotation matrix, then `Rotation3::from_matrix(&M)` returns
```
Rotation {
    matrix: Matrix {
        data: [
            1.0,
            0.0,
            0.0,
            0.0,
            1.0,
            0.0,
            0.0,
            0.0,
            1.0,
        ],
    },
}.
```

After the reading the source code and the paper this function implements, I understand why this is happening, but I was wondering if this is the desired behavior of`from_matrix`. If yes, then I'd expand the documentation of this function a little, in particular I'd mention that for some rotation matrices `M` (like the one above), `Rotation3::from_matrix` might return a rotation which is different than the one described by `M`. Perhaps it's worth mentioning the `Rotation::from_matrix_unchecked` method here too. If this is not the desired behavior then maybe, as suggested by the authors of the paper, randomly perturbating the `rot` matrix to escape the local maximum would do the trick (it could be computationally costly, though). 

I think it would be cool if `Rotation3::from_matrix` checked whether the input matrix `M` already is a rotation matrix and, if it is, just return `Rotation::from_matrix_unchecked(M.clone())`.

Thanks!
Kamil

timethy:
This is also fixed with #1101

PR #1101, opened 2022-03-31
Fix from_matrix when argument is already a valid rotation matrix

timethy:
One possibility of fixing issues #627, #1078 by starting from a random rotation.

I'm not too familiar with how this whole Standard: Distribution<Rotation3<T>> works, so feel free to change to something better.

Of course, one downside of this approach is that in the very rare case of running into the bug (if one gets really unlucky), it might disappear when you re-run it.

Hence, another option would be (if one does not always want to start from a new random rotation) to fix the "random" rotation.

An idea could also be to only do this extra check if the output happens to match the first guess (up to ε), and then start from the next "deterministic" random rotation.

Alternatively, one could check if the determinant of the provided matrix is 1, and then use from_matrix_unchecked, however, this would not fix the case, when the input is a scaled "wrong" rotation matrix such as [2, 0, 0, 0, -2, 0, 0, 0, -2]. (This was suggested in https://github.com/dimforge/nalgebra/issues/627)

PS: The suggested method in the paper, of trying different initializations and checking if the frobenius norm changes, only works for non-orthonormal matrices, as all rotation matrices have the same Frobenius norm of √3. One would have to combine it with the latter approach to first detect orientation matrices.

### DIFF
--- a/src/geometry/quaternion_construction.rs
+++ b/src/geometry/quaternion_construction.rs
@@ -413,8 +413,7 @@ where
     #[cfg(feature = "rand-no-std")]
     pub fn from_matrix(m: &Matrix3<T>) -> Self
     where
-        T: RealField + Scalar,
-        Standard: Distribution<Rotation3<T>>,
+        T: RealField,
     {
         Rotation3::from_matrix(m).into()
     }
--- a/src/geometry/quaternion_construction.rs
+++ b/src/geometry/quaternion_construction.rs
@@ -410,9 +410,11 @@ where
     /// This is an iterative method. See `.from_matrix_eps` to provide mover
     /// convergence parameters and starting solution.
     /// This implements "A Robust Method to Extract the Rotational Part of Deformations" by Müller et al.
+    #[cfg(feature = "rand-no-std")]
     pub fn from_matrix(m: &Matrix3<T>) -> Self
     where
-        T: RealField,
+        T: RealField + Scalar,
+        Standard: Distribution<Rotation3<T>>,
     {
         Rotation3::from_matrix(m).into()
     }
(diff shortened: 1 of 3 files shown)


timethy:
I changed the method to perturb the rotation after convergence, and if it changes (to smaller norm), it continues from there. This should work for every possible input, and it is deterministic.

Note that the perturbation axes has to change to avoid being stuck in a maximum for ever (with no change in norm). (If e.g. axis = (1, 0, 0) and m = diag(1, -1, -1) it stays stuck starting from rot = diag(1, 1, 1).)

If I understand correctly, there is at least one axes in which the norm changes, so by cycling through three orthogonal axes it should prevent being stuck.

cpg314:
@sebcrozet 
I think this is a super important PR that fixes a critical issue that has already consumed a lot of time from multiple users (#627, @1078).
Any chance this could be merged? I can vouch for the quality of @timethy 's implementation.

timethy:
Hi, I created an alternative way of fixing this bug here:
https://github.com/dimforge/nalgebra/pull/1132

Pick your poison -- the other one relies on sqrt() and this one here uses a fixed constant that is somewhat arbitrary.

sebcrozet:
@timethy Hey! Thanks. What’s the benefit of using `eps.sqrt()` instead of the arbitrary constant?

timethy:
> 

It has the benefit of being very small (pi/8 is not really very small), but being larger than eps.

One could have another small constant, but it should be strictly larger than eps.

Downside is it assumes that 0 < eps < 1, and the latter assumption might not be true, although I don't see a use case of that big of an epsilon.

sebcrozet:
@timethy That makes sense thanks. Could you please:
- Apply the `sqrt` change to this PR (on top of the coding-style changes I made here).
- Use the `pi / 8` increment if `eps >= 1`, just to handle this very unlikely case.

timethy:
So eps_disturbance needs to be significantly larger then eps to make convergence fast (especially when eps = T::default_epsilon()), so I changed it to eps_disturbance = max(sqrt(eps), eps^2) and added some tests.

sebcrozet:
Perfect, thanks!

cpg314:
Thanks, @timethy and @sebcrozet !

PR #1107, opened 2022-05-03
adds row and column iterators for sparse matrices

aujxn:
A nice convenience iterator that I end up using often. Iterates over `(column_index, &T)` for CSR rows and `(row_index, &T)` for CSC columns (and mutable counterparts).

### DIFF
--- a/nalgebra-sparse/src/csc.rs
+++ b/nalgebra-sparse/src/csc.rs
@@ -7,7 +7,9 @@
 mod csc_serde;
 
 use crate::cs;
-use crate::cs::{CsLane, CsLaneIter, CsLaneIterMut, CsLaneMut, CsMatrix};
+use crate::cs::{
+    CsInnerIter, CsInnerIterMut, CsLane, CsLaneIter, CsLaneIterMut, CsLaneMut, CsMatrix,
+};
 use crate::csr::CsrMatrix;
 use crate::pattern::{SparsityPattern, SparsityPatternFormatError, SparsityPatternIter};
 use crate::{SparseEntry, SparseEntryMut, SparseFormatError, SparseFormatErrorKind};
@@ -717,6 +719,14 @@ macro_rules! impl_csc_col_common_methods {
             pub fn get_entry(&self, global_row_index: usize) -> Option<SparseEntry<'_, T>> {
                 self.lane.get_entry(global_row_index)
             }
+
+            /// Iterator over the row indices and elements of a column of a CSC matrix.
+            /// Equivalent to `col.row_indices().iter().zip(col.values().iter())`.
+            #[inline]
+            #[must_use]
+            pub fn iter(&self) -> CsInnerIter<'_, T> {
+                self.lane.iter()
+            }
         }
     };
 }
@@ -743,6 +753,14 @@ impl<'a, T> CscColMut<'a, T> {
     #[must_use]
     pub fn get_entry_mut(&mut self, global_row_index: usize) -> Option<SparseEntryMut<'_, T>> {
         self.lane.get_entry_mut(global_row_index)
+    }
+
+    /// Iterator over the row indices and mutable elements of a column of a CSC matrix.
+    /// Equivalent to `col.row_indices().iter().zip(col.values_mut().iter_mut())`.
+    #[inline]
+    #[must_use]
+    pub fn iter_mut(&mut self) -> CsInnerIterMut<'_, T> {
+        self.lane.iter_mut()
     }
 }
 
--- a/nalgebra-sparse/src/csr.rs
+++ b/nalgebra-sparse/src/csr.rs
@@ -7,7 +7,9 @@
 mod csr_serde;
 
 use crate::cs;
-use crate::cs::{CsLane, CsLaneIter, CsLaneIterMut, CsLaneMut, CsMatrix};
+use crate::cs::{
+    CsInnerIter, CsInnerIterMut, CsLane, CsLaneIter, CsLaneIterMut, CsLaneMut, CsMatrix,
+};
 use crate::csc::CscMatrix;
 use crate::pattern::{SparsityPattern, SparsityPatternFormatError, SparsityPatternIter};
 use crate::{SparseEntry, SparseEntryMut, SparseFormatError, SparseFormatErrorKind};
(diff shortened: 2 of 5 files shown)


nalgebra-sparse/src/csr.rs
```diff
@@ -719,6 +721,14 @@ macro_rules! impl_csr_row_common_methods {
             pub fn get_entry(&self, global_col_index: usize) -> Option<SparseEntry<'_, T>> {
                 self.lane.get_entry(global_col_index)
             }
+
+            /// Iterator over the column indices and elements of a row of a CSR matrix.
+            /// Equivalent to `row.col_indices().iter().zip(row.values().iter())`.
+            #[inline]
+            #[must_use]
+            pub fn iter(&self) -> CsInnerIter<'_, T> {
```
Andlon, line 729: I'm wondering if perhaps it would be better to have a more descriptive name, to indicate what precisely is being iterated over. Otherwise I feel as if code using this iterator might be hard to read, since the reader might not have any clue what *precisely* is being iterated over? What about  `col_and_value_iter`?

On the other hand, I guess it also comes down to how the user calls the iterator. For example,
```rust
for (col_idx, value) in row.iter() {}
```
is pretty clear. In fact, we could maybe even implement `IntoIterator` :thinking:
```rust
for (col_idx, value) in row {}
```
but we can leave this for the future though!

Do you have any thoughts on the naming?

nalgebra-sparse/src/csr.rs
```diff
@@ -7,7 +7,9 @@
 mod csr_serde;
 
 use crate::cs;
-use crate::cs::{CsLane, CsLaneIter, CsLaneIterMut, CsLaneMut, CsMatrix};
+use crate::cs::{
+    CsInnerIter, CsInnerIterMut, CsLane, CsLaneIter, CsLaneIterMut, CsLaneMut, CsMatrix,
```
Andlon, line 11: Forgive me if I'm wrong, but I think `CsInnerIter` is exposed in the API but it's not actually publicly exported? I guess the compiler warns about this?

nalgebra-sparse/src/csr.rs
Andlon: I'm also thinking maybe we should do like e.g. `CsrRow` and `CscCol` and wrap `CsInnerIter` in CSR/CSC-specific types, so that we can provide more specific documentation.

One thing I've tried hard to do in `nalgebra-sparse` is to make it easy to understand how things work. So if you want to use a CSR matrix you shouldn't need to know about CSC matrices. In particular, the generalization of rows/cols to lanes is something I've mostly tried to "hide" in the API of CSR/CSC, because this generalization makes it harder to understand. This means of course that we pay with this with some slight duplication in the APIs. What's your opinion on this?

Andlon (commented):
Thank you very much @aujxn! Sorry that I've taken so much time in getting back to you, I've been very busy in the last weeks.

I think this is already in very good shape. I'm wondering if perhaps we should wrap the `CsInner` stuff in CSR/CSC-specific types for clarity of documentation. See my comments and let me know what you think?

nalgebra-sparse/src/csr.rs
aujxn: I like your idea of hiding the lane iterator behind both a specific row and column version. I agree that makes things more clear and has little cost. I also like the idea of implementing 'IntoIterator', I don't mind changing the name if you think the clarity is needed here as well.

nalgebra-sparse/src/csc.rs
```diff
@@ -720,12 +720,14 @@ macro_rules! impl_csc_col_common_methods {
                 self.lane.get_entry(global_row_index)
             }
 
-            /// Iterator over the row indices and elements of a column of a CSC matrix.
+            /// Iterator over `(row_idx, value)`, the row indices and elements of a column of a CSC matrix.
             /// Equivalent to `col.row_indices().iter().zip(col.values().iter())`.
```
Andlon, line 724: Can you perhaps manually check the rendering for the docs for the new functionality by running `cargo doc`? I think you need a line break after the first sentence here, otherwise you'll get the second line in the "summary", which bloats the function overview for the module.

Same applies for `iter_mut` and CSR variants etc.!

nalgebra-sparse/src/csc.rs
```diff
@@ -792,3 +796,35 @@ where
         self.lane_iter.next().map(|lane| CscColMut { lane })
     }
 }
+
+/// Iterator for [CscCol](struct.CscCol.html).
```
Andlon, line 800: You can directly use [Rustdoc intra-doc links](https://doc.rust-lang.org/nightly/rustdoc/write-documentation/linking-to-items-by-name.html) here, which is preferred to directly writing the URL. In this case I believe it would be simply ``[`CscCol`]``, although please test this by rendering the docs with `cargo test`.

The same applies to the other doc comments with doc links!

Andlon (changes requested):
Thanks, this looks great!

I only have a few minor comments related to documentation and naming. For example,
maybe we should rename `CsrInnerIter` to `CsrRowIter` and `CscInnerIter` to `CscColIter`? What do you think?

Please see code comments for the other stuff.

aujxn:
> Thanks, this looks great!
> 
> I only have a few minor comments related to documentation and naming. For example, maybe we should rename `CsrInnerIter` to `CsrRowIter` and `CscInnerIter` to `CscColIter`? What do you think?
> 
> Please see code comments for the other stuff.

This is actually what I wanted to originally name them but these names are used for the iterators over the lanes (outer iterators). So I wasn't sure what name would be next best.

Could possibly rename those outer iterators but then I'm not sure what would be a clear name for them. This is partially what motivated the `inner` scheme because it clearly distinguishes it as iterating within the lane instead of over the lanes.

Andlon:
@aujxn: ah, I had totally forgotten about the naming of "outer" iterators. Right, I suppose `Inner` is not too bad then :-)

Yeah, in retrospect the naming of the outer iterators is maybe not ideal. Perhaps they should have been `CsrRowsIter` and `CscColsIter`? Not sure... In any case, it seems like a bad idea to directly "replace" `CsrRowIter` with `CsrInnerIter`, since the current `CsrRowIter` may be directly named in user code, which would cause some very confusing error messages upon upgrade.

If we want to go ahead with a name change we should probably rather have a long term perspective:
1. rename `CsrRowIter` to `CsrRowsIter`
2. create *deprecated* type alias `CsrRowIter = CsrRowsIter`
3. in a (much) later release, in which `CsrRowIter` has already been deprecated for several releases, rename `CsrInnerIter` to `CsrRowIter`

I'll create a separate issue to track this plan if we want to go ahead with it. In any case, it seems to me that we should keep the current naming. What do you think?

Andlon:
@aujxn: this seems to have stalled, and that might have been partially my fault, I'm sorry. Do you have any interest in pushing this across the finishing line? I think it would make a nice addition to `nalgebra-sparse`.

Issue #1108, opened 2022-05-08
Method `rank` exists, but its trait bounds were not satisfied


mmagician:
I'm trying to compute a rank of a square 256x256 matrix. It works for smaller sizes (up to 127), but for anything bigger I get:
```
the method `rank` exists for struct `Matrix<f32, nalgebra::Const<256_usize>, nalgebra::Const<256_usize>, ArrayStorage<f32, 256_usize, 256_usize>>`, but its trait bounds were not satisfied
the following trait bounds were not satisfied:
`nalgebra::Const<256_usize>: DimMin<nalgebra::Const<256_usize>>`
`nalgebra::Const<256_usize>: ToTypenum`rustc[E0599](https://doc.rust-lang.org/error-index.html#E0599)
dimension.rs(212, 1): doesn't satisfy `_: DimMin<nalgebra::Const<256_usize>>`
dimension.rs(212, 1): doesn't satisfy `nalgebra::Const<256_usize>: ToTypenum`
```

It sounds like a pretty basic operation to do, and the matrices aren't THAT big. Is there an easy solution to this?
Many thanks!

Ralith:
Can you use dynamically sized matrices for this? 256x256 is rather large for a value type.

mmagician:
I'll try

trueb2:
In embedded (no_std), a dynamically sized matrix isn't really an option, but there are still plenty of matrix operations that would be reasonable to perform on 128, 256, 512, 1024, etc. This `rank` invocation issue isn't really about `rank`, but rather the impl's of `Const<T>`.

Is there a limiting factor preventing add those to the library? I don't know the story behind Const only being implemented up to 127, i.e. why not at least stop on a power of 2. https://github.com/dimforge/nalgebra/pull/820 seems to imply that the Const generics improvement stopped at 127 because no one asked for more. I would be happy to have more on the power of two.

Here is a link to the macro impl for ToTypenum through 127: https://github.com/dimforge/nalgebra/blob/dev/src/base/dimension.rs#L309

mmagician:
@Ralith Thanks, that indeed works with dynamically sized matrices. It's a shame that static allocation only works for such small matrices though.

PR #1111, opened 2022-05-25
Add Const<T> impl through 1024

trueb2:
This allows using larger constant size dimensions for a number of interfaces. This is particularly useful in no_std situations where statically sized matrix algebra should be possible but isn't because of the hard limit at 127.

I am not sure if this affects anything other than compile times, so I checked that before and after. Incremental builds take a couple seconds depending on the file touched, but there is a difference on a clean build.

* Before this change from clean:
  * Debug: 17.52s
  * Release: 15.94s
* After this change from clean:
  * Debug: 52.69s
  * Release: 45.35s

There are a number of related issues like https://github.com/dimforge/nalgebra/issues/1108


Ralith:
I would very much prefer for the clean build time to not be so massively inflated. If this is absolutely needed for your use case (are you sure you can't use alloc?) then perhaps it could be feature-gated?

trueb2:
> I would very much prefer for the clean build time to not be so massively inflated. If this is absolutely needed for your use case (are you sure you can't use alloc?) then perhaps it could be feature-gated?

I agree that clean build time inflation is an issue of consideration but do think shouldn't prevent using the library in no_std or no alloc situations. Since this library is so powerfully generic and this is common problem that new comers hit, I think there should be a solution that generally works and has a workaround rather than having to dive into the macro call site and GitHub issues.

Personally, I have a couple embedded projects, where I have a couple KBs left. I mostly want to use power of two sizes, but the RAM constraints of embedded make it such that I can't always pick a nice round number. For example, various applications may only be able to dedicate something like 10-50KB for the stack usage and the static memory buffers. On top of that dynamic memory allocation is simply unacceptable due to the unpredictable regularity of OOM.

Feature-gating is a problem for me because libraries that use nalgebra as a dependency and interface on the types would need features configured. However, cargo doesn't have a pattern that maintainers follow for saying no_std, no alloc and just feature X enabled in a shared dependency.

If I had the option of specializing Const for Typenum as needed for my use case from the call site module, I would be satisfied. If I had just powers of two and multiples of 25, I would also be satisfied. I would also be satisfied if the compiler error was a little more helpful when I try to use Const and Typenum on a Const number that is out of range.

sebcrozet:
Hey! Thank you for the proposal. Clean compile times of `nalgebra` is already a common subject of complain within the community, so multiplying them by 3 isn’t a good option. This will also increase the sime of the `.rlib` (but I don’t that maters much). Feature-gating sounds like the only reasonable solution here.

> However, cargo doesn't have a pattern that maintainers follow for saying no_std, no alloc and just feature X enabled in a shared dependency.

I’m not sure I understand the problem with feature-gating here. The user could depend on `nalgebra` like this:

```toml
[dependencies]
nalgebra = { version = "*", default-features = false, features = [ "large-const-generics" ] }
```

This won’t enable the `std` or `alloc` features of `nalgebra`.


> If I had the option of specializing Const for Typenum as needed for my use case from the call site module, I would be satisfied. If I had just powers of two and multiples of 25, I would also be satisfied. I would also like if the compiler error was a little more helpful when I try to use Const and Typenum on a Const number that is out of range.

Unfortunately, we are limited by Rust’s minimal support of const-generics here.

Ralith:
> will require documentation and feature additions to bubble up through the community.

Yeah, fortunately not necessary.

> this is common problem that new comers hit

Is it? I'm not sure what proportion of users are doing high-dimensional math to begin with, and I expect the overwhelming majority of those are happily using dynamic sizes.

> Build from clean with just 0 through 16, debug was 13.26s and release was 11.81.

Ooh, about 25% savings, not bad! Might be nice to get in the next breaking release. We could even go lower by gating some of the specializations; there's certainly a segment of users that don't need static sizes for anything > 4, if that, but I expect the returns start diminishing at some point.

trueb2:
I checked trimming down some of those vector and matrix specializations on my machine. The const R values are the minimum before some of the other stuff that uses const like DualQuaternion fails to build.

```
// Implement for vectors of dimension 1 .. 4.
impl_from_into_asref_1D!(
    // Row vectors.
    (U1, U1 ) => 1;  (U1, U2 ) => 2;  (U1, U3 ) => 3;  (U1, U4 ) => 4;

    // Column vectors.
    (U2 , U1) => 2;  (U3 , U1) => 3;  (U4 , U1) => 4;
);

// Implement for matrices with shape 2x2 .. 4x4.
impl_from_into_asref_borrow_2D!(
    (U2, U2) => (2, 2); (U2, U3) => (2, 3); (U2, U4) => (2, 4); 
    (U3, U2) => (3, 2); (U3, U3) => (3, 3); (U3, U4) => (3, 4);
    (U4, U2) => (4, 2); (U4, U3) => (4, 3); (U4, U4) => (4, 4);
);

from_to_typenum!(
    /*U1,1;*/ U2, 2; U3, 3; U4, 4; U5, 5; U6, 6; U8, 8
);
```

There wasn't any further change in build time on my machine (13.1s and 11.9s). Since it seems that the majority of the build time changes are just due to the const to typenum invocation, I suppose that I should update the PR 
* without changes to the vector and matrix specializations
* with 0 through 16 always enabled
* with 'const-generics-128' feature flag
* with 'const-generics-512' feature flag
* with 'const-generics-1024' feature flag

Ralith:
A change that move existing code behind a feature gate is breaking, so it might be useful to split that up from the one that adds new code, if you/@sebcrozet want to get the extended sizes in a patch release soon.

trueb2:
I should have commented earlier, but the previous commit moved just the new specializations through 1024 behind a feature flag. Therefore, this PR is not a breaking change.

PR #1112, opened 2022-05-25
Make "Point::new" a const fn

SolraBizna:
Based on my brief testing, these two changed lines are all that is needed to make this work.


Ralith:
We should probably do this for more than just the 1D case.

SolraBizna:
My change works for every case that a `Point::new` is generated for, not just 1D. (GitHub's diff makes this hard to tell.)

Ralith:
Right you are.

Ralith (approved):

necauqua:
Hi!
Any movement on this?
Seems like an extremely simple improvement

sebcrozet:
Thanks!

Issue #1115, opened 2022-06-05
Point's Display implementation doesn't respect formatter params


RReverser:
Normally, Rust data structures propagate formatter params recursively to individual values.

nalgebra does this too for matrices and vectors, but Points seem to ignore the formatter params.

Sample code:

```rust
let matrix = nalgebra::Matrix3::<f64>::new_random();
let vec = nalgebra::Vector3::<f64>::new_random();
let point = nalgebra::Point3::from(nalgebra::Vector3::<f64>::new_random());

println!("Matrix {matrix}");
println!("Vector {vec}");
println!("Point {point}");
println!("---");
println!("Matrix {matrix:.3}");
println!("Vector {vec:.3}");
println!("Point {point:.3}");
```

Output:

```
Matrix 
  ┌                                                             ┐
  │  0.8598913012304822  0.9260866371615168 0.46836528545042544 │
  │  0.6925437372908951  0.5359048989574623  0.3807341326886613 │
  │  0.6538521469022629 0.45065417326429935 0.46616147800849017 │
  └                                                             ┘


Vector
  ┌                    ┐
  │ 0.5324899032766206 │
  │ 0.9758313347345605 │
  │ 0.7908423868945029 │
  └                    ┘


Point {0.5597363785908459, 0.49737074839321604, 0.5159609296629915}
---
Matrix
  ┌                   ┐
  │ 0.860 0.926 0.468 │
  │ 0.693 0.536 0.381 │
  │ 0.654 0.451 0.466 │
  └                   ┘


Vector
  ┌       ┐
  │ 0.532 │
  │ 0.976 │
  │ 0.791 │
  └       ┘


Point {0.5597363785908459, 0.49737074839321604, 0.5159609296629915}
```

waywardmonkeys:
So, there were 2 PRs that addressed this (both still open): #1116 and #1119 ... but I wanted to mention a different way, by passing the formatter directly through to `fmt::Display::fmt`, you get the forwarding of all of the formatter properties.  For `Point`, the diff looks like this:

```
diff --git a/src/geometry/point.rs b/src/geometry/point.rs
index e936c3da..39d19a59 100644
--- a/src/geometry/point.rs
+++ b/src/geometry/point.rs
@@ -499,10 +499,11 @@ where

         let mut it = self.coords.iter();

-        write!(f, "{}", *it.next().unwrap())?;
+        fmt::Display::fmt(it.next().unwrap(), f)?;

         for comp in it {
[...]

PR #1116, opened 2022-06-06
Fix Point display to respect formatting parameters

owenbrooks:
This changes Point Display formatting to match formatting of other geometry types, including Isometry, Scale, and Rotation.
Fixes issue #1115 since it passes through the desired precision formatting parameter.

Example:
```rust
let point = nalgebra::Point3::new(1.12345678, 2.12345678, 3.12345678);
println!("{point:.4}");
```

Before:
```
{1.12345678, 2.12345678, 3.12345678}
```
After:
```
Point {
  ┌        ┐
  │ 1.1235 │
  │ 2.1235 │
  │ 3.1235 │
  └        ┘

}
```

### DIFF
--- a/src/geometry/point.rs
+++ b/src/geometry/point.rs
@@ -458,16 +458,10 @@ where
     DefaultAllocator: Allocator<T, D>,
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{{")?;
+        let precision = f.precision().unwrap_or(3);
 
-        let mut it = self.coords.iter();
-
-        write!(f, "{}", *it.next().unwrap())?;
-
-        for comp in it {
-            write!(f, ", {}", *comp)?;
-        }
-
+        write!(f, "Point {{")?;
+        write!(f, "{:.*}", precision, self.coords)?;
         write!(f, "}}")
     }
 }


Ralith:
I think it would be better to just forward to the underlying `Vector`'s `fmt`. Reduces verbosity and handles flags other than precision correctly as well.

owenbrooks:
That does seem simpler, thanks. I've made that change. Which other flags were you expecting it to handle when forwarding? From what I can see it looks like the underlying matrix formatting ignores most parameters aside from precision. I'm wondering if it would be worth forwarding `fmt` for the other geometry types too.

Ralith:
Ah, if `Matrix::display` isn't handling e.g. padding and signs properly then we at least have a single place to fix that. What other types currently having `Display` impls did you have in mind?

src/geometry/point.rs
```diff
@@ -458,16 +458,8 @@ where
...
-
-        let mut it = self.coords.iter();
-
-        write!(f, "{}", *it.next().unwrap())?;
-
-        for comp in it {
-            write!(f, ", {}", *comp)?;
-        }
-
+        write!(f, "Point {{")?;
```
Ralith, line 461: I'd rather omit this wrapping for concision, personally. It's also unconventional for `Display` impls in general.

owenbrooks:
The Isometry, Scale, Rotation, and Translation types all have similar Display impls that only pass through precision and have similar wrapping with “Typename {“ and “}”. I would prefer omitting the wrapper myself too but would probably want it to be consistent.

Ralith:
Ah, good call.

RReverser:
(reporter of the referenced issue here) Hm, I actually liked the concise single-line form of Point, while the new form makes it a bit more verbose.

I guess it does make it more consistent with other formats though...

owenbrooks:
> (reporter of the referenced issue here) Hm, I actually liked the concise single-line form of Point, while the new form makes it a bit more verbose.
> 
> I guess it does make it more consistent with other formats though...

Ah, have you tried debug printing? It's concise and actually does follow all format specifiers:
```rust
let point = nalgebra::Point3::new(1.12345678, 2.12345678, 3.12345678);
println!("{point:.4?}");
```
Gives `[1.1235, 2.1235, 3.1235]`

RReverser:
> Ah, have you tried debug printing? It's concise and actually does follow all format specifiers:

Huh, I haven't! It's actually a bit surprising that Debug formatting in this case looks more like what I'd expect from Display (being more human-readable and all) while Display formatting in this PR is more like what I'd expect from Debug (showing type names and all).

Ralith:
Conversely, fancy many-line 2D layout with unicode block drawing is *worlds* apart from what you'd expect in `Debug`.

RReverser:
> Conversely, fancy many-line 2D layout with unicode block drawing is _worlds_ apart from what you'd expect in `Debug`.

Sounds like we're in agreement. For matrix types I'd expect neat block drawing in Display, but type names and such in Debug.

In case of `Point` it's just a bit different because, personally, I wouldn't want either of Display or Debug to show a vertical block but rather show values in line for convenience like `Point` does today.

owenbrooks:
I've tried to reduce these inconsistencies in #1119. Would appreciate feedback!

PR #1120, opened 2022-06-11
Implement `cast` for `Unit<Vector<T, D, S>>`

Ralith:
Currently, `cast` is resolved via `Unit`'s `Deref` impl, which leads to it confusingly stripping the `Unit` from `UnitVector`s. Add an inherent impl which takes precedence, similar to the existing specialization for `UnitQuaternion`.

Wasn't entirely sure where to put the definition, hope it makes sense where it is.

### DIFF
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -2186,3 +2186,28 @@ where
         }
     }
 }
+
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thanks!

PR #1129, opened 2022-07-19
update try_inverse doc comment

zhiburt:
Hi there,

I was assuming that `try_*` method is not suppose to panic, so initially I intended to make it return `false` but then decided to just create a comment.

What do you think?


sebcrozet:
Thanks!

PR #1130, opened 2022-07-23
Support conversion for glam `0.21`.

devil-ira:
Add feature `convert-glam021`.

Conversion macros changed to use the `From` trait to convert glam types to arrays because The `BVec*` types no longer implement `AsRef`.


sebcrozet:
Thanks!

PR #1138, opened 2022-08-09
Update DefaultAllocator docs to say ArrayStorage instead of GenericArray

ritobanrc:
Minor modification to docs, since IIUC `GenericArray` is no longer used by nalgebra.


sebcrozet:
Thanks!

PR #1140, opened 2022-08-12
Use #[inline] to improve opt-level 1 performance

### LINKED ISSUE #1139: Performance regression since v0.26.0 with likely simple fix
### Summary and likely fix
From v0.25.4 to v0.26.0, certain functions that used to be marked `#[inline]` are no longer marked inline. Specifically, these are the three functions in `unsafe impl<const T: usize> Dim for Const<T>` in `dimension.rs`. This has caused a 5x performance regression on a project depending on nalgebra, and I am wondering if it's possible to add `#[inline]` back to these 3 functions.

Please let me know if there are additional details I can provide or alternative suggestions, as this is currently blocking me from updating nalgebra in this project.

### Details
Specifically, in a project I am involved in, multiplying a `Matrix<f64, Const<4>, Const<4>, ArrayStorage<f64, 4, 4>>` by a `&Matrix<f64, Const<4>, Const<4>, ArrayStorage<f64, 4, 4>>` and assigning that result to a new variable of type `Matrix<f64, Const<4>, Const<4>, ArrayStorage<f64, 4, 4>>` became about 5 times slower when the project was compiled with optimization level 1, even when nalgebra was compiled with optimization level 2.

I am able to fix this performance regression by adding `#[inline]` to each of the three functions in this trait implementation. Would it be possible to apply this fix to the canonical version of nalgebra, or is there a reason these functions cannot be marked inline?

I have had trouble trying to make a minimal working example that demonstrated this issue. While I was able to reproduce a performance regression along with the fix, it was an almost 1.8x regression instead of 5x, although this still shows that the fix is worthwhile.

### Extra context
- I noticed the issue when trying to upgrade `nalgebra` in Hypermine. The link to the matrix multiplication I have been testing with is here: https://github.com/Ralith/hypermine/blob/5ba603caf26012688f71265cb0868caff38b7c04/client/src/graphics/draw.rs#L604
- A `git bisect` on nalgebra yielded b2dadffcf24836ced2d16b05b2a17bf3815977b8 as the last good commit and 35ec135d2cf777bc1d2b8e4a519d5bb1150892b0 as the first bad commit. There is one commit between these two, but it doesn't compile.
- Running the debugger (using VSCode with rust-analyzer on Windows) on the good commit in Hypermine using "step into" every time only hit the functions `Matrix::from_uninitialized_data`, `Dim::is`, `array_axc`, `array_axcpy`, and `Scalar::inlined_clone`. On the bad commit, it also hit `Dim::value` many times, along with `gemm`, `gemv`, and some other functions.

### How to reproduce partially
I tried several things to get example code to behave the same way as Hypermine, but I was unsuccessful. However, the following sample code ran in 230ms on my computer, and it improved to 130ms after applying the fix:

main.rs
```rust
fn main() {
    let first_matrix = na::Matrix4::<f64>::new(1., 1., 1., 1., 1., 2., 3., 4., 1., 4., 9., 16., 1., 8., 27., 64.);
    let second_matrix = na::Matrix4::<f64>::new(2., 3., 5., 7., 11., 13., 17., 19., 23., 29., 31., 37., 41., 43., 47., 53.);

    let start = std::time::Instant::now();

    let mut accumulator = 0.;

    for _ in 0..10_000_000 {
        accumulator += inner(&first_matrix, &second_matrix);
    }

    println!("Accumulator to avoid optimizing away work: {}", accumulator);
    println!("{:?}", start.elapsed());
}

#[inline(never)]
fn inner(first_matrix: &na::Matrix4<f64>, second_matrix: &na::Matrix4<f64>) -> f64 {
    let third_matrix = first_matrix * second_matrix;
    *unsafe { third_matrix.get_unchecked((0, 0)) }
}
```

Cargo.toml
```toml
[package]
name = "sandbox"
version = "0.1.0"
edition = "2018"

[dependencies]
na = { package = "nalgebra", path = "my/local/path/to/nalgebra" }

# Replicating Hypermine's cargo.toml
[profile.dev]
opt-level = 1

[profile.dev.package."*"]
opt-level = 2

[profile.release.build-override]
opt-level = 0
```

patowen:
fixes #1139

My dev environment is not set up properly for nalgebra, so I've been unable to test this locally in any of the standard ways, but these changes work correctly with Hypermine, and the changes are simple enough that I believe it's worth creating a PR anyway.

Please let me know if there's anything I need to do before this can be merged, if it can be merged at all.

### DIFF
--- a/src/base/dimension.rs
+++ b/src/base/dimension.rs
@@ -252,14 +252,17 @@ pub trait ToTypenum {
 }
 
 unsafe impl<const T: usize> Dim for Const<T> {
+    #[inline]
     fn try_to_usize() -> Option<usize> {
         Some(T)
     }
 
+    #[inline]
     fn value(&self) -> usize {
         T
     }
 
+    #[inline]
     fn from_usize(dim: usize) -> Self {
         assert_eq!(dim, T);
         Self


sebcrozet:
Thanks!

PR #1141, opened 2022-08-16
[Sparse] Add `clear_triplets()`

lsh:
This PR adds the `clear_triplet()` method to the `CooMatrix` struct.
The goal of the method is to be able to remove triplets before construction.

Some outstanding questions:
* Spelling of the method? @Andlon suggested `clear_triplet` in the Discord. Another option might be `pop` or `pop_triplet` to correspond to `push`.
* Should out of bounds values cause a panic? It is probably fine to just return a `None` value.
* What should the return type be? Currently it is `Option<T>`.

### DIFF
--- a/nalgebra-sparse/src/coo.rs
+++ b/nalgebra-sparse/src/coo.rs
@@ -211,6 +211,26 @@ impl<T> CooMatrix<T> {
         self.values.push(v);
     }
 
+    /// Remove a single triplet from the matrix.
+    ///
+    /// This removes the value `v` from the `i`th row and `j`th column in the matrix.
+    pub fn clear_triplet(&mut self, i: usize, j: usize, v: T) -> Option<(usize, usize, T)>
+    where
+        T: PartialEq,
+    {
+        let triple_idx = self
+            .triplet_iter()
+            .position(|triplet| triplet == (i, j, &v));
+        if let Some(triple_idx) = triple_idx {
+            self.row_indices.remove(triple_idx);
+            self.col_indices.remove(triple_idx);
+            let retv = self.values.remove(triple_idx);
+            Some((i, j, retv))
+        } else {
+            None
+        }
+    }
+
     /// The number of rows in the matrix.
     #[inline]
     #[must_use]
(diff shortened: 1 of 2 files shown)


Andlon:
Thanks for working on this @lsh!

Actually, I suggested `clear_triplets` (note plural), for removing *all* triplets. This is very useful as it would allow us to re-use already allocated COO storage and therefore amortize allocations. I'm not so sure if removing a *single* triplet based on coordinates is very useful as it is terribly inefficient - do we have a compelling motivating use case for this?

lsh:
@Andlon whoops, sorry about my misunderstanding 😅. Updated the code to implement `clear_triplets`.

Andlon:
@lsh: great, thanks. CI is failing because you have the wrong return type for `clear_triplets`, maybe try running tests locally?

nalgebra-sparse/src/coo.rs
```diff
@@ -211,6 +211,16 @@ impl<T> CooMatrix<T> {
         self.values.push(v);
     }
 
+    /// Clear all triplets from the matrix.
+    pub fn clear_triplets(&mut self)
+    where
+        T: PartialEq,
```
Andlon, line 217: This bound is not necessary

Andlon (changes requested):
@lsh: just one unnecessary trait bound that needs to be removed, otherwise looks good! Tests are still running but assuming they succeed after removing the trait bound we should be good to go.

Andlon:
@sebcrozet: not sure why the CUDA build is failing. Do you? Looks like some strange issues with the recent changes that made `Point::new` a `const fn`?

sebcrozet:
@Andlon You’re right, looks like https://github.com/dimforge/nalgebra/pull/1112 broke the CUDA build. I think this is due to the fact that the CUDA build uses an older version of the compiler which only allowed `Sized` trait bounds for `const fn`. We should add a non-const version when targeting `cuda`.

Andlon:
Looks good to me, ready for merge. @sebcrozet: do we wait with merging until the CUDA CI problems are fixed, or what's your preference?

sebcrozet:
Thank you both! No need to delay this on something unrelated.

PR #1153, opened 2022-09-15
Fix usage of CheckBytes

peng1999:
According to the [rkyv book](https://rkyv.org/validation.html), the `CheckBytes` trait derive should be applied inside `archive_attr` attribute, or we cannot deserialize data without `unsafe` code.

This PR also added some trivial tests for rkyv serialization and deserialization.

### DIFF
--- a/.github/workflows/nalgebra-ci-build.yml
+++ b/.github/workflows/nalgebra-ci-build.yml
@@ -61,7 +61,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: test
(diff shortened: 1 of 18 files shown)


peng1999:
@sebcrozet The CI failed on proptest, but the failed function seems unrelated to this PR and I cannot reproduce using the seed proptest prompts.

peng1999:
@sebcrozet The CI has been fixed in dev, and I have rebased on it. Could you please re-trigger the workflows?

sebcrozet:
Thanks!

PR #1158, opened 2022-10-01
[Sparse] Add `triplet_iter_mut()`

lsh:
There are certain cases where one might want to modify the values stored in the triplets in a COO matrix before constructing a `Csr` or `Csc` matrix.

Currently the way to do this would either be by decomposing the `CooMatrix`

```rust
let mut coo = /* ... */;
let nrows = coo.nrows();
let ncols = coo.ncols();
let (rows, cols, mut values) = coo.disassemble();
values.iter_mut().for_each(|v| /* ... */);
CooMatrix:: try_from_triplets(nrows, ncols, rows, cols, values).unwrap();
```
Or clearing the triplets and inserting new values
```rust
let mut coo = /* ... */;
let new_triplets = coo.triplet_iter().map(|(i, j, &v)|  (i, j, /* mutate v */));
coo.clear_triplets();
for (i, j, v) in new_triplets {
  coo.push(i,j,v);
}
```

### DIFF
--- a/nalgebra-sparse/tests/unit_tests/coo.rs
+++ b/nalgebra-sparse/tests/unit_tests/coo.rs
@@ -95,16 +95,30 @@ fn coo_triplets_iter_mut() {
     let v = vec![2, 3, 4, 7, 1, 3, 1, 5];
     let mut coo =
         CooMatrix::<i32>::try_from_triplets(3, 5, i.clone(), j.clone(), v.clone()).unwrap();
-    coo.triplet_iter_mut().for_each(|(_i, _j, v)| *v = *v * *v);
+
+    let actual_triplets: Vec<_> = coo.triplet_iter_mut().map(|(i, j, v)| (i, j, *v)).collect();
 
     let expected_triplets: Vec<_> = i
         .iter()
         .zip(&j)
         .zip(&v)
-        .map(|((i, j), v)| (*i, *j, v * v))
+        .map(|((i, j), v)| (*i, *j, *v))
         .collect();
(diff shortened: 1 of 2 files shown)


Andlon:
Thanks, this looks like a worthwhile addition. Would be great if you can add some tests, then I think we should be good to merge!

nalgebra-sparse/tests/unit_tests/coo.rs
```diff
@@ -87,6 +87,26 @@ fn coo_construction_for_valid_data() {
...
 
+#[test]
+fn coo_triplets_iter_mut() {
+    // Arbitrary matrix, with duplicates
+    let i = vec![0, 1, 0, 0, 0, 0, 2, 1];
+    let j = vec![0, 2, 0, 1, 0, 3, 3, 2];
+    let v = vec![2, 3, 4, 7, 1, 3, 1, 5];
+    let mut coo =
+        CooMatrix::<i32>::try_from_triplets(3, 5, i.clone(), j.clone(), v.clone()).unwrap();
+    coo.triplet_iter_mut().for_each(|(_i, _j, v)| *v = *v * *v);
```
Andlon, line 98: By ignoring the indices, we are not testing that the indices produced are correct (for example, they could be transposed!). The easiest way to do that is probably just to collect the mutable iterator, i.e. `triplet_iter_mut().collect()`, instead of `triplet_iter()` below and check the result against expected values.

Moreover, I think it'd be a good idea to manually write out the expected values rather than relying on programmatically producing them.

Andlon (changes requested):
Almost there - see comments :-)

Andlon:
Thanks @lsh!

PR #1162, opened 2022-10-12
Make CSR matrix example more self contained

dstansby:
Currently the CSR matrix example isn't self contained, because `coo` isn't defined in the code. This PR fixes that by defining a `coo` matrix, so users reading the example can copy/paste the code for a working example.

### DIFF
--- a/nalgebra-sparse/src/csr.rs
+++ b/nalgebra-sparse/src/csr.rs
@@ -25,6 +25,7 @@ use std::slice::{Iter, IterMut};
 /// # Usage
 ///
 /// ```
+/// use nalgebra_sparse::coo::CooMatrix;
 /// use nalgebra_sparse::csr::CsrMatrix;
 /// use nalgebra::{DMatrix, Matrix3x4};
 /// use matrixcompare::assert_matrix_eq;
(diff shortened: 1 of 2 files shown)


Andlon:
Thanks, I agree, this is a very helpful addition. Do you think you could perhaps also apply the same change to `CscMatrix`? Its documentation mirrors that of `CsrMatrix` and currently suffers from the same issue. That would be great!

nalgebra-sparse/src/csr.rs
```diff
@@ -33,8 +34,9 @@ use std::slice::{Iter, IterMut};
 /// // change the sparsity pattern of the matrix after it has been constructed. The easiest
 /// // way to construct a CSR matrix is to first incrementally construct a COO matrix,
 /// // and then convert it to CSR.
-/// # use nalgebra_sparse::coo::CooMatrix;
-/// # let coo = CooMatrix::<f64>::new(3, 3);
+///
+/// let coo = CooMatrix::<f64>::new(3, 3);
+/// coo.push(2, 0, 1.0);
```
Andlon, line 39: Ah, didn't catch this upon review, but thankfully the CI did. Need to make `coo` mutable if you want to push something to it!

dstansby:
👍 updated with both those changes

Andlon:
Thanks @dstansby!

Issue #848, opened 2021-03-10
Question: is it possible to modify different DMatrix columns in parallel (via Rayon) without Mutex?


pgagarinov:
I'm trying to make the following code work:

use nalgebra::*;
use rayon::prelude::*;

let mut m = DMatrix::<f64>::from_vec(2, 3, (0..6).map(|n| n as f64).collect());
dbg!(&m);
m.column_iter_mut()
    .par_bridge()
     .for_each(|mut c| c[0] = 111.0);
dbg!(&m);

But the code doesn't compile:

no method named `par_bridge` found for struct `ColumnIterMut<'_, f64, Dynamic, Dynamic, VecStorage<f64, Dynamic, Dynamic>>` in the current scope
the method `par_bridge` exists but the following trait bounds were not satisfied:
`ColumnIterMut<'_, f64, Dynamic, Dynamic, VecStorage<f64, Dynamic, Dynamic>>: Send`
which is required by `ColumnIterMut<'_, f64, Dynamic, Dynamic, VecStorage<f64, Dynamic, Dynamic>>: rayon::iter::ParallelBridge`
`&ColumnIterMut<'_, f64, Dynamic, Dynamic, VecStorage<f64, Dynamic, Dynamic>>: Send`
which is required by `&ColumnIterMut<'_, f64, Dynamic, Dynamic, VecStorage<f64, Dynamic, Dynamic>>: rayon::iter::ParallelBridge`
`&ColumnIterMut<'_, f64, Dynamic, Dynamic, VecStorage<f64, Dynamic, Dynamic>>: Iterator`
which is required by `&ColumnIterMut<'_, f64, Dynamic, Dynamic, VecStorage<f64, Dynamic, Dynamic>>: rayon::iter::ParallelBridge`
`&mut ColumnIterMut<'_, f64, Dynamic, Dynamic, VecStorage<f64, Dynamic, Dynamic>>: Send`
which is required by `&mut ColumnIterMut<'_, f64, Dynamic, Dynamic, VecStorage<f64, Dynamic, Dynamic>>: rayon::iter::ParallelBridge`

Is there a relatively short path for making the code above work (with some changes to either the code or nalgebra source code)? I can do some digging on my own, I just need the direction. Thanks!

sebcrozet:
[...]

PR #1165, opened 2022-10-21
Parallel Column Iterators with Rayon

geo-ant:
**Short**: Enable parallel column iterators for matrices with rayon

**Fixes**: #848 

**Changes**:
* Added functions `par_column_iter` and `par_column_iter_mut` to enable iteration over (mutable) columns in parallel
* The serial `ColumnIterMut` is now Send. I changed a few things that I am not sure make the code any more or less safe than before. I'm grateful for a review.
* Both `ColumnIter` and `ColumnIterMut` are now double ended (rayon needs this). This required me to stick another pointer sized integer in there, increasing the size of an iterator by 50%. (from 2 pointer sized members to 3)
* Added tests covering serial and parallel iterators

**Notes**:
* I will want to feature gate this functionality on crate feature `rayon` or `parallel`. I'll do that when I get greenlight on the rest of the code.

**Example Code**
We can now modify the columns of a matrix in parallel like so
```rust
//silly example, I know
let mut matrix = DMatrix::<f32>::zeros(100,100);
matrix
  .par_column_iter_mut()
  .enumerate()
  .for_each(|(idx, mut col)| col[idx] = 1.);
assert_eq!(matrix, DMatrix::identity(100,100));
```

Cheers and thank you,
Geo

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ libm    = [ "simba/libm" ]
 libm-force = [ "simba/libm_force" ]
 macros = [ "nalgebra-macros" ]
 cuda   = [ "cust_core", "simba/cuda" ]
-rayon  = [ "std", "dep:rayon" ]
+par-iter  = [ "std", "rayon" ]
 
 # Conversion
 convert-mint = [ "mint" ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ libm    = [ "simba/libm" ]
 libm-force = [ "simba/libm_force" ]
 macros = [ "nalgebra-macros" ]
 cuda   = [ "cust_core", "simba/cuda" ]
-rayon  = [ "std", "dep:rayon" ]
+rayon-par  = [ "std", "rayon" ]
 
 # Conversion
 convert-mint = [ "mint" ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ libm    = [ "simba/libm" ]
 libm-force = [ "simba/libm_force" ]
 macros = [ "nalgebra-macros" ]
 cuda   = [ "cust_core", "simba/cuda" ]
-rayon  = [ "dep:rayon" ]
+rayon  = [ "std", "dep:rayon" ]
 
 # Conversion
 convert-mint = [ "mint" ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ libm    = [ "simba/libm" ]
 libm-force = [ "simba/libm_force" ]
 macros = [ "nalgebra-macros" ]
 cuda   = [ "cust_core", "simba/cuda" ]
+rayon  = [ "dep:rayon" ]
 
 # Conversion
 convert-mint = [ "mint" ]
@@ -101,7 +102,7 @@ glam020        = { package = "glam", version = "0.20", optional = true }
 glam021        = { package = "glam", version = "0.21", optional = true }
 glam022        = { package = "glam", version = "0.22", optional = true }
 cust_core      = { version = "0.1", optional = true }
-rayon = "1.5" # TODO make this feature gated
+rayon          = { version = "1.5", optional = true }
 
 [dev-dependencies]
 serde_json = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,7 +102,7 @@ glam020        = { package = "glam", version = "0.20", optional = true }
 glam021        = { package = "glam", version = "0.21", optional = true }
 glam022        = { package = "glam", version = "0.22", optional = true }
 cust_core      = { version = "0.1", optional = true }
-rayon          = { version = "1.5", optional = true }
+rayon          = { version = "1.6", optional = true }
 
 [dev-dependencies]
 serde_json = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ libm    = [ "simba/libm" ]
 libm-force = [ "simba/libm_force" ]
 macros = [ "nalgebra-macros" ]
 cuda   = [ "cust_core", "simba/cuda" ]
-rayon-par  = [ "std", "rayon" ]
+rayon  = [ "std", "dep:rayon" ]
 
 # Conversion
 convert-mint = [ "mint" ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,3 +138,5 @@ lto = true
 [package.metadata.docs.rs]
 # Enable all the features when building the docs on docs.rs
 all-features = true
+# define the configuration attribute `docsrs`
+rustdoc-args = ["--cfg", "docsrs"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ libm    = [ "simba/libm" ]
 libm-force = [ "simba/libm_force" ]
 macros = [ "nalgebra-macros" ]
 cuda   = [ "cust_core", "simba/cuda" ]
-par-iter  = [ "std", "rayon" ]
+
 
 # Conversion
 convert-mint = [ "mint" ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,7 +101,7 @@ glam020        = { package = "glam", version = "0.20", optional = true }
 glam021        = { package = "glam", version = "0.21", optional = true }
 glam022        = { package = "glam", version = "0.22", optional = true }
 cust_core      = { version = "0.1", optional = true }
-
+rayon = "1.5" # TODO make this feature gated
 
 [dev-dependencies]
 serde_json = "1.0"
--- a/src/base/iter.rs
+++ b/src/base/iter.rs
@@ -6,8 +6,6 @@ use std::iter::FusedIterator;
 use std::marker::PhantomData;
 use std::mem;
 
-use rayon::iter::plumbing::Producer;
-
 use crate::base::dimension::{Dim, U1};
 use crate::base::storage::{RawStorage, RawStorageMut};
 use crate::base::{Matrix, MatrixView, MatrixViewMut, Scalar};
@@ -365,34 +363,6 @@ impl<'a, T: Scalar, R: Dim, C: Dim, S: 'a + RawStorage<T, R, C>> ExactSizeIterat
     }
 }
 
-impl<'a, T, R: Dim, Cols: Dim, S: RawStorage<T, R, Cols>> Producer for ColumnIter<'a, T, R, Cols, S>
-where
-    T: Send + Sync + Debug + PartialEq + Clone + 'static,
-    S: Sync,
-{
-    type Item = MatrixSlice<'a, T, R, U1, S::RStride, S::CStride>;
-    type IntoIter = ColumnIter<'a, T, R, Cols, S>;
-
-    fn split_at(self, index: usize) -> (Self, Self) {
-        // the index is relative to the size of this current iterator
-        // it will always start at zero so it serves as an offset
-        let left = Self {
-            mat: self.mat,
-            range: self.range.start..(self.range.start + index),
-        };
-
-        let right = Self {
-            mat: self.mat,
-            range: (self.range.start + index)..self.range.end,
-        };
-        (left, right)
-    }
-
-    fn into_iter(self) -> Self::IntoIter {
-        self
-    }
-}
-
 /// An iterator through the mutable columns of a matrix.
 #[derive(Debug)]
 pub struct ColumnIterMut<'a, T, R: Dim, C: Dim, S: RawStorageMut<T, R, C>> {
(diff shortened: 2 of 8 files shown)


src/base/iter.rs
```diff
@@ -370,10 +416,11 @@ impl<'a, T, R: Dim, C: Dim, S: 'a + RawStorageMut<T, R, C>> Iterator
...
 
     #[inline]
-    fn next(&mut self) -> Option<Self::Item> {
-        if self.curr < self.ncols() {
-            let res = unsafe { (*self.mat).column_mut(self.curr) };
-            self.curr += 1;
+    fn next(&'_ mut self) -> Option<Self::Item> {
+        if self.range.start < self.ncols() {
+            let pmat: *mut _ = self.mat;
+            let res = unsafe { (*pmat).column_mut(self.range.start) };
```
Andlon, line 422: From a brief look, I suspect both the original impl and the new impl might be UB due to provenance issues. Basically here we're creating a temporary mutable reference to the matrix, creating a derived matrix slice and returning it. Because the internal pointers stored in the matrix slice will have provenance derived from the temporary reference, I don't think this is technically legal. It would be great for someone more knowledgeable in provenance / UB to comment on this.

Andlon (commented):
I'm not able to give a complete review right now, but I wanted to say that it's a good idea to investigate integrating `rayon` for things like this. Absolutely should be feature-gated though.

From a brief look I have some concerns about the existing column iter impl, and therefore the new one since it inherits the same possibly problematic code. See comment. It might be that it's not a problem at all but I think it might be worth a closer look.

src/base/iter.rs
geo-ant: Hey, thanks a lot for taking a look so quickly. I had the same fears concerning the sequential iterator but unfortunately I don't know enough yet to fix it for the general case.

I wonder if I can do that for matrices with contiguous storage by using slices to access the underlying data. I might try implementing a special iterator only for those types of matrices. But there's probably something I'm not seeing that would make it hard again. Also I'd be interested in a general solution.

geo-ant: I really habe to read up on the whole provenance business. I can surely change the logic back to the original pointer. I changed that around because that way I tricked the compiler into deriving `Send` for me, but I think I should be fine deriving it unsafely...

From the Rustonomicon:
> Something can safely be Send unless it shares mutable state with something else without enforcing exclusive access to it. [link](https://doc.rust-lang.org/nomicon/send-and-sync.html)

I think the iterator does enforce exclusive access (not to the whole matrix but to the memory locations that can be manipulated). But I'm too tired to reason about that right now :) I'll take a look tomorrow. Let me know what you think, when you have time.

geo-ant: I really habe to read up on the whole provenance business. I can surely change the logic back to the original pointer. I changed that around because that way I tricked the compiler into deriving `Send` for me, but I think I should be fine deriving it unsafely...

From the Rustonomicon:
> Something can safely be Send unless it shares mutable state with something else without enforcing exclusive access to it. [link](https://doc.rust-lang.org/nomicon/send-and-sync.html)

I think the iterator does enforce exclusive access (not to the whole matrix but to the memory locations that can be manipulated). But I'm too tired to reason about that right now :) I'll take a look tomorrow. Let me know what you think, when you have time.

geo-ant:
Hey, I reverted the `ColumnIterMut` implementation back to the original logic and manually implemented `Send`, because (to my understanding, please see above) that is safe to do. Please let me know if that makes you happier :)

src/base/iter.rs
geo-ant: hey again, I've reverted the iterators back to the original pointer logic and taken a stab at feature gating the rayon functionality.

geo-ant:
Hi, would you be so kind to re-run the workflow again, so I can see whether I've fixed the pipeline? There's no rush with the review, but I might have time to at least get the CI green on the weekend :)

src/base/par_iter.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    fn opt_len(&self) -> Option<usize> {
+        Some(self.mat.ncols())
+    }
+}
+
+impl<'a, T, R: Dim, Cols: Dim, S: RawStorage<T, R, Cols>> IndexedParallelIterator
+    for ParColumnIter<'a, T, R, Cols, S>
+where
+    T: Send + Sync + Clone + Debug + PartialEq + 'static,
```
Andlon, line 46: What's the purpose of the `PartialEq` and `Debug` bounds? Maybe just require `Scalar + Send + Sync`? (applies throughout)

src/base/par_iter.rs
```diff
@@ -0,0 +1,156 @@
+//! this module implements parallelators to make matrices work with
+//! the rayon crate seamlessly
```
Andlon, line 2: Some typos and typographical things here. Maybe replace with

> Parallel iterators for matrices compatible with `rayon`.

src/base/par_iter.rs
```diff
@@ -0,0 +1,156 @@
...
+
+impl<T, R: Dim, Cols: Dim, S: RawStorage<T, R, Cols> + RawStorageMut<T, R, Cols>>
+    Matrix<T, R, Cols, S>
+where
+    T: Send + Sync + Clone + Debug + PartialEq + 'static,
+    S: Sync,
+{
+    /// Mutably iterate through the columns of this matrix in parallel using rayon
+    pub fn par_column_iter_mut(&mut self) -> ParColumnIterMut<'_, T, R, Cols, S> {
+        ParColumnIterMut::new(self)
```
Andlon, line 154: I think it would be nice to include an example in the docs here for how the user can use this functionality in practice. Same goes for `par_column_iter`.

src/base/iter.rs
```diff
@@ -396,6 +418,96 @@ impl<'a, T: Scalar, R: Dim, C: Dim, S: 'a + RawStorageMut<T, R, C>> ExactSizeIte
...
+            };
+            (left, right)
+        }
+
+        fn into_iter(self) -> Self::IntoIter {
+            self
+        }
+    }
+
+    impl<'a, T: Scalar, R: Dim, C: Dim, S: 'a + RawStorageMut<T, R, C>> Producer
```
Andlon, line 475: Note: The fact that `Producer` is implemented for `ColumnIter`/`ColumnIterMut` will be part of the public API. I'm not sure if this is desirable or not. It seems to me that this ought to be only an implementation detail used in the `ParallelIndexedIterator` impl. Something to think about anyway.

tests/core/matrix.rs
```diff
@@ -1136,3 +1136,124 @@ fn omatrix_to_string() {
...
+    let cloned = dmat.clone();
+    // test that correct columns are iterated over
+    dmat.par_column_iter().enumerate().for_each(|(idx, col)| {
+        assert_eq!(col, cloned.column(idx));
+    });
+    // test that a more complex expression produces the same
+    // result as the serial equivalent
+    let par_result: f64 = dmat.par_column_iter().map(|col| col.norm()).sum();
+    let ser_result: f64 = dmat.column_iter().map(|col| col.norm()).sum();
+    assert_eq!(par_result, ser_result);
```
Andlon, line 1221: Maybe repeat this test for the mutable case?

tests/core/matrix.rs
```diff
@@ -1136,3 +1136,124 @@ fn omatrix_to_string() {
...
+}
+
+#[test]
+#[cfg(feature = "rayon")]
+fn colum_iteration_mut_double_ended() {
```
Andlon, line 1226: typo: `colum`

Andlon (commented):
This is a preliminary review - I've looked over it and it looks mostly good to me, but I am also not intimately familiar with the iterator impls in `nalgebra`, so I think someone like @sebcrozet should have a look over it to make sure there are not important considerations to take into account.

In general, I think the implementation looks sensible. The docs are a bit too minimal and could use a bit more polish (proper punctuation and capitalization, for example) and a couple of examples. Please see my comments.

I think it would also be nice to advertise that this functionality is available, maybe [in the root module](https://docs.rs/nalgebra/latest/nalgebra/). We should also enable the appropriate features on `docs.rs` (see `Cargo.toml` for the special meta tag for `docs.rs`) to make sure that docs are built with these features.

Cargo.toml
```diff
@@ -33,6 +33,7 @@ libm    = [ "simba/libm" ]
 libm-force = [ "simba/libm_force" ]
 macros = [ "nalgebra-macros" ]
 cuda   = [ "cust_core", "simba/cuda" ]
+rayon-par  = [ "std", "rayon" ]
```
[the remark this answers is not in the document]
geo-ant: Completely agree this would be more elegant. I tried this last time and I can achieve that by writing the dependency as `"dep:rayon"`. Last PR this broke the test on one of the architectures. Most likely because the compiler was out of date. If it's ok to bump the MSRV for that feature I'd also prefer the more elegant way of doing that. Let me know what you think

src/base/par_iter.rs
geo-ant: Oh heck, this was a classic case of what Matt Godbolt calls Compiler-Error-Driven-Development XD. What I should have said was `Scalar` instead of the Confusing `Clone + Debug + PartialEq`. This took me a minute to figure out. I'll fix it.

src/base/par_iter.rs
geo-ant: done.

src/base/iter.rs
geo-ant: I'm in two minds about this. The other way I can see to solve this is to create basically a 1-to-1 code duplicate of the iterators that are not part of the public API. I don't like code duplication, but I can also see how you might want to hide the implementation. I'll go whichever way you like and maybe there is a more elegant solution that I am not seeing right now.

tests/core/matrix.rs
geo-ant: fixed

tests/core/matrix.rs
geo-ant: I had to look it up myself, but I think I did that. The test is called `parallel_column_iteration_mut()` and is located around line 1250. Are you unhappy with the test for the mutable iteration?

geo-ant: I had to look it up myself, but I think I did that. The test is called `parallel_column_iteration_mut()` and is located around line 1250. Are you unhappy with the test for the mutable iteration?

src/base/par_iter.rs
geo-ant: I'll try and add some nicer documentation.

*TODO* @geo-ant

geo-ant: I've added better documentation (so I hope) and examples. I've also advertized the feature on the main page of the `Matrix` struct using an additional section.

geo-ant:
Hi, could you kick off the ci again so I see check if I need to fix a failure?

src/base/iter.rs
Andlon: I think you could create a private wrapper type around the iterator and implement the `Producer` for that. That should result in fairly minimal code duplication I think?

tests/core/matrix.rs
Andlon: That test appears to do something completely different, if I'm not mistaken?

Andlon: I mean, that test is also useful, but testing both behaviors seems like a good idea?

geo-ant: ok sure, I'll add another test

src/base/iter.rs
geo-ant: Great idea. You mean like a newtype, right?

Andlon: That's right!

tests/core/matrix.rs
geo-ant: Hey thanks for the suggestion, I've appended the `parallel_column_iteration_mut` test to cover mutable iterators as well. That made me find a bug and I've added another test `column_iterator_double_ended_mut` that covers that bug.

src/base/iter.rs
geo-ant: Okay, I refactored this into a separate producer type that is not publicly visible. Let me know if you like that implementation and thanks for all the improvements thus far.

geo-ant:
Hello, would you kindly kick of the CI again :) ?

geo-ant:
Hey, I looking at the CI error it does not seem like something I can fix:

```
/usr/bin/sudo /opt/hostedtoolcache/cuda_installer-linux/11.2.2/x64/cuda_installer_11.2.2.run --silent --toolkit --samples
 Failed to verify gcc version. See log at /var/log/cuda-installer.log for details.
Total size of all the files uploaded is 211 bytes
```
Am I missing something or did the CI just have a hiccup?

Also are you happy with the changes I made?

geo-ant:
Hey, I looking at the CI error it does not seem like something I can fix:

```
/usr/bin/sudo /opt/hostedtoolcache/cuda_installer-linux/11.2.2/x64/cuda_installer_11.2.2.run --silent --toolkit --samples
 Failed to verify gcc version. See log at /var/log/cuda-installer.log for details.
Total size of all the files uploaded is 211 bytes
```
Am I missing something or did the CI just have a hiccup?

Also are you happy with the changes I made?

Cargo.toml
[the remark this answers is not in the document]
sebcrozet: Using `dep:rayon` isn’t possible right now since the `cuda` version requires a very specific (older) version of the compiler that doesn’t understand that annotation. Nonetheless, we don’t need to make the `par-iter` imply `std` (because `nalgebra` itself won’t need the `std` feature enabled to compile with `par-iter` enabled, and `rayon` itself will have `std` enabled which it needs to wore).

So I suggest we just use `rayon` as the feature name (and don’t declare explicitly a feature like `par-iter`).

sebcrozet:
Thank you for this PR @geo-ant! I think it’s in good shape now. I made a few small changes and rebased on the `dev` branch.
As suggested in my last comment, I removed the `par-iter` feature so we just use `rayon` as the feature name.

geo-ant:
Hey @sebcrozet thanks for the changes and @Andlon for the reviews. Concerning the feature name, I think I did not outright call it rayon because it made some part of the CI fail that depended on an older Rust version. Let me know if you need me to do more things to the PR :)

sebcrozet:
@geo-ant No more change needed, the CI passes now. Thank you again!

geo-ant:
Thanks you both again, it was really fun working on the project and I think I'll do that again some time :)

Would you think it's time to close #848 ?

Andlon:
@geo-ant: thanks for persevering. It's a great new feature. I'm sorry for my tardiness in reviewing/responding, I've been totally swamped recently.

geo-ant:
Hey @Andlon, no worries at all, I assumed work and life got in the way. I'm really happy to see the feature merged and your reviews made it better.

PR #1171, opened 2022-10-27
Added `try_cast()` to `Matrix`.

md2468:
Just like the `cast()` method in `Matrix` is the equivalent of `convert()` with more concise syntax, it would be convenient to have a method that enables the less restrictive conversions of `try_convert()` in the same way.


sebcrozet (approved):

sebcrozet:
Thanks!

PR #1173, opened 2022-11-04
Fix lower bound typo

### LINKED ISSUE #1164: Typo in documentation for quaternion axis_angle
"The rotation axis and angle in ]0, pi] of this unit quaternion."
Was reading over the documentation, and it seems like "]0, pi]" should be "[0, pi]"
https://github.com/dimforge/nalgebra/blob/2f1f4414e23f64edab4a51961898986df18bec8f/src/geometry/quaternion.rs#L1326

allan2:
All instances of this typo fixed.
Resolves #1164.

### DIFF
--- a/src/geometry/quaternion.rs
+++ b/src/geometry/quaternion.rs
@@ -1323,7 +1323,7 @@ where
         }
     }
 
-    /// The rotation axis and angle in ]0, pi] of this unit quaternion.
+    /// The rotation axis and angle in [0, pi] of this unit quaternion.
     ///
     /// Returns `None` if the angle is zero.
     ///
--- a/src/geometry/rotation_specialization.rs
+++ b/src/geometry/rotation_specialization.rs
@@ -877,7 +877,7 @@ impl<T: SimdRealField> Rotation3<T> {
         }
     }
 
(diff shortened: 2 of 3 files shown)


sebcrozet:
Thank you for the PR. However the `]0, pi]` bounds are correct because these methods return `None` if the angle is 0.

allan2:
What about `(0, pi]`? Parantheses are more commonly used for exclusive bounds.

Andlon:
Have to agree with @allan2, I've never seen the `]0,pi]` notation used outside of `nalgebra`. Upon some quick research, it seems that this is notably [used by French speakers](https://hsm.stackexchange.com/questions/142/why-is-american-and-french-notation-different-for-open-intervals-x-y-vs-x/193#193), but the English speaking world prefers `(0, pi]`. I was raised in Norway and I've also never seen the "French version". It seems to me that `(0, pi]` would be more universally understood?

Of note, [Wikipedia appears to use `(0, 1]`](https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals).

sebcrozet:
Ok, switching to `(0, pi]` makes sense!

PR #1176, opened 2022-11-06
replace matrixmultiply with gemm

sarah-ek:
this PR replaces the `matrixmultiply` dependency with `gemm`, which is a crate i developed for high performance matrix multiplication. its single threaded performance is about 10-20% better than `matrixmultiply` for medium/large matrices. and its multithreaded performance is competitive with [blis](https://github.com/flame/blis) and intel mkl (and about 10% faster than eigen on my machine)

the multithread parallelism isn't implemented in this PR. since i'm not sure how to expose it in `nalgebra` (global variable, thread_local, per call basis?)

`gemm` also has a `nightly` feature that enables avx512 for extra performance, which is also not implemented in this PR since i don't know how we want to expose it either.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ path = "src/lib.rs"
 
 [features]
 default = [ "std", "macros" ]
-std     = [ "matrixmultiply", "simba/std" ]
+std     = [ "gemm", "simba/std" ]
 sparse  = [ ]
 debug   = [ "approx/num-complex", "rand" ]
 alloc   = [ ]
@@ -80,7 +80,7 @@ approx         = { version = "0.5", default-features = false }
 simba          = { version = "0.7", default-features = false }
 alga           = { version = "0.9", default-features = false, optional = true }
 rand_distr     = { version = "0.4", default-features = false, optional = true }
-matrixmultiply = { version = "0.3", optional = true }
+gemm           = { version = "0.11", optional = true }
 serde          = { version = "1.0", default-features = false, features = [ "derive" ], optional = true }
 rkyv           = { version = "~0.7.1", optional = true }
 bytecheck      = { version = "~0.6.1", optional = true }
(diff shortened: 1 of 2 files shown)


sarah-ek:
matmul benchmark results on my machine after enabling `gemm`:

```
mat100_mul_mat100       time:   [39.299 µs 39.346 µs 39.398 µs]                               
                        change: [-4.7112% -4.5478% -4.3846%] (p = 0.00 < 0.05)
                        Performance has improved.
mat500_mul_mat500       time:   [3.8708 ms 3.8769 ms 3.8843 ms]                               
                        change: [-12.788% -12.574% -12.344%] (p = 0.00 < 0.05)
                        Performance has improved.
```

sarah-ek:
the CI seems to be failing on cuda but i'm not sure what's causing it. it says it can't find `aligned-vec = "0.5"` but it's right there in crates.io https://crates.io/crates/aligned-vec

Andlon:
I forgot to mention, but having more contributors run the benchmarks on their computers would also be helpful, since these low-level kernels might be highly sensitive to the hardware, and I'm sure you've tested it most extensively on your own hardware @sarah-ek. I'm out of time right now but I'll try to do it in due time.

sarah-ek:
so, gemm is based on the BLIS papers, which can be found here https://github.com/flame/blis#citations
it's more or less a faithful implementation of the algorithms described in the fourth paper.

i'm aware the documentation is sparse at the moment, and i'm willing to put in the work to improve it. i have a more extensive benchmark suite on the gemm repository, and i can post the results on my machine. i can also make it so that it's easier for other people to benchmark the code. i'm also curious how it performs on machines other than my desktop and laptop (both x86_64), and i'm open to suggestions regarding that

as for the unsafety of the code. i run a subset of my tests in miri to make sure there's no UB. i'm also willing to write more tests and make sure all the code paths are covered.

sarah-ek:
these are all great ideas. im not familiar with how CI works but i can use this as a learning opportunity. having it as an experimental feature also sounds good to me. as for the `dep:` issue, i can take care of it

PR #1178, opened 2022-11-14
Rename slice to view

### LINKED ISSUE #1076: Consider renaming matrix slice to "matrix view"
# Motivation

A "matrix slice" is indeed analogous to Rust slices (`&[T]`), so the name isn't inherently problematic. However, the terminology quickly becomes ambiguous. For example, consider the following methods from `nalgebra`'s API:

- [Matrix::slice](https://docs.rs/nalgebra/latest/nalgebra/base/struct.Matrix.html#method.slice) returns a `MatrixSlice`
- [Matrix::as_slice](https://docs.rs/nalgebra/latest/nalgebra/base/struct.Matrix.html#method.as_slice) returns `&[T]`
- [Matrix::copy_from_slice](https://docs.rs/nalgebra/0.30.1/nalgebra/base/struct.Matrix.html#method.copy_from_slice) constructs a matrix from a Rust slice `&[T]`.

There are many more examples in the API, I'm just trying to point out how our current terminology leads to ambiguous naming. I therefore suggest that we rename our current `slice` types to `view`, e.g.:

`MatrixSlice(Mut)` -> `MatrixView(Mut)`

Similarly, the methods that *actually* return a `MatrixSlice(Mut)` would be renamed from `*slice*` to `*view*`.

I am however not only motivated by fixing up inconsistencies: I'm sorely in need of a method that, given any matrix, gives me a matrix slice. The reason is that I have many APIs that take e.g. a `DMatrixSlice(Mut)`:

```rust
fn compute_something<T>(input: DMatrixSlice<T>, output: DMatrixSliceMut<T>);
```
and calling these methods is often a bit of a pain, e.g.
```rust
compute_something(DMatrixSlice::from(&input), DMatrixSliceMut::from(&output));
```
With the new terminology we could provide a method named `as_view(_mut)` that would simply return the same matrix as a view:
```rust
compute_something(input.as_view(), output.as_view_mut());
```

# Implementation

In order to avoid breaking *everybody*'s code, it looks to me as if it would be plausible to do the rename in a fairly smooth way:

- Rename current slice types to `*View(Mut)`.
- Provide type aliases for the existing `Slice` names, e.g. `MatrixSlice = MatrixView`.
- Duplicate existing `_slice_` methods that work with matrix slices as `_view_`
- Mark the existing `_slice_` methods as `deprecated`.
- Update documentation and user guide (a bit of work probably).

I think this way we don't get direct breakage but users will just receive deprecation warnings. I also believe for most users the upgrade is a simply manner of Search + Replace (e.g. `MatrixSlice` -> `MatrixView`). Then in a few releases' time we can remove the deprecated methods.

I'd be happy to make the changes necessary, but I'd like to get some feedback first: are there compelling reasons against this change? Should we bikeshed over naming? (though I struggle to imagine there's a better name than `View`, but I may be surprised!) Open to feedback!

Since this is a somewhat sweeping change, it would be great to have your thoughts on this @sebcrozet. (I think we've briefly discussed it on Discord in the past)

Andlon:
Resolves #1076.

This PR changes terminology from "slice" to "view" as described in #1076. I've followed the general strategy described there. This means that essentially no user should experience breakage, and instead be met with deprecation warnings. The only breaking change that I'm aware of is the fact that the `SliceRange` trait was renamed to `DimRange`. However, I've provided a "trait alias" in the form of a trait `SliceRange` with a blanket impl, which should mitigate most potential problems. This is still technically a breaking change, but I don't think there are many - if any - users that directly work with the `SliceRange` trait in such a way that they would experience breakage.

In essence, I have:
- renamed `*MatrixSlice*` types to `*MatrixView*`
- added immediately deprecated type aliases with the old names, so that code using e.g. `DMatrixSlice` still compiles but gives a deprecation warning.
- similarly, all functions like `slice`, `fixed_slice` etc. have been deprecated and just call their `view` counterparts `view`, `view_slice` and so on.
- `SliceRange` has been renamed (in the manner described above) to `DimRange`.
- I have deprecated `MatrixSliceMutMN` as it's definition was anyway already the same as `MatrixSliceMut`, as well as `MatrixSliceMutN` as there is no other alias with this kind of naming, and I believe it is a remnant of the "N" traits we used to have in the past.
- Added new methods `Matrix::as_view` and `Matrix::as_view_mut`, which are very useful when working with view-based APIs. See the doc examples for those methods for further motivation, as well as the brief description in #1076.

Reviewing this PR is presumably difficult since it consists of a large amount of search-and-replace-like changes. Therefore I've tried to preempt this issue to some degree by not using any automatic large-scale search-and-replace. I've manually verified every single replacement as I went along, but I cannot rule out that I have made mistakes nonetheless.

After the CI has run, I will try this branch in [fenris](https://github.com/InteractiveComputerGraphics/fenris), which has a complex generic codebase built on top of `nalgebra` to see if it uncovers any mistakes (i.e. any kind of failure other than deprecation warnings). I would also be grateful if others would be willing to try the same with their code bases to check that they don't run into any unexpected problems.

To get an overview of the changes, it might be useful to look at the commit messages, as they outline a step-by-step process that I have followed.


Andlon:
Oh, and let's not forget that the user guide needs to be updated to reflect the changes here! I can try to get around to that tomorrow. Are there other concerns that I have not thought of?

Andlon:
CI failure appears unrelated... A proptest fluke?

Andlon:
@sebcrozet: it would be great with some basic instructions for how to build the website/user guide. I've done so in the past but since I never work with these JS/node-based tools I feel like I'm starting a reverse engineering process from scratch again... Can you give me a hint on how to get started?

Andlon:
OK, I can confirm that `fenris` appears to work fine with this branch. Though `fenris` is complex, it still doesn't exercise anything close to the full extent of the slice API in `nalgebra`, so it is by no means an exhaustive check.

sebcrozet:
Thanks for this PR! Regarding building the user-guide: running `yarn install` and `yarn start` on the `nalgebra.org` repo should do the trick.

Andlon:
I just filed a [PR on the nalgebra.org repository](https://github.com/dimforge/nalgebra.org/pull/73) that updates the user guide with the new terminology.

sebcrozet:
Fantastic work ! Thanks @Andlon.

PR #1180, opened 2022-11-24
Enable `reshape_generic` for slices + other things

Andlon:
This PR consists of 3 commits, which can be split out as different PRs as necessary:

- Add `U0, U1, ...` constants alongside the `U0, U1, ...` types. This lets us write `U4` instead of `U4::name()` or `Const::<4>` when we need const dimensions. It used to be like this before we got const generics, but it got lost in the transition. I don't think there's any implied breakage from this.
- Rename `Dynamic` to `Dyn` and make `Dyn` a tuple struct. 
  - This hasn't really been discussed much before, but it's a simple change so I thought I'd give it a go and see what people think.
  - This means we can write `Dyn(3)` instead of `Dynamic::new(3)`, which is very verbose.
  - It also makes error messages like `Matrix<T, _, Dynamic, VecStorage<...>>`  more concise, which might make them easier to read.
  - To avoid breakage, a type alias `Dynamic = Dyn` is also provided.
  - We might consider deprecating `Dynamic`.
- Implement `ReshapableStorage` for matrix slices (only for unit strides at the moment), which was missing.
  - This means we can use `slice.reshape_generic(...)` on slices as well, something I've needed many times.
  - I discovered that there were no real tests for the existing `reshape_generic` functionality, so I added some.

This PR can either be merged before #1178, in which case I'd update #1178, or it can be merged afterwards, in which case I'll update this PR with the new view terminology.

### DIFF
--- a/examples/reshaping.rs
+++ b/examples/reshaping.rs
@@ -2,7 +2,7 @@
 
 extern crate nalgebra as na;
 
-use na::{DMatrix, Dynamic, Matrix2x3, Matrix3x2, Const};
+use na::{DMatrix, Dyn, Matrix2x3, Matrix3x2, Const};
 
 fn main() {
     // Matrices can be reshaped in-place without moving or copying values.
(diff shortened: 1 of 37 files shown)


Andlon:
CI failure seems to be due to a spurious failure of an eigen proptest.

Andlon:
Now that #1178 was merged, I will endeavor to update this PR soon. @sebcrozet: any thoughts on whether you would want to merge all 3 parts of this PR, or if I should leave some out (either for a separate PR or not at all, if the changes are not desirable).

sebcrozet:
@Andlon Keeping the 3 parts of this PR is fine. I agree with deprecating `Dynamic`.

sebcrozet:
@Andlon I took the liberty to update the PR and deprecate `Dynamic` in favor of `Dyn`. I’m merging this now.
Thank you again for this PR!

Andlon:
@sebcrozet: that's amazing, thank you! Pretty swamped with work and life at the moment, so haven't been able to dedicate any time to this. Great to see it merged and in the new release :D

PR #1188, opened 2022-12-21
Update isometry.rs

Spodeian:
Fix formatting of documentation so that "A 3D isometry is composed of:" is on its own line.


sebcrozet:
Thanks!

PR #1192, opened 2023-01-11
Add ln_determinant to Cholesky

Shatur:
This PR implements a function that returns natural logarithm of determinant similar to the one in [NumPy](https://numpy.org/doc/stable/reference/generated/numpy.linalg.slogdet.html) and in [ndarray-linalg](https://docs.rs/ndarray-linalg/latest/ndarray_linalg/cholesky/trait.DeterminantC.html#tymethod.ln_detc).

### DIFF
--- a/src/linalg/cholesky.rs
+++ b/src/linalg/cholesky.rs
@@ -2,6 +2,7 @@
 use serde::{Deserialize, Serialize};
 
 use num::One;
+use num::Zero;
 use simba::scalar::ComplexField;
 use simba::simd::SimdComplexField;
 
@@ -41,6 +42,7 @@ where
 impl<T: SimdComplexField, D: Dim> Cholesky<T, D>
 where
     DefaultAllocator: Allocator<T, D, D>,
+    T::SimdRealField: Zero,
 {
     /// Computes the Cholesky decomposition of `matrix` without checking that the matrix is definite-positive.
     ///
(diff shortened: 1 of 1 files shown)


src/linalg/cholesky.rs
```diff
@@ -41,6 +42,7 @@ where
 impl<T: SimdComplexField, D: Dim> Cholesky<T, D>
 where
     DefaultAllocator: Allocator<T, D, D>,
+    T::SimdRealField: Zero,
```
sebcrozet, line 45: This shouldn’t be needed.

src/linalg/cholesky.rs
```diff
@@ -2,6 +2,7 @@
 use serde::{Deserialize, Serialize};
 
 use num::One;
+use num::Zero;
```
sebcrozet, line 5:
```suggestion
use num::{One, Zero};
```

sebcrozet (changes requested):
Thank you for this PR!

src/linalg/cholesky.rs
Shatur: But I need to create `T::SimdRealField::zero()`. It won't work with `T::zero()`.

sebcrozet: You can call `T::SimdRealField::zero()` without the explicit `T::SimdRealField: Zero` bound.

Shatur: Oh, right! For some reason I thought that it is necessary.

sebcrozet (approved):

sebcrozet:
All good, thanks!

Issue #1204, opened 2023-02-09
Implement Default for all matrix types in `nalgebra` and `nalgebra-sparse`


Andlon:
Currently, `VecStorage` has no `Default` impl, and `CsrMatrix`, `CscMatrix` etc. have no default impls.

AnthonyKalaitzis:
I wouldn't mind giving this one a try. Would it  be preferred that they generate identity matrices or 0 matrices as defaults?

acxz:
I just picked up this library and came across this issue as well.
Just running the following line:
```rust
let x = nalgebra::SVector::<f64, 37>::default();
```
fails with the following:
```
error[E0599]: the function or associated item `default` exists for struct `Matrix<f64, Const<37>, Const<1>, ArrayStorage<f64, 37, 1>>`, but its trait bounds were not satisfied
   --> src\bin\main.rs:136:51
    |
136 |     let x = nalgebra::SVector::<f64, 37>::default();
    |                                                   ^^^^^^^ function or associated item cannot be called due to unsatisfied trait bounds
    |
   ::: C:\Users\acxz\.cargo\registry\src\index.crates.io-6f17d22bba15001f\nalgebra-0.32.3\src\base\array_storage.rs:46:1
    |
46  | pub struct ArrayStorage<T, const R: usize, const C: usize>(pub [[T; R]; C]);
    | ---------------------------------------------------------- doesn't satisfy `ArrayStorage<f64, 37, 1>: Default`
    |
   ::: C:\Users\acxz\.cargo\registry\src\index.crates.io-6f17d22bba15001f\nalgebra-0.32.3\src\base\matrix.rs:175:1
    |
175 | pub struct Matrix<T, R, C, S> {
    | ----------------------------- doesn't satisfy `_: Default`
    |
    = note: the following trait bounds were not satisfied:
            `ArrayStorage<f64, 37, 1>: Default`
            which is required by `Matrix<f64, Const<37>, Const<1>, ArrayStorage<f64, 37, 1>>: Default`   
```

Values up to 32 do work however, looks like for static vectors, default isn't implemented for sizes up to 64.

Would implementing this be feasible and part of the roadplan?

arscisca:
Hello! I haven't checked back on this since the pull request, but yes I implemented the `Default`s for the mentioned structs.

PR #1206, opened 2023-02-13
Implement the Default trait for VecStorage, CsMatrix, CscMatrix, etc.

SoutrikBandyopadhyay:
This commit solves #1204 and implements the Default trait for CsMatrix, CscMatrix, SparsityPattern, VecStorage.

### DIFF
--- a/nalgebra-sparse/src/pattern.rs
+++ b/nalgebra-sparse/src/pattern.rs
@@ -52,6 +52,17 @@ pub struct SparsityPattern {
     minor_dim: usize,
 }
 
+///The default implementation for the SparsityPattern
+impl Default for SparsityPattern {
+    fn default() -> Self {
+        Self {
+            major_offsets: Default::default(),
+            minor_indices: Default::default(),
+            minor_dim: Default::default(),
+        }
+    }
+}
+
 impl SparsityPattern {
     /// Create a sparsity pattern of the given dimensions without explicitly stored entries.
     pub fn zeros(major_dim: usize, minor_dim: usize) -> Self {
--- a/src/base/vec_storage.rs
+++ b/src/base/vec_storage.rs
@@ -31,6 +31,20 @@ pub struct VecStorage<T, R: Dim, C: Dim> {
     ncols: C,
 }
 
+/// The default implementation for the VecStorage struct.
+/// Requires R and C to implement the default trait.
+impl<T, R: Dim + std::default::Default, C: Dim + std::default::Default> Default
+    for VecStorage<T, R, C>
+{
+    fn default() -> Self {
+        Self {
+            data: Default::default(),
+            nrows: Default::default(),
+            ncols: Default::default(),
+        }
+    }
+}
+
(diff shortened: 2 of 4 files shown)


src/base/vec_storage.rs
```diff
@@ -31,6 +31,20 @@ pub struct VecStorage<T, R: Dim, C: Dim> {
...
+{
+    fn default() -> Self {
+        Self {
+            data: Default::default(),
+            nrows: Default::default(),
+            ncols: Default::default(),
+        }
+    }
+}
+
```
Andlon, line 47: I don't think any dimensions *actually* implement `Default` at the moment :thinking: 

I think an alternative implementation could divide this into several cases:
- `R: DimName`, `C: DimName`. In this case we need to require `T: Default`
- `R: DimName`, `C = Dynamic`. In this case we can *avoid* requiring `T: Default` and create an "empty" matrix.
- `R = Dynamic`, `C: DimName`. Same as the previous.
- `R = Dynamic`, `C = Dynamic`. Same, we don't need `T: Default`.

This makes it possible to do
```rust
#[derive(Default)]
struct Wrapper<T> {
   vector: DVector<T>,
   matrix: DMatrix<T>,
}
```
without requiring `T: Default`. This is a common problem I'm frequently running into, at least.

nalgebra-sparse/src/pattern.rs
```diff
@@ -52,6 +52,17 @@ pub struct SparsityPattern {
...
 }
 
+///The default implementation for the SparsityPattern
+impl Default for SparsityPattern {
+    fn default() -> Self {
+        Self {
+            major_offsets: Default::default(),
+            minor_indices: Default::default(),
+            minor_dim: Default::default(),
+        }
```
Andlon, line 62: This is incorrect since `major_offsets` should have length `major_dim + 1`. An "empty" pattern would look like:
```rust
Self {
  major_offsets: vec![0],
  minor_indices: vec![],
  minor_dim: 0
}
```

Andlon (changes requested):
Thanks for this PR, I agree that having more `Default` impls would make life easier, and I regularly run into the lack of these myself.

I'd prefer to have some sanity tests for the `Default` impls where you use the "try_from..." constructors to validate the output against what you expected. The current impl of `CsMatrix` is unsound and would be caught by such a test, see comments.

For the `VecStorage` impl I have an alternative suggestion that I think would be more practically useful, please see comment there as well.

src/base/vec_storage.rs
Andlon: Note also that the proposed impl is unsound, because it always generates an empty vector regardless of whatever the dimensions are. This would lead to unsoundness whenever the dimensions are non-zero and any operation in nalgebra would try to index into the matrix.

tpdickso:
It looks as if these changes with Andlon's recommendations were implemented in https://github.com/dimforge/nalgebra/pull/1312 , so I'll close this pull request.

Issue #1109, opened 2022-05-10
Incorrect symmetric eigendecomposition for a 4x4 f64 matrix (swapped eigenvectors)


acshi:
For a certain 4x4 symmetric matrix, nalgebra does not compute a correct eigendecomposition. The maximum relative error in the recomposed matrix is 3.4 (so 340% off). As far as I can tell, this is because two of the eigenvectors (or eigenvalues, equivalently) are in the wrong order. If we swap them back into the right order, we get a maximum relative error of 4.1e-11. If we instead use peroxide to compute the eigendecomposition, we get a relative error of 1.9e-15.

I've tried looking at the code for the eigendecomposition, but I'm really not familiar enough with the specific algorithms (some kind of implicit QR iteration with various optimizations?) to know how to debug this.

```rust
#[test]
#[rustfmt::skip]
fn symmetric_eigen_issue_4x4() {
    let m = Matrix4::new(
        -19884.07f64, -10.07188, 11.277279, -188560.63,
        -10.07188, 12.518197, 1.3770627, -102.97504,
        11.277279, 1.3770627, 14.587362, 113.26099,
        -188560.63, -102.97504, 113.26099, -1788112.3,
    );

    let swap_wrong_columns = false;
    let use_eigen = false;

    let (evals, evecs) = if !use_eigen {
        let mut eig = m.clone().symmetric_eigen();
        if swap_wrong_columns {
            let (eval1, eval2) = (eig.eigenvalues[1], eig.eigenvalues[2]);
            eig.eigenvalues[1] = eval2;
            eig.eigenvalues[2] = eval1;
        }

        (eig.eigenvalues, eig.eigenvectors)
    } else {
        let eig = eigen(&peroxide::structure::matrix::Matrix {
            data: m.iter().copied().collect(),
            row: 4,
            col: 4,
            shape: peroxide::structure::matrix::Shape::Col,
        });
        let evals = nalgebra::Vector4::from_iterator(eig.eigenvalue.iter().copied());
        let evecs =
            nalgebra::Matrix4::from_iterator(eig.eigenvector.data.iter().copied());
        (evals, evecs)
    };

    eprintln!("evecs: {evecs:.4e}");
    eprintln!("evals: {:.4e}", nalgebra::Matrix::from_diagonal(&evals));

    let recomp = evecs * nalgebra::Matrix::from_diagonal(&evals) * evecs.transpose();
[...]

PR #1210, opened 2023-02-21
RFC: Fix SymmetricEigen Routine

CattleProdigy:
I've got some proposed fixes to address the issue referenced below.

The bug itself is in the the `subdim=2` special case. As best as I can tell, in that block we try to simply replace the unreduced block by its diagonalization, with the eigenvectors (called `basis` in the code)  serving a similar role as a Givens rotation does in the Implicit-Shift QR part of the algorithm. The eigenvector / eigenvalue correspondence is mismatched. Unfortunately, upon fixing that I find the proptests break regularly. The results are correct but not to within the specified threshold. I should also note that the proptests are also failing on `dev` though you have to run it a few times to see a failure.

I'm not sure if there's a better way to salvage the `subdim=2` case or if there's any advantage to doing so.  In any case, I ultimately just removed it and let the Implicit-Shift QR part handle everything.  That passes the test case from the user who reported the issue and proptest passes (100s of times without failing).

For feedback:

- Do we want to keep the user-submitted test case?
- Do we want to try to keep the part I removed?

Addresses https://github.com/dimforge/nalgebra/issues/1109


Andlon:
I haven't yet had time to review this, but I also don't know the specific motivation for the `subdim = 2` special case. @sebcrozet, would you be able to comment on the motivation for this? Is it e.g. significantly faster in some cases important to you?

Andlon:
@CattleProdigy: as I said on Discord, thank you so much for looking into this! I hope to be able to review soon, but in the meantime I'll try to answer one more of your questions. I think we definitely want to keep *all* user-submitted test cases in the test suite, with a clear reference to the issue in question, e.g. `symmetric_eigen_regression_issue_1109()` or similar (in a way that aligns with the current naming conventions for the other `SymmetricEigen` tests, I haven't looked it up).

CattleProdigy:
> @CattleProdigy: as I said on Discord, thank you so much for looking into this! I hope to be able to review soon, but in the meantime I'll try to answer one more of your questions. I think we definitely want to keep _all_ user-submitted test cases in the test suite, with a clear reference to the issue in question, e.g. `symmetric_eigen_regression_issue_1109()` or similar (in a way that aligns with the current naming conventions for the other `SymmetricEigen` tests, I haven't looked it up).

I've updated the name of the test and the docstrings to match the other issue/regression tests

CattleProdigy:
@Andlon @sebcrozet 

I've decided to adopt the solution that disrupts the code the least: maintaining the subdim=2 case. I synced my branch with `dev`. It passes the regression test and the `linalg::eigen` proptests. It should be ready to review and merge.

Andlon:
Thanks for working on this @CattleProdigy, this is really great. I am not at all familiar with the code here, so given the subtlety of eigenvalue routines, I don't feel quite comfortable merging it right away, although of course the passing tests give great confidence to the solution here. Still, @sebcrozet, I think you were the original author of these routines, would you be able to take a look?

DevonMorris:
@Andlon @sebcrozet any update on merging this?

sebcrozet:
That’s looking good, thanks! My apology for not checking this out sooner.

PR #1213, opened 2023-02-27
Support conversion for glam 0.23

MarijnS95:


devil-ira (approved):

bugQ:
looks like the only reason this doesn't pass is #1216 eh

sebcrozet:
Thanks!

MarijnS95:
Thank _you_ for already pushing this into a release!

PR #1216, opened 2023-03-06
Fix CI by removing unnecessary parentheses

geieredgar:
This fixes the CI breaking because nightly now correctly detects these unnecessary parentheses.


sebcrozet:
Thanks!

PR #1219, opened 2023-03-09
Allow trailing punctuation in macros

ModProg:
When splitting e.g. fields over multiple lines it is quite common to add a
trailing comma in the last line in rust code.

This pr makes it so that this is allowed now:

```rust
vector![
    some_longer_expression() + something,
    some_longer_expression() + something_else,
    some_longer_expression() + another_thing,
    some_longer_expression() + 42,
];
```

I also added a test verifying that this was already the case with `matrix!`.


sebcrozet:
Thanks! Yeah, using proc-macro makes the code simpler overall.

PR #1220, opened 2023-03-14
Serialize statically sized matrices as tuples

Ralith:
Breaking change. Consistent with how serde serializes plain arrays. Saves a length tag in common serialization formats (e.g. bincode, punchcard).


sebcrozet:
Thanks!

PR #1227, opened 2023-04-08
fix: remove Scalar trait bound for Matrix PartialEq and Eq

hqurve:
I believe that the Matrix PartialEq implementation does not require that T implements Scalar.

Removing this bound would allow for users to derive PartialEq on custom generic structures without requiring that T: Scalar. For example
```rust
#[derive(PartialEq)]
struct MyCoolStructure<T>{
     m: Vector2<T>
}
```

### DIFF
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -1859,14 +1859,14 @@ where
 
 impl<T, R: Dim, C: Dim, S> Eq for Matrix<T, R, C, S>
 where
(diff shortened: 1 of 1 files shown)


Ralith (approved):

Andlon:
I can't think of anything negative consequences of this, and I like the justification.

Andlon (approved):

sebcrozet:
Thanks!

PR #1235, opened 2023-04-22
Add euler_angles_ordered function on Rotation

Masterzach32:
At the moment, Rotation only has a way to get Euler angles in the order roll, pitch, yaw. This PR adds a euler_angles_ordered function to Rotation which allows the user to specify any order of Euler axes. It also allows the user to specify if the returned angles should be for an intrinsic or extrinsic rotation.

The implementation is based upon this paper:
Malcolm D. Shuster, F. Landis Markley, “General formula for extraction the Euler angles”, Journal of guidance, control, and dynamics, vol. 29.1, pp. 215-221. 2006

This is a function I wrote in one of my own projects, and thought it would be a good addition to the library.

### DIFF
--- a/src/geometry/rotation_specialization.rs
+++ b/src/geometry/rotation_specialization.rs
@@ -983,7 +983,64 @@ impl<T: SimdRealField> Rotation3<T> {
     /// Represent this rotation as Euler angles.
     ///
     /// Returns the angles produced in the order provided by seq parameter, along with the
-    /// observability flag. If the rotation is gimbal locked, then the observability flag is false.
+    /// observability flag. The Euler axes passed to seq must form an orthonormal basis. If the
+    /// rotation is gimbal locked, then the observability flag is false.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the Euler axes in `seq` are not orthonormal.
+    ///
+    /// # Example 1:
+    /// ```
+    /// use std::f64::consts::PI;
+    /// use approx::assert_relative_eq;
+    /// use nalgebra::{Matrix3, Rotation3, Unit, Vector3};
+    ///
+    /// // 3-1-2
+    /// let n = [
+    ///     Unit::new_unchecked(Vector3::new(0.0, 0.0, 1.0)),
+    ///     Unit::new_unchecked(Vector3::new(1.0, 0.0, 0.0)),
+    ///     Unit::new_unchecked(Vector3::new(0.0, 1.0, 0.0)),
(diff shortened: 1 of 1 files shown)


src/geometry/rotation_specialization.rs
```diff
@@ -979,6 +979,113 @@ impl<T: SimdRealField> Rotation3<T> {
...
+    /// Algorithm based on:
+    /// Malcolm D. Shuster, F. Landis Markley, “General formula for extraction the Euler
+    /// angles”, Journal of guidance, control, and dynamics, vol. 29.1, pp. 215-221. 2006,
+    /// and modified to be able to produce extrinsic rotations.
+    #[must_use]
+    pub fn euler_angles_ordered(
+        &self,
+        mut seq: [Unit<Vector3<T>>; 3],
+        extrinsic: bool,
+    ) -> (Vector3<T>, bool)
```
sebcrozet, line 997: Let’s return `([T; 3], bool)` since the  input vectors is an array.

src/geometry/rotation_specialization.rs
```diff
@@ -979,6 +979,113 @@ impl<T: SimdRealField> Rotation3<T> {
...
+
+    /// Represent this rotation as Euler angles.
+    ///
+    /// Returns the angles produced in the order provided by seq parameter, along with the
+    /// observability flag. If the rotation is gimbal locked, then the observability flag is false.
+    ///
+    /// Algorithm based on:
+    /// Malcolm D. Shuster, F. Landis Markley, “General formula for extraction the Euler
+    /// angles”, Journal of guidance, control, and dynamics, vol. 29.1, pp. 215-221. 2006,
+    /// and modified to be able to produce extrinsic rotations.
```
sebcrozet, line 991: Could you please add a couple of examples (as doc-tests) here? Perhaps the ones presented in the papier.
In addition, does `seq` have to match some properties (being an orthonormal basis)? If so, that should be mentioned in the documentation here.

sebcrozet (changes requested):
That’s very interesting! Thank you for this contribution.

src/geometry/rotation_specialization.rs
Masterzach32: Yes, the axes must be orthonormal (satisfying n1 dot n2 = 0 and n3 dot n1 = 0). I added asserts for that as well

sebcrozet (approved):

sebcrozet:
Thanks!

Issue #1218, opened 2023-03-09
Matrix::variance() incorrect results when all elements are the same


bushkov:
If all elements in a matrix are the same I would expect that `Matrix::variance()` would give a result close to 0.0. And it works for f64 values that don't have a fractional part:
```rust
    let v = DVector::repeat(10_000, 100000000.0);
    assert_eq!(v.variance(), 0.0)
```
But it gives results which are far from 0.0, once a fractional part is introduced:
```rust
    let v = DVector::repeat(10_000, 100000000.815068511);
    assert_eq!(v.variance(), 0.0)
```
```shell
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `2258.0`,
 right: `0.0`', src/main.rs:5:5
```

Simpler variance implementation gives more precise results which are closer to the true variance value of 0.0:
```rust
    let v = DVector::repeat(10_000, 100000000.815068511);
    let m = v.mean();
    let variance = v.fold(0.0, |acc, e| acc + (e - m) * (e - m)) / 10_000.0;
    assert_eq!(variance, 0.0)
```
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `2.1201040922846914e-11`,
 right: `0.0`', src/main.rs:7:5
```

sebcrozet:
That makes sense. The current implementation is being more efficient since we only iterate through the vector once.
Changing it to a more numerically stable version sounds good to me.

jamesneb:
I would like to try implementing this. Can anyone point me to the relevant sections of the codebase?

vasilNnikolov:
I started fixing the issue, hopefully will be done with a first draft in 24 hours.

bushkov:
Fixed in v0.32.3.

PR #1236, opened 2023-04-24
Fix bug 1218

vasilNnikolov:
A proposed fix to bug [1218](https://github.com/dimforge/nalgebra/issues/1218). The bug was a typical case of [Catastrophic cancellation](https://en.wikipedia.org/wiki/Catastrophic_cancellation), where the variance was computed as $E[X^2] - E[X]^2$ and not as $E[(X-E[X])^2]$. The bug is fixed now, the only downside I see is the usage of two folds over the array, one of which is for computation of the mean of the matrix entries $E[X]$. Looking forward to any feedback.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,6 +111,7 @@ serde_json = "1.0"
 rand_xorshift = "0.3"
 rand_isaac = "0.3"
 criterion = { version = "0.4", features = ["html_reports"] }
+nalgebra = { path = ".", features = ["debug", "compare", "rand", "macros"]}
 
 # For matrix comparison macro
 matrixcompare = "0.3.0"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ documented here.
 
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## Unreleased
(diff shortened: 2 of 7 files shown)


wisp3rwind:
Wouldn't it be better to implement something like [Welford's algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm)?

Cargo.toml
```diff
@@ -111,6 +111,7 @@ serde_json = "1.0"
 rand_xorshift = "0.3"
 rand_isaac = "0.3"
 criterion = { version = "0.4", features = ["html_reports"] }
+nalgebra = { path = ".", features = ["debug", "compare", "rand", "macros"]}
```
sebcrozet, line 114: That’s an interesting trick!

sebcrozet (approved):

sebcrozet:
Thanks!

Cargo.toml
vasilNnikolov: Thanks. Did it because rust-analyzer was giving errors whenever I opened the `tests` folder, since it only parses the code reachable with the default features

PR #1248, opened 2023-06-19
Remove unnecessary normalization in `Rotation3::face_towards`

AlexanderEkdahl:
`zaxis` and `xaxis` are already normalized which means `zaxis.cross(&xaxis)` should already be normalized.


sebcrozet:
Thanks!

PR #1251, opened 2023-06-27
Add `Clone` to `MatrixIter` and `MatrixIterMut`

Shatur:
Could be needed for crates like https://docs.rs/logaddexp/latest/logaddexp/.
Other iterators in nalgebra also have `Clone` impl.

### DIFF
--- a/src/base/iter.rs
+++ b/src/base/iter.rs
@@ -17,7 +17,7 @@ use crate::base::{Matrix, MatrixView, MatrixViewMut, Scalar};
 macro_rules! iterator {
     (struct $Name:ident for $Storage:ident.$ptr: ident -> $Ptr:ty, $Ref:ty, $SRef: ty) => {
         /// An iterator through a dense matrix with arbitrary strides matrix.
-        #[derive(Debug)]
+        #[derive(Clone, Debug)]
         pub struct $Name<'a, T, R: Dim, C: Dim, S: 'a + $Storage<T, R, C>> {
             ptr: $Ptr,
             inner_ptr: $Ptr,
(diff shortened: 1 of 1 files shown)


Andlon:
I don't think it's sound for `MatrixIterMut` to implement `Clone`, since this could allow you to have two separate mutable iterators for the same mutable matrix entries, from which you could recover two mutable references to the same element in the matrix.

For `MatrixIter` I suppose it could be sound. But directly deriving might not be desirable since it will add a `Clone`  bound for every generic parameter. In this case it's better to do a manual impl with only the bounds you actually need, I think.

Shatur:
Makes sense, will do!

Andlon:
> But we already have `Clone` for `RowIter` and `ColumnIter`.

Ah, I didn't realize that. That's quite unfortunate, and that's essentially a mistake. We should rather fix those `Clone` impls than perpetuate the same mistake for `MatrixIter` IMO.

Shatur:
Are you sure that it's a mistake? What issues current `Clone` derive causes?

sebcrozet:
Directly deriving `Clone` sounds fine. All the other type parameters already implement `Clone` anyway, and `T: Scalar`  already implies `T: Clone`. So I would be in favor of keeping the clone derive, and relaxing the requirement late if we actually find a use-case that doesn’t fit.

Shatur:
@sebcrozet thanks for letting me know!

Should I add additional parameter to `iterator!` and pass `Clone, Debug` for `MatrixIter` and just `Debug` for MatrixIterMut?

sebcrozet:
@Shatur Yeah, I think that would make sense.

Shatur:
@sebcrozet done!

sebcrozet:
Thanks!

PR #1265, opened 2023-08-06
docs: Use intradoc links rather than HTML.

waywardmonkeys:
This fixes almost all HTML links to be intradoc links that Rust can verify during `cargo doc`. This will help prevent future broken links.


sebcrozet:
Fantastic! Thanks!

PR #1266, opened 2023-08-10
docs: Improve view alias docs.

waywardmonkeys:
* Indicate whether they are immutable/mutable clearly.
* Link to the other form (immutable link to mutable, mutable to immutable).
* Consistently include the text about it being an alias and to look elsewhere for the methods.


Andlon:
Thanks, this is very useful!

Issue #734, opened 2020-05-30
Docs right handed typo


vadixidav:
https://docs.rs/nalgebra/0.21.0/nalgebra/geometry/struct.Isometry.html#method.look_at_lh

It says "right handed", but it should say "left handed".

waywardmonkeys:
This cn be closed now.

PR #1270, opened 2023-08-12
doc: Isometry's `look_at_lh` is left-handed.

waywardmonkeys:
Fixes issue #734.


Andlon:
Thanks! @Ralith: would you be able to vouch for the correctness of this change? I'm just not familiar with these geometry APIs and coordinate conventions is not my strong suit. I'm quite sure this is the correct change but I'm queasy about merging it without confirmation from a long-term user with in-depth knowledge.

sebcrozet:
This looks correct to me. Thank you @waywardmonkeys for the PR and @Andlon for the review!

PR #1271, opened 2023-08-12
Use `assert_relative_eq!` instead of relative_eq!`.

waywardmonkeys:
When testing for something, need to use the assert form.

### DIFF
--- a/src/geometry/dual_quaternion.rs
+++ b/src/geometry/dual_quaternion.rs
@@ -97,7 +97,7 @@ where
     ///
     /// let dq_normalized = dq.normalize();
     ///
-    /// relative_eq!(dq_normalized.real.norm(), 1.0);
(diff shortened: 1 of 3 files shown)


Andlon:
Thanks!

It strikes me that these examples also use the very old Rust style involving `extern crate` and no macro imports. Do you think you could write up a brief issue for transitioning to `use` macro imports in examples, @waywardmonkeys? That would be much appreciated! Only if you have time, of course :-)

waywardmonkeys:
@Andlon On my list already! That was how I found these. :) But I was wary of how big the change might be and how it would fit into feeding bite-sized pieces of things for review / merge. I'll write an issue this week.

Andlon:
Makes sense. I really like these small changes as they're very easy for me to review and merge, and I can slot that in between my other work to a limited extent (while waiting for a compile or simulation to complete :joy:)

PR #1273, opened 2023-08-14
Use std::hint::black_box consistently.

waywardmonkeys:
This also removes the `#![feature(bench_black_box)]`. This was stabilized in Rust 1.66 and anyone building benchmarks will be on that or later (as they previously would have been on nightly).

This also allows building `cargo build --all-targets` on stable Rust as it no longer dies when hitting the feature addition in the benchmarks.

### DIFF
--- a/benches/core/matrix.rs
+++ b/benches/core/matrix.rs
@@ -142,7 +142,7 @@ fn iter(bench: &mut criterion::Criterion) {
     bench.bench_function("iter", move |bh| {
         bh.iter(|| {
             for value in a.iter() {
-                criterion::black_box(value);
+                std::hint::black_box(value);
             }
         })
     });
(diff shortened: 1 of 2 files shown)


Andlon:
Thanks, this will be a very useful change. I'm not entirely sure about merging this right away, as this means that we cannot run benchmarks against older Rust versions (despite needing nightly, we could use an old version). I'm not sure how relevant this is though - @sebcrozet do you have any opinions on this? My feeling is that we should probably just merge this, since I don't think we'll need to consider performance pre 1.66 anyway, but I don't have a full overview of use cases, especially with respect to the CUDA work etc.

Andlon:
Hmm, are benchmarks even built in CI...? :thinking:

sebcrozet:
Thanks! Merging this should be fine. The CUDA  version won’t attempt to compile/run benchmarks.

PR #1277, opened 2023-08-14
`split_at` is only used by Rayon code.

waywardmonkeys:
This fixes an unused code warning. Since the code is `pub(crate)`, it was only available within the crate and only used by Rayon code, so it is no functional change to not compile it except when the right feature is enabled.

Also, fix a minor typo in some non-doc comments.


Andlon:
Test failure appears to be a flaky test (sigh). Looks good to me though, thanks!

Issue #1125, opened 2022-07-01
Native Gauss-Jordan Elimination


sciguyryan:
Hi.

Just a quick question. Is there a native implementation of the Gauss-Jordan elimination method (or something similar)? I'm working on a chemical equation balancer, and I'm using RREF (via Gauss-Jordan elimination) as a stage in that process.

I went through the effort of writing my own Matrix implementation before I was pointed in this direction... and yours is far better so I'm going to swap it out. Fortunately it was not a waste of time since I learned a lot by implementing it.

If anyone could point me in the right direction, that would be fantastic.

sciguyryan:
> Are you sure you _really_ need RREF? If you do, I suppose there is no implementation in `nalgebra`. If there is a different way to solve your problem, then perhaps there is something you can use.

Possibly not, I'm using the method I am familiar with, but there are likely others that can work too. I'll try to summarize that I would typically do here.

Take all of the terms on the right and left hand side and tokenize them. For each unique symbol (H for Hydrogen, He for Helium and so on.) count the number of instances of each of those symbols. Put that into a matrix. For example, with the following equation:

Ca(OH)2 + H3PO4 ->Ca3(PO4)2 + H20

The unique tokens are Ca, O, H and P, indexed 0 to 3 respectively.

From that we can build a matrix of the number of instances of each symbol (terms from the right-hand side must be negative), as follows:

{ [ 1, 2, 2, 0], [0, 4, 3, 1], [-3, -8, 0, -2], [0, -1, -2, 0] }

From there I would use RREF to reduce the matrix, which allows it to be solved (by computing the null-space of the matrix, or at least that is how I was taught to do it). In this specific example the end result would be something like this: { [1/2], [1/3], [1/6], [1] }. Identifying the LCM will tell tell you the coefficients for each of the components of the chemical equation.

There might be another way to achieve that result, but I don't know what that would be.

Makogan:
> There certainly is no implementation of one in `nalgebra`. In numerical linear algebra, Gauss elimination is usually supplanted by the LU decomposition (which is in fact Gaussian elimination but we build the matrix factors in the process).
> 
> To my knowledge, (reduced) row echelon form is something that is very rarely done numerically (rather it's mostly used as a teaching tool, for doing computations by hand). Are you sure you _really_ need RREF? If you do, I suppose there is no implementation in `nalgebra`. If there is a different way to solve your problem, then perhaps there is something you can use.

[...]

PR #1280, opened 2023-08-15
Implement reduced row echelon form function

Makogan:
I am trying to address this issue https://github.com/dimforge/nalgebra/issues/1125

I have run into needing a rref about 3 times by now so I figured implementing it in nalgebra would help everyone one out.


Andful:
I need this too. The decomposition methods don't work from num-rational and are not exact.

PR #1281, opened 2023-08-18
nalgebra-glm: Fix clippy single_component_path_imports.

waywardmonkeys:
These imports are redundant and not needed.


sebcrozet:
Thanks!

PR #1283, opened 2023-08-18
DualQuaternion: Fix to_vector self convention.

waywardmonkeys:
By taking a ref, we can avoid an extra copy on the caller side.


sebcrozet:
Thanks!

PR #1287, opened 2023-08-21
Implement `Clone` for `CsrTripletIter` and `CscTripletIter`

mohe2015:


nalgebra-sparse/src/csc.rs
```diff
@@ -611,7 +611,7 @@ fn pattern_format_error_to_csc_error(err: SparsityPatternFormatError) -> SparseF
 }
 
 /// Iterator type for iterating over triplets in a CSC matrix.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
tpdickso, line 614: I'm wondering... is this more conservative than it needs to be? I think using `derive` here will force a `Clone` bound on `T`. But since this is just cloning the internal state of an iterator over immutable references, that bound is unnecessary, right? i.e. we could instead have
```rust
impl<'a, T> Clone for CscTripletIter<'a, T> {
  fn clone(&self) -> Self {
    CscTripletIter {
      pattern_iter: self.pattern_iter.clone(),
      values_iter: self.values_iter.clone()
    }
  }
}
```
And this would provide a broader correct implementation. It might not be *that* useful to have a CscMatrix of non-`Clone` types but it seems like it may as well be supported here.

tpdickso (changes requested):
LGTM! One small comment.

nalgebra-sparse/src/csc.rs
mohe2015: I confirmed that and therefore adjusted the code.

tpdickso (approved):

PR #1288, opened 2023-08-22
Add `try_from_triplets_iter`

### LINKED ISSUE #1286: Sparse COO Matrix `try_from_triplets` could have `from_iterator`
Currently the sparse COO_matrix requires passing 3 vectors, for row indices, col indices and values.
I was wondering if it would make sense to have a function signature like below, which does the construction of each vector,
while also guaranteeing their sizes are the same.
 `some_fn_name(rows: usize, cols: usize, impl IntoIterator<Item=(usize, usize, T)>) -> CooMatrix`

JulianKnodt:
Calls `try_from_triplets` for now, and is mentioned in the documentation.

Closes #1286

### DIFF
--- a/nalgebra-sparse/src/coo.rs
+++ b/nalgebra-sparse/src/coo.rs
@@ -160,6 +160,25 @@ impl<T> CooMatrix<T> {
         }
     }
 
+    /// Try to construct a COO matrix from the given dimensions and a finite iterator of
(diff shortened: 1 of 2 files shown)


sebcrozet:
Thanks @JulianKnodt for the PR and @Andlon for the review!

PR #1302, opened 2023-09-30
DOC: Use norm_squared() in its doctest

mgeier:
This looks like copy-pasted from `magnitude_squared()`.


sebcrozet:
Thanks! CI failure is unrelated to this PR.

Issue #1249, opened 2023-06-26
Different license in crates and LICENSE file


klukola:
Different license in crates and LICENSE file. In crates is still BSD3 (as in Cargo.toml file line 13) and in LICENSE file is Apache License 2.0.
Can you please update license in Cargo.toml?

BatmanAoD:
I think #1308 resolved this.

Issue #972, opened 2021-08-19
LICENSE issues


ThatGeoGuy:
Hi! 

I noticed that nalgebra switched it's licensing a year ago to Apache 2.0. 

However, the `Cargo.toml` still specifies `BSD-3`. This makes tooling like `cargo license` output the following:

```
BSD-3-Clause (1): nalgebra
```

Now, I don't have a problem with either of these licenses, but it's pretty confusing. Can this be updated in the Cargo.toml to make it more clear what license we should be using with this software?

willu:
It looks like this was changed from Apache-2.0 back to BSD-3 in this commit: https://github.com/dimforge/nalgebra/commit/7d5cc4912dc8928e5d476d75a6b9f430468b0cdb

But it looks like some of the sub-crates never had their license changed from BSD-3 to Apache-2.0 in the earlier license change.  I also couldn't find much documentation on the earlier license change.  Were all contributors contacted, or did someone just decide to change the license?

Issue #912, opened 2021-06-14
include license and readme file in crate source of nalgebra-macros


remilauzier:
Hi. i am in the proccess of packaging nalgebra-macros in fedora and they would like to have the license and the readme include in the crate source.

https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/
https://docs.fedoraproject.org/en-US/packaging-guidelines/LicensingGuidelines/#_license_text

BatmanAoD:
I don't know if this issue is fixed for Fedora, but I think it still persists for `cargo` installs.

The tarball for `nalgebra-macros` (as of the latestversion, 0.2.2) does include a `LICENSE` file, but rather than containing the license text, it contains the symlink path, `../LICENSE`. This can be verified either by [direct download](https://static.crates.io/crates/nalgebra-macros/0.2.2/download) or using `cargo`:

```
cargo add nalgebra-macros
cargo check
cat ~/.cargo/registry/src/xxxxx/nalgebra-macros-0.2.2/LICENSE
```

decathorpe:
> rather than containing the license text, it contains the symlink path, ../LICENSE

This is often a problem when running `cargo package` / `cargo publish` on an OS that does not support symbolic links (i.e. Windows), which might explain why it works for some people and doesn't for others.

BatmanAoD:
@decathorpe Do you know how `nalgebra`'s cargo-publish CI is set up? I would be surprised if it's run on Windows for some reason. (Also, I'm kind of surprised to hear that `cargo publish` doesn't support symbolic links, since you _can_ configure `git` to use symbolic links in Windows.)

decathorpe:
I don't know - though it looks like publishing to crates.io isn't automated in CI.

> (Also, I'm kind of surprised to hear that cargo publish doesn't support symbolic links, since you can configure git to use symbolic links in Windows.)

The problem isn't that `cargo publish` doesn't support symbolic links - the problem is that support for symbolic links needs to be configured on both the Windows side *and* in git for them to work correctly for this use case AFAIK - and its off by default in both places.

PR #1308, opened 2023-10-24
Fix and clarify license in crate metadata and add missing license files

decathorpe:
I am responsible for packaging some crates from this project for Fedora Linux, and have noticed that there are some issues with the license metadata / missing license files in published crates. I see that there are also reported issues for this (#912, #972, #1249), and attempted to fix the issues reported there, and some issues I noticed myself:

---

It appears that the project was intentionally relicensed from BSD-3-Clause to Apache-2.0 in commit https://github.com/dimforge/nalgebra/commit/d81a895a87e2e5a94a847c8033a7c2cce2fbed9a, but this was, at least in part, not updated in crate metadata. This results in crates.io and tools like cargo-license to show the wrong license for some crates in this project. To address this, I adapted the `package.license` field in Cargo.toml files for the `nalgebra`, and `nalgebra-glm` accordingly.

---

The `nalgebra-lapack` crate is MIT licensed (there is a license file for the MIT license included, and the documentation agrees), but the crate metadata claims the crate is `BSD-3-Clause` licensed. Maybe this is a copy-paste error from when the crate was originally imported? To address this discrepancy, I changed the `package.license` field in Cargo.toml from `BSD-3-Clause` to `MIT`.

---

One of the terms of the Apache-2.0 license is that any (re)distributed sources contain a copy of the license text. This is currently not the case for the `nalgebra-glm`, `nalgebra-macros`, and `nalgebra-sparse` crates.  This is problematic for Linux distributions that package these crates, and potentially also for publishing them on crates.io. Adding a symbolic link to the LICENSE file from the project's root directory should cause `cargo package` / `cargo publish` to properly include this file when publishing.


sebcrozet:
Your analysis checks out. Thanks!

decathorpe:
Great - thank you! 🎉

decathorpe:
It looks like new releases of this project are currently getting pushed from the dev-0.32.x branch. Would it be possible to cherry-pick this change onto that branch as well, so that it actually starts showing up in releases?

PR #1309, opened 2023-10-25
Implement gradient method on Matrix

lukerob01:
I've been using nalgebra in a robotics project and needed to implement a function for approximating the gradient of a matrix similar to `numpy.gradient` so I have added it here as a pull request to try and contribute back to the awesome project.

I've tried to make it as generic as possible but I am not quite sure how the storage and allocator traits work so I've done my best by looking at other similar methods (if someone has a resource and can point me in the right direction to learn this I would be very grateful).

I also wasn't sure what the standard here is for testing so I haven't implemented any tests yet but would be happy to provide some tests if told where to put them.

P.S. this is my first ever contribution to open source so apologies in advance for any bits I've left out

### DIFF
--- a/src/base/componentwise.rs
+++ b/src/base/componentwise.rs
@@ -351,3 +351,91 @@ impl<T: Scalar, R1: Dim, C1: Dim, SA: Storage<T, R1, C1>> Matrix<T, R1, C1, SA> 
         }
     }
 }
+
+// Calculus
+impl<T: Into<f64> + Copy, R1: Dim, C1: Dim, SA: Storage<T, R1, C1>> Matrix<T, R1, C1, SA>
+where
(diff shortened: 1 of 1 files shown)


Andlon:
Hi @lukerob01, thanks for this PR! From a cursory glance, it looks well done. However, I think perhaps this is slightly out of scope for `nalgebra`, which is primarily focused on linear algebra and (Euclidean) geometry. The gradient of a matrix is not really a well-defined concept; from what I can tell, what's actually going on is that you have made some assumptions about what adjacent element in a matrix mean with respect to some function that was sampled on a regular grid. I think this might be better suited for a library on function approximation, numerical solutions of differential equations, or in a library for domain-specific functionality.

With that in mind, I'll close this PR as being out of scope.

Issue #1313, opened 2023-11-06
Incorrect results from pseudo_inverse


Ralith:
The matrix constructed by
```
na::DMatrix::from_column_slice(3, 8, &[10.0f32, 6.123234e-16, 20.0, 0.0, 10.0, -20.0, 0.0, 0.0, 0.0, 0.0, 10.0, 20.0, -10.0, 6.123234e-16, 20.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0])
```

produces a `pseudo_inverse` of approximately:

```
  ┌                      ┐
  │  0.050 -0.000 -0.013 │
  │  0.000 -0.050  0.013 │
  │  0.000 -0.000 -0.000 │
  │  0.000 -0.050 -0.013 │
  │ -0.050 -0.000 -0.013 │
  │  0.000 -0.000  0.000 │
  │  0.000 -0.000  0.000 │
  │  0.000 -0.000  0.000 │
  └                      ┘
```
If `f32` is replaced with `f64`, *or* if the entries having value `6.123234e-16` are zeroed out, `pseudo_inverse` instead yields approximately:
```
  ┌                      ┐
  │  0.050  0.000  0.013 │
  │  0.000  0.050 -0.013 │
  │  0.000  0.000  0.000 │
  │  0.000  0.050  0.013 │
  │ -0.050  0.000  0.013 │
  │  0.000  0.000  0.000 │
  │  0.000  0.000  0.000 │
  │  0.000  0.000  0.000 │
  └                      ┘
```
Note that the signs of the second and third columns are inverted. This result is the expected one. This is a massive change for a tiny perturbation of the input, and is breaking the algorithm I have built around this operation. Is this expected?

rasmusgo:
The result is not a proper pseudo inverse for this matrix. Here is a test that fails for the given matrix:
```rs
#[test]
fn test_pseudo_inverse() {
    let s = 6.123234e-16;
    // let s = 0.0;
    let m = nalgebra::DMatrix::from_column_slice(
        3,
        8,
        &[
            10.0f32, s, 20.0, 0.0, 10.0, -20.0, 0.0, 0.0, 0.0, 0.0, 10.0, 20.0, -10.0, s, 20.0,
            0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
        ],
    );
[...]

PR #1314, opened 2023-11-09
Fix bug in SVD related to values near zero

rasmusgo:
This resolves issue #1313.

The problem was in the bidiagonalization where U and V is computed. There is a corner case where U D V doesn't agree how the sign should be handled when the reflection axis has zero magnitude.

This solution fixes the correctness but I would like some input on how to make the test for zero magnitude more efficient.


Ralith:
This indeed fixes my issue :tada:

sebcrozet:
Thank you for this PR!
I pushed a small change: using `norm_squared` that is marginally more efficient than `magnitude` because it skips a `sqrt`.

PR #1315, opened 2023-11-12
Allow creating matrix iter with an owned view

### LINKED ISSUE #747: Implement `IntoIter` for `Matrix<N, R, C, S>` where `S: IntoIter`, etc.
I can whip up a PR for this if it's considered a desired feature.

This would make it easier to write iterators that flatten matrices, where some or all of these matrices may be temporary objects that need to be moved into the iterator so as not to be referencing a temporary. For example:
```rust
// How it can be written currently
let vertex_data = as.iter().zip(bs.iter()).flat_map(|(a, b)|
    a.cross(&b).data.into_iter()
).collect::<Vec<f32>>();

// How it could be written afterward
let vertex_data = as.iter().zip(bs.iter()).flat_map(|(a, b)|
    a.cross(&b).into_iter()
).collect::<Vec<f32>>();
```
I think this is a useful change, because having to dig into the matrix storage requires a deeper knowledge of the implementation details of `Matrix`, and is *much* less discoverable since there is no implementation of `IntoIter` for `Matrix` and solving it requires you to make the leap to the (hidden-by-default) `data` member to know that it does have an `IntoIter` implementation (which you can currently only find by going sideways through `DefaultAllocator` to `ArrayStorage`'s `Deref` impl to `GenericArray`, manually googling it since the docs don't hyperlink types from external crates.)

Additionally, the current autoref to the `IntoIter` impl for `&'a Matrix` is unintuitive. If a user knows they have an owned `MatrixMN`, `VectorN` or `PointN`, it's intuitive that they should be able to convert it to an owned iterator, and it's confusing for it to borrow instead.

However, it would be a **breaking change**. The rust language itself is phasing in a similar breaking change where `[T; N].into_iter()` takes ownership of the array instead of auto-ref'ing to a slice iterator, so there is precedent for this.

yotamofek:
Fixes #747 

First of all - thanks for this crate, it's really fun to use, except for the lack of `IntoIterator` impls on (owned) view storages.

This allows overcoming the borrow checker issues when trying to do something like this:
image

Using the `IntoIterator` impl from this PR:
image

I implemented this feature by extracting a common subset of `MatrixIter` and `MatrixIterMut` into a `RawIter` struct which erases the lifetime it is created with, and iterates over raw pointers instead of references.
Then, the `MatrixIter(Mut)` structs become simple wrappers over `RawIter`, but with an added lifetime constraint.

It's technically possible to keep the old code (and remove most of the code added in this PR) and impl the new `new_owned` and `new_owned_mut` using a transmute from `MatrixIter<'outer, ...>` to a `MatrixIter<'inner, ...>`, but I'm not even sure that's not possibly UB.

Let me know what you think :)

### DIFF
--- a/src/base/conversion.rs
+++ b/src/base/conversion.rs
@@ -98,6 +98,18 @@ impl<'a, T: Scalar, R: Dim, C: Dim, S: RawStorage<T, R, C>> IntoIterator
     }
 }
 
+impl<'a, T: Scalar, R: Dim, C: Dim, RStride: Dim, CStride: Dim> IntoIterator
+    for Matrix<T, R, C, ViewStorage<'a, T, R, C, RStride, CStride>>
+{
+    type Item = &'a T;
+    type IntoIter = MatrixIter<'a, T, R, C, ViewStorage<'a, T, R, C, RStride, CStride>>;
+
+    #[inline]
+    fn into_iter(self) -> Self::IntoIter {
+        MatrixIter::new_owned(self.data)
+    }
+}
+
 impl<'a, T: Scalar, R: Dim, C: Dim, S: RawStorageMut<T, R, C>> IntoIterator
     for &'a mut Matrix<T, R, C, S>
 {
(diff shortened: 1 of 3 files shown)


sebcrozet:
Thanks!

PR #1321, opened 2023-11-17
Forward `std` feature to some deps.

### LINKED ISSUE #623: Some `std` features of external crates aren't re-exported
Some functions, for example [`Complex::to_polar`](https://www.nalgebra.org/rustdoc/nalgebra/struct.Complex.html#method.to_polar), are listed in the docs, but since they are under the `std` feature of external crates (e.g. `num-complex`) and not re-exported by `nalgebra` these methods can't actually be used.

zachs18:
Fixes #623 (specifically about `Complex::from_polar` at least)

`nalgebra`'s `std` cargo feature now also enables the `std` feature of some dependencies that use `default-features = false` (`num-traits`, `num-rational`, `num-complex`, `approx`, (if enabled) `alga`)

(Note: the  `"alga?/std"` syntax for enabling features of optional dependencies without necessarily enabling the optional dependency itself was only stabilized in Rust 1.60, but `nalgebra` seems to state that it's MSRV policy is ["latest stable"](https://docs.rs/nalgebra/latest/nalgebra/#using-nalgebra) so that would be fine.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ path = "src/lib.rs"
 
 [features]
 default = [ "std", "macros" ]
-std     = [ "matrixmultiply", "simba/std" ]
+std     = [ "matrixmultiply", "num-traits/std", "num-complex/std", "num-rational/std", "approx/std", "simba/std", "alga?/std" ]
 sparse  = [ ]
 debug   = [ "approx/num-complex", "rand" ]
 alloc   = [ ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ path = "src/lib.rs"
 
 [features]
 default = [ "std", "macros" ]
-std     = [ "matrixmultiply", "num-traits/std", "num-complex/std", "num-rational/std", "approx/std", "simba/std", "alga?/std" ]
+std     = [ "matrixmultiply", "num-traits/std", "num-complex/std", "num-rational/std", "approx/std", "simba/std" ]
 sparse  = [ ]
 debug   = [ "approx/num-complex", "rand" ]
 alloc   = [ ]


Ralith (approved):
Thanks!

Ralith:
Looks like the CUDA test in CI needs to be updated to support the optional dependency feature syntax. Want to have a look at that?

zachs18:
It looks like `rust-CUDA`/`cuda_std` (which the `nalgebra/cuda` feature enables via `simba/cuda`) currently only supports `nightly-2021-12-04` (https://github.com/Rust-GPU/Rust-CUDA/pull/98), which is too old to support the `?/` syntax. Recent nightlys fail since `cuda_std` uses nightly features which have been removed. (https://github.com/zachs18/nalgebra/actions/runs/7161897944/job/19498133305 running on `nightly-2023-12-10`)

I don't see a way to resolve this with both `rust-CUDA` support (in CI or otherwise) and  `alga?/std` in Cargo.toml.

I guess `alga?/std` can be removed, and `alga` will just not have the `std` feature enabled by `nalgebra`'s `std` feature? The others (`num-traits`, `num-complex`, `num-rational`, `approx`, and `simba`) will still have their `std` features enabled. Does that sound fine?

zachs18:
Note that the cuda CI job will fail anyway as it has before due to version resolution https://github.com/dimforge/nalgebra/actions/runs/6839550246/job/18605831607 (probably related to https://github.com/rust-lang/cargo/issues/10623?)

Ralith:
Ah, yeah, looks like Rust-CUDA is de facto unmaintained :(

I saw some mention of https://github.com/coreylowman/cudarc, maybe we should port to that? But that's out of scope here.

> I guess alga?/std can be removed, and alga will just not have the std feature enabled by nalgebra's std feature? The others (num-traits, num-complex, num-rational, approx, and simba) will still have their std features enabled. Does that sound fine?

Sounds good to me. I'd prefer to drop CUDA support and hit all the deps, but that's @sebcrozet's call, and in the mean time this is a strict improvement. Alga users can enable std on that dep themselves if they need it.

Ralith (approved):

sebcrozet:
@Ralith Updating CUDA to the latest rustc/llvm version is a very difficult piece of work so it will probably not happen any time soon (if ever). So let’s just remove CUDA support. Users of cuda can pin the dependency of nalgebra that supported it.

PR #1324, opened 2023-11-21
add  implementation of From trait for SMatrix<->DMatrix, SVector<->DVector

Lishen1:


Andlon:
Thanks for the PR @Lishen1, I think this kind of functionality is sorely needed in `nalgebra`. Converting between different storages often requires different approaches depending on what needs to be done. A more unified approach to conversion would be great.

That said, I think the current approach is a little bit too narrow. For example, it implements directly on `DMatrix<T>`, which excludes e.g. `DVector<T>`.

I wonder if we instead could do something like `From` and `TryFrom` impls that just forward to the underlying `Storage`, and then implement the proper conversion routines for storages instead. Essentially, we'd have `TryFrom<VecStorage> for ArrayStorage` and `From<ArrayStorage> for VecStorage>` assuming that dimensions satisfy something like `DimEq` (or ideally `SameNumberOfRows`/`SameNumberOfColumns` if possible, for more granular type mismatch errors).

Do you have any thoughts on this?

Lishen1:
probably we can combine implementation for vector and matrix into one implementation like: 
```
impl<'a, T: Scalar, const R0: usize, const C0: usize, R1: Dim, C1: Dim>
    From<&'a Matrix<T, R1, C1, VecStorage<T, R1, C1>>>
    for Matrix<T, Const<R0>, Const<C0>, ArrayStorage<T, R0, C0>>
{
    fn from(_m: &'a Matrix<T, R1, C1, VecStorage<T, R1, C1>>) -> Self {
    ...
    }
}
```
i don't know how we can use SameNumberOfRows/SameNumberOfColumns since R1: Dim, C1: Dim both can be Dyn and we can't proof shape match in compile time.
anyway, i'll think how to improve implementation with yor suggestions

PR #1331, opened 2023-12-13
remove unnecessary trait bounds in matrix conversions.

taketo1024:
Removed some of the unnecessary trait bounds in matrix conversions.

### DIFF
--- a/nalgebra-sparse/src/convert/impl_std_ops.rs
+++ b/nalgebra-sparse/src/convert/impl_std_ops.rs
@@ -29,7 +29,7 @@ where
 
 impl<'a, T> From<&'a CooMatrix<T>> for CsrMatrix<T>
 where
-    T: Scalar + Zero + ClosedAdd,
+    T: Scalar + Zero,
 {
     fn from(matrix: &'a CooMatrix<T>) -> Self {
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Seems reasonable, but I'll let someone who has some idea about sparse matrices verify.

Andlon:
I was surprised to see that we can remove `ClosedAdd` from COO->CSR conversion, since if there are duplicates in the COO format, we add them together. Digging a little into it, I see that [convert_coo_cs](https://github.com/dimforge/nalgebra/blob/a01fa48e33b120fa701bdbb471679f896b2c88aa/nalgebra-sparse/src/convert/serial.rs#L237C10-L237C10), which `convert_coo_csr` calls, somehow calls `combine_duplicates` with `std::ops::Add::add`, yet it has *no* bounds for `Add`. I'm perplexed - what am I missing here?

Andlon:
Ah, `num_traits::Zero` requires `Add`. That is uh, not so obvious IMO. Well, in any case, we can remove these bounds. But since we try to provide the same API for `CsrMatrix and `CscMatrix`, can you please add the same changes to `CsrMatrix` @taketo1024? That would be great!

PR #1336, opened 2023-12-20
Make OPoint call `T::fmt` to respect formatting modifiers

kurtlawrence:
Invokes `T::fmt` rather than `write!`, which allows points to be formatted with modifiers like so:

```rust
let p = Point3::new(1.23, 3.45, 5.67);
assert_eq!(&format!("{p}"), "{1.23, 3.45, 5.67}");
assert_eq!(&format!("{p:.1}"), "{1.2, 3.5, 5.7}");
assert_eq!(&format!("{p:.0}"), "{1, 3, 6}");
```

I was not sure where to place the simple test code, it is currently adjacent in the file.

### DIFF
--- a/src/geometry/point.rs
+++ b/src/geometry/point.rs
@@ -511,12 +511,26 @@ where
 
         let mut it = self.coords.iter();
 
-        write!(f, "{}", *it.next().unwrap())?;
+        <T as fmt::Display>::fmt(it.next().unwrap(), f)?;
 
         for comp in it {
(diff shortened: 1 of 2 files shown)


Ralith:
Nice fix, thanks!

> I was not sure where to place the simple test code, it is currently adjacent in the file.

Maybe tests/geometry/point.rs?

kurtlawrence:
Good call, I've moved it there 🙂

Ralith (approved):
Thanks!

PR #1339, opened 2023-12-24
Allow multiplying matrix by `Complex<>` on the left

tpdickso:
This simply inserts `Complex<f32>` and `Complex<f64>` into the `left_scalar_mul_impl!` macro alongside the primitives' trait impls.


tpdickso:
@Andlon would you happen to know what the CUDA error here is? Looks like it's a lockfile issue but I'm not yet familiar with the CI setup.

PR #1342, opened 2023-12-31
Update syn to `2.0`

yotamofek:
Requires no adaptations.

### DIFF
--- a/nalgebra-macros/Cargo.toml
+++ b/nalgebra-macros/Cargo.toml
@@ -16,7 +16,7 @@ license = "Apache-2.0"
 proc-macro = true
 
 [dependencies]
-syn = { version="1.0", features = ["full"] }
+syn = { version = "2.0", features = ["full"] }
 quote = "1.0"
 proc-macro2 = "1.0"
 


CAD97:
Cargo didn't find `ahash@0.7.7` for some reason. Seems spurious and unrelated to the syn update.

yotamofek:
Ah great, now it failed due to a different spurious and unrelated problem 😅
Let's run CI again..

djc:
Friendly ping? Would be nice to get this merged (and ideally, released) to reduce compile-times for downstream projects?

Andlon:
> Friendly ping? Would be nice to get this merged (and ideally, released) to reduce compile-times for downstream projects.

I agree. However, if I run `cargo tree -p nalgebra --all-features`, there are a few (opt-in) dependencies that still seem to rely on `syn 1.0`. Having both `syn 1.0` and `syn 2.0` in the dependency tree seems strictly worse. Perhaps we can upgrade these too though. It's not clear to me how much effort is required, as I suspect some packages might simply not have been updated to support `syn 2.0`.

Andlon:
Actually, I realized that I hadn't updated dependencies in my dev environment for a while. After running `cargo update`, I see that the only other dependency that has not been updated to use `syn 2.0` is `rkyv`. Since we *anyway* might have both `syn 1.0` and `syn 2.0` in the tree at this point due to `serde` using `2.0`, I think we should go ahead and merge this as-is, and then log an issue for `rkyv` to also upgrade its `syn` version to `2.0`.

djc:
> However, as far as I can tell, we don't actually need a new `nalgebra` release for this. If we release `nalgebra-macros` with a new patch version, then existing `nalgebra` users should get the change through `cargo update`.

Yes, that sounds correct. I think that means you don't even need to change nalgebra to bump the nalgebra-macros dependency.

bash:
Resolves #1403 (I forgot to check PRs before opening an issue ^^)

sebcrozet:
Thank you all for making/reviewing this PR!

PR #1350, opened 2024-01-13
Fix `glm::is_null` epsilon test

CAD97:
The existing implementation compares each component to zero with an epsilon; effectively `glm::all(glm::is_comp_null(v, epsilon))`. This probably isn't the desired semantics when calling `glm::is_null`; rather, we want to determine if the magnitude of the vector is within `epsilon` units of zero. It's the question of circle versus square.

The new behavior matches that of OpenGL Mathematics.


CAD97:
Unfortunately it seems that the existing bound isn't sufficient to call `.norm_squared()`, so this would be an API breaking change.

CAD97:
(Apologies for the force pushes; I initially wrote the patch in GitHub's UI without rustc to yell at me when I got things wrong. It should be fixed and squashed now.)

tpdickso:
Looks like this is correct; [glm computes this for vectors](https://github.com/g-truc/glm/blob/5d73e17e58a7ddb0c6c5ce7a4783a2c60df804d0/glm/gtx/vector_query.inl#L106) differently than for matrices.

Code change LGTM but I'll leave the semver-wrangling to someone with more experience.

sebcrozet:
Thanks!

PR #1356, opened 2024-01-23
Added into_vec method

DumitrescuToma:
By maintaining the code structure and format, I've added a method that directly converts a matrix into a vector by concatenating rows. However, the code is flexible and allows developers to use more algorithms for transforming a matrix into an array (concatenating columns, dividing the matrix into block matrices and concatenating the rows for each submatrix in a specific order, traversing the parallel diagonals if the matrix is squared). The code is memory safe and does some optimization by using get_unchecked. This solves the issue #479.


tpdickso:
Thank you for the contribution! Do you know if this has advantages over `.iter().collect::<Vec<_>>()`?

PR #1357, opened 2024-01-27
Support Glam 0.25 type conversion

Jondolf:
Glam was updated to 0.25 a few weeks ago, so this PR adds type conversion for it through the `glam-0.25` feature. This will be needed for migrating e.g. Bevy physics engines to Bevy 0.13 once it is released.

Let me know if I missed something I need to update!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Add the methods `Matrix1::as_scalar`, `::as_scalar_mut`, `::to_scalar`, `::into_scalar`.
 - Add `Rotation3::euler_angles_ordered`, a generalized euler angles calculation.
 - Add the `glam-0.24` feature to enable conversion from/to types from `glam` v0.24.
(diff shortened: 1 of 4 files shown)


tpdickso:
Thank you for the contribution! LGTM.

tpdickso:
@Ralith @Andlon Sorry for the direct ping -- this looks like an easy PR to merge, but `test-nalgebra` appears to be flaky (I've run it twice and it failed on a `proptest` of two different items.) I was able to pull the commit and run the tests locally and they pass fine. Do you know if this flake rate is expected, and would it be alright to merge in this state, or is there something I should dig into to address it?

Ralith:
Proptest is randomized, right? "Flakes" from it are probably real errors, or at least buggy tests, then. If the tests are related to the changes in this PR, then they should block; otherwise, open an issue for them but don't treat them as blocking for this PR.

tpdickso:
The changes are fully unrelated, so shouldn't be blocking. I'll open an issue. I think the proptests may just have had too tight a tolerance.

PR #1371, opened 2024-03-20
Fix redundant import errors in `nalgebra-glm`

yotamofek:
More fallout from https://github.com/rust-lang/rust/issues/121708
Should make CI green again

### DIFF
--- a/nalgebra-glm/src/ext/quaternion_common.rs
+++ b/nalgebra-glm/src/ext/quaternion_common.rs
@@ -1,4 +1,4 @@
-use na::{self, Unit};
+use na::Unit;
 
 use crate::aliases::Qua;
 use crate::RealNumber;
--- a/nalgebra-glm/src/traits.rs
+++ b/nalgebra-glm/src/traits.rs
@@ -1,7 +1,6 @@
 use approx::AbsDiffEq;
 use num::{Bounded, Signed};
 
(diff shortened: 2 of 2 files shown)


Andlon:
Thanks for figuring out the cause of this @yotamofek! I'm unsure why this compiles though. There's a call to `na::zero` on [Line 7](https://github.com/dimforge/nalgebra/blob/cf44429837fd508ae94d2afa29f0b1f85ae5a998/nalgebra-glm/src/ext/quaternion_common.rs#L13). How does this compile without `na` having been explicitly imported? What am I missing?

Andlon:
Ah, I see; it's because of `extern crate nalgebra as na;` in `lib.rs`. OK, I guess it should be safe to merge as far as I can tell. @tpdickso: you've been looking into this issue. Does this adequately fix the issue or can you think of any lingering problems?

tpdickso:
@Andlon The top-level crate has `extern nalgebra as na;` which is sufficient to put it in the namespace, and is what made the `use na;` in the file redundant in the first place.

Anyway, this is good to merge! I don't expect it to fix the other redundant import issues which were being caused by an "invisible" `use na;` on line 2 of several other files -- but those seem to have resolved themselves, perhaps with a version bump of `xargo` or something.

PR #1379, opened 2024-04-07
Introduce Storage::forget_elements() to fix memory leak in Matrix::generic_resize()

### LINKED ISSUE #1378: Memory leak in generic_resize/reallocate_copy
I believe there is a memory leak arising from `DefaultAllocator::reallocate_copy` and its use in `Matrix::resize_generic`. Anecdotally, I observed memory leakage in the wild in some pre-production code and narrowed its source down to here.

The resulting matrix is created by this line of `resize_generic`:

```rust
let res = unsafe { DefaultAllocator::reallocate_copy(new_nrows, new_ncols, data.data) };
```

In the case where at least one of the dimensions of the input matrix to `resize_generic` is `Dyn`, it is backed (by default) by a `VecStorage`; this is the type of `data.data` in the above line. However, in the case where the output matrix is statically sized, `reallocate_copy` moves the data out of this buffer to the new one, and then forgets the buffer:

```rust
// Safety:
// - We don’t care about dropping elements because the caller is responsible for dropping things.
// - We forget `buf` so that we don’t drop the other elements.
std::mem::forget(buf);
```

I believe the intention here is to ensure that the destructor does not get called on the *elements* of the old matrix, as the ones being used are moved to the new buffer, and the others are explicitly dropped by `resize_generic`; the latter part is, I believe, what is referred to by "the caller is responsible for dropping things." However, the caller cannot be responsible for dropping the buffer itself, as it is consumed by `reallocate_copy`; that is the source of the issue here.

I will try to put together a PR shortly to fix this and ensure that the allocation gets cleaned up, while maintaining memory safety regarding not double-dropping the matrix elements.

nickmertin:
Fixes #1378 

Adds a trait method `Storage::forget()` to clean up the storage while forgetting (i.e., not dropping) the contents. Makes use of this method instead of `mem::forget()` in the implementations of `Reallocator` for `DefaultAllocator`.

### DIFF
--- a/src/base/storage.rs
+++ b/src/base/storage.rs
@@ -150,8 +150,8 @@ pub unsafe trait Storage<T, R: Dim, C: Dim = U1>: RawStorage<T, R, C> {
     where
         DefaultAllocator: Allocator<T, R, C>;
 
-    /// Cleans up the storage without calling the destructors on the contained elements.
-    fn forget(self);
+    /// Drops the storage without calling the destructors on the contained elements.
+    fn forget_elements(self);
 }
 
 /// Trait implemented by matrix data storage that can provide a mutable access to its elements.
--- a/src/base/storage.rs
+++ b/src/base/storage.rs
@@ -149,6 +149,9 @@ pub unsafe trait Storage<T, R: Dim, C: Dim = U1>: RawStorage<T, R, C> {
     fn clone_owned(&self) -> Owned<T, R, C>
     where
         DefaultAllocator: Allocator<T, R, C>;
+
+    /// Cleans up the storage without calling the destructors on the contained elements.
+    fn forget(self);
 }
 
 /// Trait implemented by matrix data storage that can provide a mutable access to its elements.
--- a/src/base/vec_storage.rs
+++ b/src/base/vec_storage.rs
@@ -281,7 +281,7 @@ where
     {
         self.clone()
     }
-    
+
     #[inline]
     fn forget(mut self) {
         // Set length to 0 so element destructors are not called.
@@ -338,7 +338,7 @@ where
     {
         self.clone()
     }
-    
+
     #[inline]
     fn forget(mut self) {
         // Set length to 0 so element destructors are not called.
--- a/src/base/vec_storage.rs
+++ b/src/base/vec_storage.rs
@@ -283,8 +283,13 @@ where
     }
 
     #[inline]
-    fn forget(mut self) {
-        // Set length to 0 so element destructors are not called.
+    fn forget_elements(mut self) {
+        // SAFETY: setting the length to zero is always sound, as it does not
+        // cause any memory to be deemed initialized. If the previous length was
+        // non-zero, it is equivalent to using mem::forget to leak each element.
+        // Then, when this function returns, self.data is dropped, freeing the
+        // allocated memory, but the elements are not dropped because they are
+        // now considered uninitialized.
         unsafe { self.data.set_len(0) };
     }
 }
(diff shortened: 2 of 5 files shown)


Andlon:
Thanks, this is great! I don't have the head space or time to wrap my head around the nuances of the `Reallocator` API here at the moment, but I wanted to ask if the documentation needs to be updated to reflect any possible changes in assumptions made. I don't fully understand them yet myself, I'd need a closer look later to try to verify.

Andlon:
I haven't forgotten this PR, but I haven't yet been able to devote time to it. It came up on Discord though, and @sebcrozet suggested that the `forget` function needs to be `unsafe`, and maybe it should be called `forget_elements` or something along those lines.

I'll try to get back to this soon, but I suspect I'll be quite busy this week.

nickmertin:
I like the name `forget_elements`, I think it gets the behaviour across more accurately than what I put right now.

I don't think it should be `unsafe`, for the [same reason that `mem::forget` isn't `unsafe`](https://doc.rust-lang.org/std/mem/fn.forget.html#safety). Safety does not guarantee that destructors are called, so there is nothing unsafe about calling this function. Marking it as `unsafe` would push obligations of ensuring safety onto everywhere it is called, which could be hard to follow given that it is a trait method and there are several implementations. Perhaps the behaviour and safety attributes just need to be better documented, as is the case for `mem::forget`.

tpdickso (approved):
This looks appropriate to me! And I agree, I think a name like `forget_elements` would be great to help highlight that it's the elements that are being forgotten and not the storage itself (which will be dropped normally absent any other intervention.)

src/base/storage.rs
```diff
@@ -149,6 +149,9 @@ pub unsafe trait Storage<T, R: Dim, C: Dim = U1>: RawStorage<T, R, C> {
     fn clone_owned(&self) -> Owned<T, R, C>
     where
         DefaultAllocator: Allocator<T, R, C>;
+
+    /// Cleans up the storage without calling the destructors on the contained elements.
```
Andlon, line 153: As discussed, perhaps change the name to `forget_elements`.

I think perhaps "cleans up" is a little vague. Perhaps "Drops the storage without ..." ?

src/base/vec_storage.rs
```diff
@@ -281,6 +281,12 @@ where
     {
         self.clone()
     }
+
+    #[inline]
+    fn forget(mut self) {
+        // Set length to 0 so element destructors are not called.
```
Andlon, line 287: Perhaps add a "SAFETY" comment here and explain that `.set_len(0)` always satisfies the invariants of `.set_len`, but the contents are deliberately leaked

(same for other `forget` impl)

the "SAFETY comment" practice is a relatively recent thing, and much of `nalgebra`'s codebase, to my knowledge, predates this convention. But I think we should try to stick with it going forward, whenever it makes sense!

Andlon (changes requested):
You're absolutely right about the fact that `forget` doesn't have to be `unsafe` @nickmertin! I just have a few minor suggestions, then we can get this merged in time for the next release :)

nickmertin:
Sounds good, I will make the necessary changes this weekend.

nickmertin:
I believe it is now good to go, let me know if any other changes are needed!

Andlon:
Thanks @nickmertin, it's ready to merge now. Great to have this bug fixed! I'm not merging immediately because of the breaking change in introducing `forget_elements` to the `Storage` trait. In principle, I think we could merge directly since the `dev` branch already contains unreleased breaking changes, but I'll run it by the others later.

sebcrozet:
Hey! Thank you for this PR @nickmertin, and @Andlon and @tpdickso for reviews.
I merged this work as part of #1416 since I wasn’t able to rebase this PR on top of `dev` for today’s release. My guess is that [Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) wasn’t checked (and I didn’t want to bother the author on a Sunday).

PR #1383, opened 2024-04-18
Fix false row-major comment in doc

derchr:
As already discussed in https://github.com/dimforge/nalgebra.org/pull/86, there is a false statement in the documentation of the RawStorage trait.

In the documentation of RawStorage::stride it was falsely claimed that the storage of the matrix happens in a row-major fashion. In fact it is column-major.


Andlon:
Thanks @derchr!

PR #1384, opened 2024-04-19
Inverting a 4x4 matrix with `try_inverse_mut` doesn't leave `self` unchanged if the inversion fails.

### LINKED ISSUE #1380: Inverting a 4x4 matrix with `try_inverse_mut` doesn't leave `self` unchanged if the inversion fails.
Calling `try_inverse_mut` on a 4x4 matrix uses [`do_inverse4`](https://github.com/dimforge/nalgebra/blob/a803815bd1d22a052690cfde3ff7fae26cd91152/src/linalg/inverse.rs#L139) which modifies `out` before checking the determinant.

jnyfah:
Fixes #1380 

Modified the `do_inverse4` function to calculate the determinant of 4x4 matrix before proceeding with the computation of the adjugate matrix to ensure that `out` is not modified if the determinant is zero.

### DIFF
--- a/src/linalg/inverse.rs
+++ b/src/linalg/inverse.rs
@@ -180,99 +180,98 @@ where
 
     if det.is_zero() {
         return false;
-    } else {
-        out[(0, 0)] = cofactor00;
-
-        out[(2, 0)] = m[1].clone() * m[6].clone() * m[15].clone()
-            - m[1].clone() * m[7].clone() * m[14].clone()
-            - m[5].clone() * m[2].clone() * m[15].clone()
-            + m[5].clone() * m[3].clone() * m[14].clone()
-            + m[13].clone() * m[2].clone() * m[7].clone()
-            - m[13].clone() * m[3].clone() * m[6].clone();
-
-        out[(3, 0)] = -m[1].clone() * m[6].clone() * m[11].clone()
-            + m[1].clone() * m[7].clone() * m[10].clone()
-            + m[5].clone() * m[2].clone() * m[11].clone()
-            - m[5].clone() * m[3].clone() * m[10].clone()
-            - m[9].clone() * m[2].clone() * m[7].clone()
-            + m[9].clone() * m[3].clone() * m[6].clone();
-
-        out[(0, 1)] = cofactor01;
-
-        out[(1, 1)] = m[0].clone() * m[10].clone() * m[15].clone()
-            - m[0].clone() * m[11].clone() * m[14].clone()
-            - m[8].clone() * m[2].clone() * m[15].clone()
-            + m[8].clone() * m[3].clone() * m[14].clone()
-            + m[12].clone() * m[2].clone() * m[11].clone()
-            - m[12].clone() * m[3].clone() * m[10].clone();
-
-        out[(2, 1)] = -m[0].clone() * m[6].clone() * m[15].clone()
-            + m[0].clone() * m[7].clone() * m[14].clone()
-            + m[4].clone() * m[2].clone() * m[15].clone()
-            - m[4].clone() * m[3].clone() * m[14].clone()
-            - m[12].clone() * m[2].clone() * m[7].clone()
-            + m[12].clone() * m[3].clone() * m[6].clone();
-
-        out[(3, 1)] = m[0].clone() * m[6].clone() * m[11].clone()
-            - m[0].clone() * m[7].clone() * m[10].clone()
-            - m[4].clone() * m[2].clone() * m[11].clone()
-            + m[4].clone() * m[3].clone() * m[10].clone()
-            + m[8].clone() * m[2].clone() * m[7].clone()
-            - m[8].clone() * m[3].clone() * m[6].clone();
-
-        out[(0, 2)] = cofactor02;
-
-        out[(1, 2)] = -m[0].clone() * m[9].clone() * m[15].clone()
-            + m[0].clone() * m[11].clone() * m[13].clone()
-            + m[8].clone() * m[1].clone() * m[15].clone()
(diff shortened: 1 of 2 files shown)


src/linalg/inverse.rs
```diff
@@ -145,124 +145,149 @@ where
...
+                - m[4].clone() * m[11].clone() * m[13].clone())
+        - m[3].clone()
+            * (m[4].clone() * m[9].clone() * m[14].clone()
+                + m[5].clone() * m[10].clone() * m[12].clone()
+                + m[6].clone() * m[8].clone() * m[13].clone()
+                - m[6].clone() * m[9].clone() * m[12].clone()
+                - m[5].clone() * m[8].clone() * m[14].clone()
+                - m[4].clone() * m[10].clone() * m[13].clone());
+    if det.is_zero() {
+        false
```
Ralith, line 177: Can we use an early return here to avoid rightward drift?

src/linalg/inverse.rs
```diff
@@ -145,124 +145,149 @@ where
...
-        + m[8].clone() * m[1].clone() * m[6].clone()
-        - m[8].clone() * m[2].clone() * m[5].clone();
-
-    let det = m[0].clone() * out[(0, 0)].clone()
-        + m[1].clone() * out[(0, 1)].clone()
-        + m[2].clone() * out[(0, 2)].clone()
-        + m[3].clone() * out[(0, 3)].clone();
-
-    if !det.is_zero() {
+    let det = m[0].clone()
```
Ralith, line 148: This seems to be adding a bunch of additional multiplies compared to the previous version. Is there a way to reuse intermediate results here?

Ralith (changes requested):

Ralith (approved):
LGTM, thanks!

Ralith:
I guess we don't have any unit tests validating this. Want to add one?

jnyfah:
@Ralith, I realized that I had mistakenly omitted the cofactor `out[(1, 0)]`  previously , which was causing the `self_mul_inv_is_id_dim4` test to fail in the workflow. I've added it now.

Ralith:
Oh, nice, looks good then.

Ralith (approved):

Andlon:
Thanks a lot @jnyfah! I'm just quickly reading over here, but it seems to me that we still don't have a test that would fail *before* this change, but now passes? I think this would be useful, and I think it's what @Ralith alluded to before. Perhaps I'm missing something though!

jnyfah:
@Andlon oh yes, added the test now

tests/core/matrix.rs
```diff
@@ -970,6 +970,14 @@ mod inversion_tests {
             }
         }
 
+        #[test]
+        fn test_inversion_failure_leaves_matrix_unchanged(m in matrix4()) {
+                let original_matrix = m.clone();
+                if m.try_inverse().is_none() {
```
Ralith, line 976: Does this branch actually get hit? I'm not sure how often proptest generates non-invertible matrices. Would a hardcoded case work better?

Ralith:
I was actually just asking about a test to make sure this isn't causing a regression, but that was already covered.

tests/core/matrix.rs
Andlon: Agree with @Ralith, I think the likelihood of `matrix4()` producing singular matrices is likely (very) small (though I haven't checked), so a test with a fixed input is better in this case.

jnyfah:
@Ralith @Andlon added a test with a fixed noninvertible input

tests/core/matrix.rs
```diff
@@ -1263,6 +1263,20 @@ fn column_iterator_double_ended_mut() {
...
+#[test]
+fn test_inversion_failure_leaves_matrix_unchanged() {
+    let mut mat = na::Matrix4::new(
+        1.0, 2.0, 3.0, 4.0,
+        2.0, 4.0, 6.0, 8.0,
+        3.0, 6.0, 9.0, 12.0,
+        4.0, 8.0, 12.0, 16.0
+    );
+    let expected = mat.clone();
+    if !mat.try_inverse_mut() {
```
Andlon, line 1275: This test is currently very brittle. Imagine that some change in an implementation detail somewhere caused `mat.determinant()` to return something not exactly `0.0`. Then `mat.try_inverse_mut()` might return `true` and the test would succeed, although it's no longer testing what we want to test it.

Perhaps replace the if by an assert? E.g. `assert!(!mat.try_inverse_mut())` followed by the `assert_eq` that you already have.

tests/core/matrix.rs
```diff
@@ -1263,6 +1263,20 @@ fn column_iterator_double_ended_mut() {
     assert_eq!(col_iter_mut.next(), None);
 }
 
+#[test]
+fn test_inversion_failure_leaves_matrix_unchanged() {
```
Andlon, line 1267: Maybe rename to `_matrix4_unchanged` to indicate that this was specifically an issue with 4x4 inversion and what we're testing here

Andlon (changes requested):
Thanks for adding the test. We're almost there, I'm unfortunately going to have to be a little bit picky still :-) See comments!

tests/core/matrix.rs
jnyfah: @Andlon fixed it

Andlon:
@jnyfah: thanks for fixing the last remaining points! We're all good to go now, except CI is failing due to formatting. I think you just need to run `cargo fmt` and it should be fine.

Andlon:
Thanks for your help with this @jnyfah! The CI failed due to an unrelated issue (a brittle test).

PR #1385, opened 2024-04-22
Fixed a (presumed) typo in documentation in matrix_view.rs

fangs124:
Fixed a (presumed) typo in documentation in matrix_view.rs.

Previously it says fixed_view and fixed_slice returns a matrix with `(R::dim(), CView::dim())` consecutive components. I just changed R to RView, because presumably this is a typo.

This is my first time submitting a pull request, and I would be more than happy to take any input.

### DIFF
--- a/src/base/matrix_view.rs
+++ b/src/base/matrix_view.rs
@@ -546,7 +546,7 @@ macro_rules! matrix_view_impl (
             $me.$generic_view_with_steps(start, shape, steps)
         }
 
-        /// Slices this matrix starting at its component `(irow, icol)` and with `(RVIEW,CVIEW)`
+        /// Slices this matrix starting at its component `(irow, icol)` and with `(RVIEW, CVIEW)`
         /// consecutive components.
         #[inline]
         #[deprecated = slice_deprecation_note!($fixed_view)]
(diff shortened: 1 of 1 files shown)


Ralith (changes requested):
Thanks for the PR! It looks like the docs you're updating affect functions with `RVIEW`/`CVIEW` const generic parameters, not `RView`/`CView` type parameters.

fangs124:
sorry, so should they all say `(RVIEW,CVIEW)` instead of `(RView::dim(), CView::diem())`? If so there's a few other functions on that file that has that typo in the documentation.

Ralith:
That makes sense to me, at least!

Ralith (approved):
Thanks!

src/base/matrix_view.rs
```diff
@@ -546,17 +546,17 @@ macro_rules! matrix_view_impl (
             $me.$generic_view_with_steps(start, shape, steps)
         }
 
-        /// Slices this matrix starting at its component `(irow, icol)` and with `(R::dim(),
-        /// CView::dim())` consecutive components.
+        /// Slices this matrix starting at its component `(irow, icol)` and with `(RVIEW,CVIEW)`
```
Ralith, line 549: nit: space after the `,` here and below

PR #1399, opened 2024-05-23
Update inverse.rs - improved code doc try_inverse_mut

tencek:
The "and leaves `self` untouched" part of the code doc was wrong. Self may be modified before early false return.

### DIFF
--- a/src/linalg/inverse.rs
+++ b/src/linalg/inverse.rs
@@ -29,8 +29,8 @@ impl<T: ComplexField, D: Dim, S: Storage<T, D, D>> SquareMatrix<T, D, S> {
 }
 
 impl<T: ComplexField, D: Dim, S: StorageMut<T, D, D>> SquareMatrix<T, D, S> {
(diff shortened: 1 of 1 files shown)


Andlon:
I believe this was fixed in #1384 for the 4x4 case. I think it should work for the other cases. Can you confirm whether this fixes the issue for you @tencek?

tencek:
@Andlon - thank you for looking into this one and I apologize for my slow response.
The 4x4 case seems to be fine. The case that caused my troubles was 6x6. Please have a look at this unit test that is failing on the dev branch:
[test_inversion_failure_leaves_matrix6_unchanged](https://github.com/dimforge/nalgebra/compare/dev...tencek:nalgebra:test_inversion_failure_leaves_matrix6_unchanged?expand=1)

Fun fact: the unit test, including the values,  was completely generated by the CoPilot. I just found the right place where to add the new test and was about to copy and paste the values I used in my software to replicate the issue. But the CoPilot was quicker and the values generated also replicated the bug.

Issue #1400, opened 2024-06-08
Matrix should have `sub_scalar` and `sub_scalar_mut` methods.


Sol-Ell:
The `Matrix` struct currently includes the `add_scalar` and `add_scalar_mut` methods; however, it lacks the `sub_scalar` and `sub_scalar_mut` methods. Given that types such as `Matrix<usize, ..>` cannot have negative values, I believe it is essential to include this functionality.

Sol-Ell:
For example, attempting to compile this code results in a compilation error.
```rust
use nalgebra::Matrix2;

fn main() {
    let mut addresses: Matrix2<usize> = Matrix2::new(1, 2, 3, 4);
    addresses.add_scalar_mut(-1);
    let expected = Matrix2::new(0, 1, 2, 3);
    assert_eq!(addresses, expected)
}
```

Andlon:
My opinion is that I don't think we should strive to support matrix computations with unsigned integers. It's already difficult to ensure that scalar `usize` arithmetic does not underflow, but correct subtractive arithmetic with *matrices* requires so much care that it's simply hardly ever a good idea. I don't think we should encourage this by adding convenience functionality that supports this use case.

I'd recommend to use signed integers for matrix arithmetic, and convert only at the end, where you can check for negative results.

Sol-Ell:
Thank you for the clarification.

PR #1401, opened 2024-06-08
Added `sub_scalar` and `sub_scalar_mut` methods to `Matrix`.

Sol-Ell:
See issue #1400 for more info.

# Changelog

* Added `sub_scalar` and `sub_scalar_mut` methods to `Matrix`.


Andlon:
I added a comment in issue #1400 why I personally don't think we should add convenience functionality that encourages use of unsigned integers in matrix arithmetic. Just noting this here as it might otherwise seem like an "easy merge".

PR #1409, opened 2024-06-19
Support conversion for glam 0.28

waywardmonkeys:


waywardmonkeys:
@sebcrozet I updated the changelog for the changes from 0.32.6 but maybe that should be a separate PR / commit.

sebcrozet:
Thanks!

PR #1447, opened 2024-10-16
feat: implement bytemuck traits for isometries and similarities

sebcrozet:
- Add implementations of bytemuck traits for isometries and similarities.
- Implement `AsRef<[T]>` for matrices with contiguous storage.
- Enable the `num-complex/bytemuck` feature when the `convert-bytemuck` feature is enabled.


sebcrozet:
Looks like xargo is not happy (but that seems unrelated to this PR). Will have a look separately.

PR #1454, opened 2024-10-29
Fix no-std CI build configuration

waywardmonkeys:
* Stop using `xargo` and just use a regular `cargo` build.
* Use `stable` rather than `nightly` as that works with `cargo`.
* Update from `actions-rs/toolchain` for this job to `dtolnay/rust-toolchain` as the `actions-rs` action is unmaintained.
* Switch from `x86_64-unknown-linux-gnu` to `x86_64-unknown-none` as that is a 64 bit target that has no `std` lib.
* Keep the 32 bit ARM target testing to keep that coverage.
* Fix the `rand-no-std` test to pass that feature to the build.

### DIFF
--- a/.github/Xargo.toml
+++ b/.github/Xargo.toml
@@ -1,5 +1,0 @@
-[target.x86_64-unknown-linux-gnu.dependencies]
-alloc = {}
-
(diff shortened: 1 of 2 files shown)


sebcrozet:
Thanks!

PR #1458, opened 2024-10-31
Fix comment on Matrix::fill_upper_triangle

RCoder01:
Matrix::fill_upper_triangle currently says that it "Sets all the elements of the _lower_-triangular part...", when it should say it fills the "_upper_-triangular part"


Ralith (approved):
Thanks!

PR #1462, opened 2024-11-11
Fix simd_try_normalize mask logic

jmagnuson:
Similar to `try_normalize`, `simd_try_normalize` compares the norm `n` against `min_norm`. The former properly allows a value if it is greater than the minimum:

https://github.com/dimforge/nalgebra/blob/a3c261051ba7c8201ef8c49640ae24c1f53360f7/src/base/norm.rs#L397-L401

The simd version uses this comparison boolean as a lanewise mask, but the logic is currently inverted, such that a valid value greater than `min_norm` results in a `false` mask.

### DIFF
--- a/src/base/norm.rs
+++ b/src/base/norm.rs
@@ -328,9 +328,9 @@ impl<T: Scalar, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
         DefaultAllocator: Allocator<R, C>,
     {
         let n = self.norm();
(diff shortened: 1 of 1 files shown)


jmagnuson:
I'm not sure why `linalg::eigen::proptest_tests::f64::symmetric_eigen` [failed](https://github.com/dimforge/nalgebra/actions/runs/11780769011/job/32812085823?pr=1462#step:4:1085); all tests passed locally using rustc 1.79.0.

PR #1470, opened 2025-01-05
Varios refactorings

adamnemecek:
- made OMatrix::sum faster by using add_assign
- use map on optionals where possible
- refactored partial_cmp

### DIFF
--- a/src/base/ops.rs
+++ b/src/base/ops.rs
@@ -398,11 +398,13 @@ where
     /// iter::empty::<DMatrix<f64>>().sum::<DMatrix<f64>>(); // panics!
     /// ```
     fn sum<I: Iterator<Item = OMatrix<T, Dyn, C>>>(mut iter: I) -> OMatrix<T, Dyn, C> {
-        if let Some(first) = iter.next() {
-            iter.fold(first, |acc, x| acc + x)
-        } else {
-            panic!("Cannot compute `sum` of empty iterator.")
+        let mut ret = iter
+            .next()
+            .expect("Cannot compute `sum` of empty iterator.");
+        while let Some(next) = iter.next() {
+            ret += next;
         }
+        ret
     }
 }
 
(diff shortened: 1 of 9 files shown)


src/base/ops.rs
```diff
@@ -398,11 +398,13 @@ where
...
-            iter.fold(first, |acc, x| acc + x)
-        } else {
-            panic!("Cannot compute `sum` of empty iterator.")
+        let mut ret = iter
+            .next()
+            .expect("Cannot compute `sum` of empty iterator.");
+        while let Some(next) = iter.next() {
+            ret += next;
         }
+        ret
```
Ralith, line 407: Why not retain the `fold`?

src/lib.rs
```diff
@@ -339,27 +339,13 @@ pub fn partial_ge<T: PartialOrd>(a: &T, b: &T) -> bool {
...
 pub fn partial_min<'a, T: PartialOrd>(a: &'a T, b: &'a T) -> Option<&'a T> {
-    if let Some(ord) = a.partial_cmp(b) {
-        match ord {
-            Ordering::Greater => Some(b),
-            _ => Some(a),
-        }
-    } else {
-        None
-    }
+    a.partial_cmp(b).map(|ord| if ord.is_ge() { b } else { a })
```
Ralith: To avoid a change in behavior, this should be `is_gt` (and similar below)

Ralith (changes requested):
Thanks, overall these are nice style improvements. For future reference, they'd be a lot easier to review if you grouped changes of the same kind into the same commit, and force-pushed when you discover an error rather than fixing it in a later commit.

src/base/ops.rs
adamnemecek: so that we use add_assign as opposed to allocate a new matrix every time.

src/lib.rs
adamnemecek: `is_ge` == `matches!(self, Greater)` which is what was there before.

Ralith: That's incorrect. The "e" stands for "equals", as in "greater or equals". This is illustrated in [the documentation](https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.is_ge).

Issue #1217, opened 2023-03-07
Confusing documentation on `strides`


johnmave126:
According to [`RawStorage::strides`](https://docs.rs/nalgebra/latest/nalgebra/base/storage/trait.RawStorage.html#tymethod.strides)
> The spacing between consecutive row elements and consecutive column elements.
>
> For example this returns `(1, 5)` for a row-major matrix with 5 columns.

But according to the implementation of `RawStorage::linear_index`: https://github.com/dimforge/nalgebra/blob/52e6db99c6e3615512801d5f43d9f9c8399627d8/src/base/storage.rs#L70-L75

`(1, 5)` is a column-major matrix with 5 rows, as elements in the same column are consecutive in memory.

My personal suggested wording would be "The spacing between consecutive elements across rows and across columns.", but let's discuss. Or we could adapt the wording from [NumPy](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.strides.html).

sebcrozet:
I like the wording you suggest. And, yes, the documentation is misleading here since `nalgebra` does not have row-major matrices anyway.

RaulWCosta:
Hey! If this is still available, I'd love to make a PR :)

PR #1473, opened 2025-01-14
docs: fix documentation on strides

wowinter13:
Resolves https://github.com/dimforge/nalgebra/issues/1217


wowinter13:
@Ralith Could you please check this one?

Ralith:
I see why you found the current text misleading, but I don't find the new text much clearer; I still had to stare at them both for a few minutes to understand what was going on.

How about something like "the offset in the underlying storage corresponding to a change in position by one row or column respectively"?

wowinter13:
@Ralith thanks, addressed!
Sorry for the delay – forgot about this PR

Ralith (approved):

PR #1479, opened 2025-01-26
bump rayon dependency

geo-ant:
I've just bumped the optional rayon dependency to the most recent version (as of today). There are failing tests that were failing before I changed anything. Everything else passes as it should.


jonaspleyer (changes requested):
If there is not conflict in specifying `version = "0.10"` I would prefer this, since it allows greater compatibility with other crates. It could also be an option to choose `version = **">=1.6, <1.11"`** for the same reason.

Ralith:
1.6 is semver-compatible with 1.10, so this doesn't seem to serve any purpose.

PR #1484, opened 2025-02-10
bump rand versions

jonaspleyer:


Ralith:
Addressed in https://github.com/dimforge/nalgebra/pull/1487. Thanks for the PR!

Issue #1446, opened 2024-10-15
Euler angles support for all conventions


tdamsma:
As I was writing python bindings for the [rs-opw-kinematics](https://github.com/bourumir-wyngs/rs-opw-kinematics) library [py-opw-kinematics](https://github.com/CEAD-group/py-opw-kinematics), I struggled with Euler angles as I wanted to support the 6 proper and 6 Tait–Bryan conventions (and both the intrinsic and extrinsic versions). I ended up porting the code from SciPy to Rust, and after plenty of headbanging and extensive testing I think I got it right. 

I ended up defining a `EulerConvention` class that can convert between Euler angles, Quaternions and Rotation matrices. Would there be any interest to incorporate this functionality in the nalgebra package? I would be willing to put some work in a PR, but I would need some guidance as I am quite new to Rust.

TrifanBogdan24:
Hi, I tried to solve you issue in this [Pull Request](https://github.com/dimforge/nalgebra/pull/1500). Please take a look.

tdamsma:
Thanks, I has a quick look and added some comments.

PR #1500, opened 2025-03-23
Implemented Euler angles a separate crate

TrifanBogdan24:
Solved issue: #1446.

I implemented the algorithm behind calculus of Euler Angles as a nalgebra crate.

### DIFF
--- a/nalgebra-euler-angles/src/euler_angles.rs
+++ b/nalgebra-euler-angles/src/euler_angles.rs
@@ -1,0 +1,183 @@
+pub fn zeros(num_lines: &usize, num_columns: &usize) -> Vec<Vec<f64>> {
+    let mut matrix: Vec<Vec<f64>> = Vec::new();
+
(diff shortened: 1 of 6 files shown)


nalgebra-euler-angles/src/main.rs
tdamsma, line 1: Should this be included?

nalgebra-euler-angles/src/euler_angles.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    for &axis in order {
+        match euler_map.get(&axis) {
+            Some(&angle) => euler_angles.push(angle), 
+            None => return Err(format!("Unexpected axis '{}'", axis)),
+        }
+    }
+
+    Ok(euler_angles)
+}
```
tdamsma, line 181: Is the using the extrinsic or extrinsic definition? And how to discern between those?

Issue #628, opened 2019-08-05
Explain why each use of `unsafe` is sound


sebcrozet:
See [that discussion on reddit](https://www.reddit.com/r/rust/comments/cm2858/about_the_future_of_nphysics_a_pure_rust_2d_and/ew0erxn/).

We should add comments explaining why each `unsafe` is sound throughout the nalgebra source code.

Doing this will take an extremely large amount of time so any incremental contribution is welcome. In other words, event adding one comment for one `unsafe` statement is a great contribution and justify a PR.

danieldinu2030:
Hello, I tried providing good explanations for several files where unsafe is used. View PR #1501 for more details.

PR #1501, opened 2025-03-23
Added comments when using unsafe

danieldinu2030:
## What changed?

- added a comment in 15 files where 'unsafe' is being used
- explained to the others why the code is still sound, following issue #628 

## Why?

To get issue #628 closer to being completely solved

### DIFF
--- a/nalgebra-glm/src/common.rs
+++ b/nalgebra-glm/src/common.rs
@@ -133,6 +133,8 @@ pub fn clamp_vec<T: Number, const D: usize>(
 ///
 /// The floating-point value's bit-level representation is preserved.
 ///
+/// Using unsafe is sound because the bitwise representation of f32 fits in i32
+///
 /// # See also:
 ///
 /// * [`float_bits_to_int_vec()`]
@@ -166,6 +168,8 @@ pub fn float_bits_to_int_vec<const D: usize>(v: &TVec<f32, D>) -> TVec<i32, D> {
 /// Returns an unsigned integer value representing the encoding of a floating-point value.
 ///
 /// The floating-point value's bit-level representation is preserved.
+///
+/// Using unsafe is sound because the bitwise representation of f32 fits in i32
 ///
 /// # See also:
 ///
--- a/nalgebra-lapack/src/cholesky.rs
+++ b/nalgebra-lapack/src/cholesky.rs
@@ -161,6 +161,7 @@ where
         lapack_check!(info);
 
         // Copy lower triangle to upper triangle.
+        // Using unsafe to ensure the bounds i and j are always valid indices,
         for i in 0..dim {
             for j in i + 1..dim {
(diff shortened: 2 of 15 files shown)


nalgebra-glm/src/common.rs
```diff
@@ -133,6 +133,8 @@ pub fn clamp_vec<T: Number, const D: usize>(
 ///
 /// The floating-point value's bit-level representation is preserved.
 ///
+/// Using unsafe is sound because the bitwise representation of f32 fits in i32
```
Ralith, line 136: This documentation goes inside the function, on the `unsafe` block. The use of unsafety in the implementation is not public; the purpose of the comment is to help a developer or auditor who is looking at the implementation.

nalgebra-glm/src/common.rs
```diff
@@ -167,6 +169,8 @@ pub fn float_bits_to_int_vec<const D: usize>(v: &TVec<f32, D>) -> TVec<i32, D> {
...
```
Ralith, line 172: As above.

nalgebra-lapack/src/cholesky.rs
```diff
@@ -161,6 +161,7 @@ where
         lapack_check!(info);
 
         // Copy lower triangle to upper triangle.
+        // Using unsafe to ensure the bounds i and j are always valid indices,
```
Ralith, line 164: This comment doesn't make sense. The unsafe operation *requires* that the indices are in bounds; it does nothing to ensure that is actually the case. The comment should explain how we know it's the case.

nalgebra-lapack/src/cholesky.rs
```diff
@@ -196,6 +197,9 @@ pub trait CholeskyScalar: Scalar + Copy {
     fn xpotri(uplo: u8, n: i32, a: &mut [Self], lda: i32, info: &mut i32);
 }
 
+/// This macro uses unsafe to manually ensure memory safety for external functions
+/// For incorrectly sized and initialized matrices and arrays, undefined behavior will occur
```
Ralith, line 201: This comment seems to be arguing that the use of unsafety is incorrect. If that's true, then we should fix the unsoundness, not just declare its presence. Several similar looking instances below.

nalgebra-lapack/src/svd.rs
```diff
@@ -105,7 +105,9 @@ macro_rules! svd_impl(
                 let mut lwork = -1 as i32;
                 let mut info  = 0;
                 let mut iwork = vec![0; 8 * cmp::min(nrows.value(), ncols.value())];
-
+                
+                // Using unsafe to manually ensure memory safety for external function lapack_func
+                // Rust cannot check the slices' or raw poinnters' safety
```
Ralith, line 110: Here and below, the comment must explain exactly what the safety invariants are, and how we know they're satisfied.

Ralith (changes requested):

PR #1502, opened 2025-03-28
Upgrade to rkyv 0.8

Vrixyz:
It's not compiling yet, but can serve as a discussion support.

## Helpful resources

- rkyv talk: https://www.youtube.com/watch?v=ON4z2LbTD-4
- wasmer migration: https://github.com/wasmerio/wasmer/pull/5142

## helpful commits from rkyv

- rkyv validation change: https://github.com/rkyv/rkyv/commit/172b451ad31691246f2aaed68dd5fb422e9572dc#diff-f2e5d2d70a5fc0728c12e8f58c7f54f5200b7c98cd09cedd89fb55c07c2b711e
- rkyv size_32 to pointer_width: https://github.com/rkyv/rkyv/commit/d592789c7a21570455682c095486916a6e7b61a8#diff-f2e5d2d70a5fc0728c12e8f58c7f54f5200b7c98cd09cedd89fb55c07c2b711e

### DIFF
--- a/src/base/matrix.rs
+++ b/src/base/matrix.rs
@@ -13,8 +13,6 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 #[cfg(feature = "rkyv-serialize-no-std")]
 use super::rkyv_wrappers::CustomPhantom;
-#[cfg(feature = "rkyv-serialize")]
-use rkyv::bytecheck;
 #[cfg(feature = "rkyv-serialize-no-std")]
 use rkyv::{with::With, Archive, Archived};
 
@@ -161,16 +159,14 @@ pub type MatrixCross<T, R1, C1, R2, C2> =
 #[cfg_attr(
     feature = "rkyv-serialize-no-std",
     derive(Archive, rkyv::Serialize, rkyv::Deserialize),
-    archive(
-        as = "Matrix<T::Archived, R, C, S::Archived>",
-        bound(archive = "
-        T: Archive,
-        S: Archive,
-        With<PhantomData<(T, R, C)>, CustomPhantom<(Archived<T>, R, C)>>: Archive<Archived = PhantomData<(Archived<T>, R, C)>>
-    ")
+    rkyv(as = Matrix<T::Archived, R, C, S::Archived>,
+        archive_bounds(
+            T: Archive,
+            S: Archive,
+            With<PhantomData<(T, R, C)>, CustomPhantom<(Archived<T>, R, C)>>: Archive<Archived = PhantomData<(Archived<T>, R, C)>>
+        )
     )
 )]
-#[cfg_attr(feature = "rkyv-serialize", derive(bytecheck::CheckBytes))]
 pub struct Matrix<T, R, C, S> {
     /// The data storage that contains all the matrix components. Disappointed?
     ///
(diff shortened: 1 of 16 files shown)


tests/core/rkyv.rs
```diff
@@ -5,16 +5,16 @@ use na::{
...
         fn $test() {
             let value: $ty<f32> = rand::random();
-			let bytes = rkyv::to_bytes::<_, 256>(&value).unwrap();
+			let bytes = rkyv::to_bytes::<rkyv::rancor::Error>(&value).unwrap();
 
-			let archived = rkyv::check_archived_root::<$ty<f32>>(&bytes[..]).unwrap();
+            let archived = rkyv::access::<<$ty<f32> as rkyv::Archive>::Archived, rkyv::rancor::Error>(&bytes)
+        .unwrap();
             // Compare archived and non-archived
 			assert_eq!(archived, &value);
```
Vrixyz, line 20: I imagine comparison between archived and original type is less automatic now, so it will have to be explicitly added for each type through `rkyv(compare(PartialEq))`

Vrixyz, line 20: The debug check underneath is adding a constraint on `Archived` to be `Debug`, which was not explicit before, I think that's what we'll want to do like what I did for `Matrix` through `archive_bounds(<S as Archive>::Archived: core::fmt::Debug)`, but it feels like it's adding potentially unnecessary constraints.

Could a manual implementation could implement Debug only for compatible types ?

src/base/matrix.rs
```diff
@@ -161,16 +157,18 @@ pub type MatrixCross<T, R1, C1, R2, C2> =
...
-        bound(archive = "
-        T: Archive,
-        S: Archive,
-        With<PhantomData<(T, R, C)>, CustomPhantom<(Archived<T>, R, C)>>: Archive<Archived = PhantomData<(Archived<T>, R, C)>>
-    ")
+    rkyv(
+        compare(PartialEq, PartialOrd),
+        derive(Debug),
+        archive_bounds(
+            <S as Archive>::Archived: core::fmt::Debug,
```
Vrixyz, line 164: I feel like this is a bit constraining ? This is only needed for automated derive (needed for tests)
 
See https://github.com/dimforge/nalgebra/pull/1502/files#r2025182401

src/base/matrix.rs
```diff
@@ -161,16 +157,18 @@ pub type MatrixCross<T, R1, C1, R2, C2> =
 #[cfg_attr(
     feature = "rkyv-serialize-no-std",
     derive(Archive, rkyv::Serialize, rkyv::Deserialize),
```
Vrixyz, line 159: Do we want to derive `rkyv::Portable` when possible ? I'm not sure what the implications are :/

tests/core/rkyv.rs
Vrixyz: going through ~`as = Matrix<Archived::T>` and leveraging `Portable` is probably the best path forward, but I'm not sure how exactly.

Vrixyz: Those must be derived manually when using `as`, otherwise `compare(PartialEq)`.

src/geometry/translation.rs
```diff
@@ -15,23 +15,18 @@ use crate::base::{Const, DefaultAllocator, OMatrix, SVector, Scalar};
...
-#[cfg(feature = "rkyv-serialize")]
-use rkyv::bytecheck;
-
 /// A translation.
 #[repr(C)]
 #[cfg_attr(
     feature = "rkyv-serialize-no-std",
     derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize),
-    archive(
-        as = "Translation<T::Archived, D>",
```
Vrixyz, line 27: This `as` being removed is a UX downgrade because we can't access underlying functions from original type on rkyv archived type.

PR #1504, opened 2025-04-07
Update errors for a more recent rust version

Vrixyz:
Current `main` branch targets rust 1.79 ; locally errors are different and are failing.

Ideally we wouldn't have local fails, should `trybuild` tests be made with a more recent rust version ?

Are these new error messages expected ? Are they better or worse ?


Vrixyz:
Closing in favor of https://github.com/dimforge/nalgebra/pull/1511

PR #1505, opened 2025-04-07
Fix CI: `half` requiring rust 1.81

Vrixyz:
Criterion brings `half` crate which Minimum Supported Rust Version is 1.81, so CI is failing, let's discuss how to fix:

- I bumped rust version and kept a check on MSRV.
- I updated the expected error message for rust 1.81 ; https://github.com/dimforge/nalgebra/pull/1504 may be relevant

I removed the warning to errors, let's see how it goes.

### DIFF
--- a/.github/workflows/nalgebra-ci-build.yml
+++ b/.github/workflows/nalgebra-ci-build.yml
@@ -2,9 +2,9 @@ name: nalgebra CI build
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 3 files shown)


Ralith (approved):
> We probably don't need to cover whatever targets pull in criterion in the MSRV test? 1.81 seems fine to me though.

Not sure I understand, are you suggesting to keep a `test` on MSRV ?
- it's challenging because criterion is pulled from our dev-dependencies, which we can't make optional (https://github.com/rust-lang/cargo/issues/1596), even though it's only used for benches,

Vrixyz:
> We probably don't need to cover whatever targets pull in criterion in the MSRV test? 1.81 seems fine to me though.

Not sure I understand, are you suggesting to keep a `test` on MSRV ?
- it's challenging because criterion is pulled from our dev-dependencies, which we can't make optional (https://github.com/rust-lang/cargo/issues/1596), even though it's only used for benches,

Ralith:
> Not sure I understand, are you suggesting to keep a test on MSRV ?

No, I was referring to the CI job that checks MSRV.

> it's challenging because criterion is pulled from our dev-dependencies, which we can't make optional (

We don't need to make it optional, we just need to not compile any targets that involve dev-dependencies for purposes of validating the MSRV, because they are not relevant to it.

Ralith:
@Vrixyz do you intend to fix the MSRV test to only compile library targets?

Vrixyz:
> @Vrixyz do you intend to fix the MSRV test to only compile library targets?

I can make fixes, but I'm not 100% sure what you mean:

I believe Cargo check doesn't test for dev-dependencies?

Do you mean passing cargo check with specific packages ? Or do you mean changing the features set ?

Thanks for your patience (I have limited computer access for ~1 week)

Ralith:
The purpose of the MSRV test is to ensure that downstream users on older rustc versions can still compile nalgebra. Criterion, being a dev dependency, is not relevant. We should configure the MSRV validation CI job to not attempt to build Criterion (or any other dev-dependency), and instead to only build library targets.

Ralith:
> I think it's interesting to know which MSRV we have for "dev-dependencies".

I don't think this is useful, but since CI does check exact error messages it seems like we need to pin this version to *something* anyway.

PR #1510, opened 2025-04-17
Disable default features for `glam`

Jondolf:
# Objective

Currently, the `convert-glamXYZ` features add `glam` as a dependency with default features enabled. This unconditionally enables the `std` feature in `glam`, making it impossible to use `nalgebra` and `glam` interoperatively with the type conversions in `no_std` environments.

## Solution

Disable default features from the `glam` dependencies. To make CI pass, `std` is still used if it's enabled for `nalgebra`.

Note that I also bumped the rustc version to 1.81.0 here since CI was failing because of it (unrelated to this PR).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,30 @@ path = "src/lib.rs"
 
 [features]
 default = ["std", "macros"]
-std = ["matrixmultiply", "num-traits/std", "num-complex/std", "num-rational/std", "approx/std", "simba/std"]
+std = [
+    "matrixmultiply",
+    "num-traits/std",
+    "num-complex/std",
+    "num-rational/std",
+    "approx/std",
+    "simba/std",
+    "glam014?/std",
+    "glam015?/std",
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -23,7 +23,30 @@ path = "src/lib.rs"
...
+    "glam021?/std",
+    "glam022?/std",
+    "glam023?/std",
+    "glam024?/std",
+    "glam025?/std",
+    "glam027?/std",
+    "glam028?/std",
+    "glam029?/std",
+    "glam030?/std",
+]
```
Jondolf, line 49: This is a bit nasty :/ not sure if there's a better way?

Ralith:
Seems like a good idea to me, but this PR seems to contain unrelated changes, e.g. to the rustc version used for trybuild tests.

Jondolf:
Yeah that's to make CI pass. Won't be needed if we first merge #1505 or some other PR that bumps the version

PR #1514, opened 2025-05-04
Clarify `try_inverse_mut` documentation

### LINKED ISSUE #1497: When Matrix::try_inverse_mut() returns false
[try_inverse_mut](https://docs.rs/nalgebra/latest/nalgebra/base/struct.Matrix.html#method.try_inverse_mut) documents:
> Attempts to invert this square matrix in-place. Returns `false` and leaves `self` untouched if inversion fails.

For matrix sizes 5 and above, the implementation is:

```rust
                    let oself = self.clone_owned();
                    lu::try_invert_to(oself, self)
```

In turn, [`lu::try_invert_to(matrix, out)`](https://docs.rs/nalgebra/latest/nalgebra/linalg/fn.try_invert_to.html) documents:

> If `matrix` is not invertible, `false` is returned and `out` may contain invalid data.

So it looks like `Matrix::try_inverse_mut` may fail to keep its documented promise to leave `self` untouched

PaulXiCao:
Closes #1497.

Updates the documentation to properly reflect the implementation that changes the input entries even when inversion fails (for matrix dimension `n>=5`).

Changing implementation to uphold its current promise is discarded as it would increase the memory footprint. A workaround is mentioned in the new documentation.

### DIFF
--- a/src/linalg/inverse.rs
+++ b/src/linalg/inverse.rs
@@ -29,8 +29,18 @@ impl<T: ComplexField, D: Dim, S: Storage<T, D, D>> SquareMatrix<T, D, S> {
 }
 
 impl<T: ComplexField, D: Dim, S: StorageMut<T, D, D>> SquareMatrix<T, D, S> {
-    /// Attempts to invert this square matrix in-place. Returns `false` and leaves `self` untouched if
-    /// inversion fails.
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thank you!

Issue #865, opened 2021-04-10
It would be great to add LDL decomposition


strasdat:
https://en.wikipedia.org/wiki/Cholesky_decomposition#LDL_decomposition

Or is this the undocumented UDU factorization? That is a little unclear to me.
https://github.com/dimforge/nalgebra/blob/dev/src/linalg/udu.rs

lebensterben:
No they are not the same one.

ajefweiss:
I would second this.

It should(?) be fairly straightforward to re-use the existing code from UDU factorization (#766) and just drop in the LDL algorithm.
I have a very basic version of this that I use internally, with added functionality for semi-definite matrices to get around the fact that the Cholesky decomposition only works for strictly positive definite matrices. It is not clear to me why the implementation for UDU was just restricted to positive definite matrices as well.

Fairly new to Rust, but if there is interest I may be willing to have a go at it.

PR #1515, opened 2025-05-06
Add LDL decomposition

ajefweiss:
See issue #865, this decomposition is also mentioned in #762 but in the end only the UDU decomposition was implemented (PR #766). I need this for my own use case where I basically want a Cholesky decomposition of a semi-definite covariance matrix to generate random correlated numbers. Currently I was applying the Cholesky decomposition on a modified matrix with all zero rows/columns removed, which are then later re-added to the lower triangular matrix.

I used the code from #766 as a template, but changed the code to implement the LDL algorithm instead, everything else is basically the same just re-named. This is not the standard algorithm, but a variant that should also work with semi-definite matrices (i.e. with eigenvalues equal to zero), which I need. I haven't added any tests for this yet though. I do not know why the existing UDU decomposition also does not include this capability as its just an extra 1-2 lines of code.

I also added a `cholesky_l` function that reconstructs the lower triangular matrix from the Cholesky decomposition.


alexhroom:
hi, looking forward to this being added - I was wondering whether it's necessary for it to insist on the scalar type having trait `RealField`? In the UDU factorisation PR it's set to `RealField` because nobody was sure whether the UDU decomposition made sense for complex Hermitian matrices, but LDL is pretty widely used for complex matrices.

I tested it out on a few complex Hermitian matrices and got the results I was expecting. Maybe the functionality could be extended to `ComplexField` and you add some tests for equality with Cholesky on positive-definite complex Hermitian examples?

ajefweiss:
I added functionality for Hermitian matrices, and rewrote the tests.

Currently, the algorithm checks for diagonal values that are exactly zero, to zero out the corresponding columns/rows. One could add a threshold value `eps`, similar to the way its done for the pseudo_inverse() function, so that values below `eps` are set to zero. Only problem here is that the diagonal values have no real relation to the eigenvalues, so the interpretation is not clear.

alexhroom:
@ajefweiss cool! for clarity i'm not a maintainer/reviewer of `nalgebra` i'm just interested in having this in the package :)

geo-ant:
Hi @ajefweiss, I'd be willing to review this but I need your help :slightly_smiling_face: I skimmed the paper linked in https://github.com/dimforge/nalgebra/issues/762, but I couldn't find an algorithm for the LDL decomposition. I see there's a screenshot in the issue for UDU decomposition. Could you point me to the reference that tells me how you implemented the LDL decomposition, so I can read up on it?

ajefweiss:
Hi @geo-ant!

Thank you for taking a look at this. The actual reference I used is just the wikipedia entry which has imo a nicer description of the algorithm (https://en.wikipedia.org/wiki/Cholesky_decomposition#LDL_decomposition_2) which cites "Fundamentals of Matrix Computations" from Watkins (PDF versions are easily obtainable online). The relevant pages here are 82-84 although the algorithm here is just explained as a variant of the LU decomposition without being explicitly written out.

The only problem is that I have not been able to find an explicit reference for an algorithm that allows for semi-positive definite matrices. Both the reference you linked, and mine, require positive-definite matrices in their algorithms due to the division by v_i / d_i. Thus when one of the diagonal entries is 0 the algorithm fails. From my understanding, the A=LL/LDL factorizations are unique if, and only if, the matrix A is positive-definite. The theorems from both references state that the LDL factorization does exist only under the requirement that A is symmetric (and not necessarily positive definite). 

I think the factorization is no longer unique because you can fill out the relevant column in the L matrix with any values as they will be subsequently zero'd out in the multiplication. In my code this column is filled with zeros.

PS: I might try to think a better name for the function "cholesky_ldl" as I actually don't think that name is appropriate and might be a bit confusing.

src/linalg/ldl.rs
```diff
@@ -0,0 +1,102 @@
+#[cfg(feature = "serde-serialize-no-std")]
+use serde::{Deserialize, Serialize};
+
+use crate::allocator::Allocator;
+use crate::base::{Const, DefaultAllocator, OMatrix, OVector};
+use crate::dimension::Dim;
+use simba::scalar::ComplexField;
+
+/// LDL factorization.
```
geo-ant: could you expand this to include more about the pre-and-postconditions? For example:
* a sentence like "LDL factorization of a (hermitian?) matrix `A` into a lower unit-triangular (??) matrix `L` and a diagonal matrix `D` such that `A = LDL^T`". 
* what assumptions are made about the matrix `A`. I think it must be Hermitian?

It's fine to keep it concise, but I find it helpful to have all the important info right there from a user's perspective.

Thinking about this some more, do you think this decomposition should be called LDL or LDLT? I'm unsure...

src/linalg/ldl.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl<T: ComplexField, D: Dim> LDL<T, D>
+where
+    DefaultAllocator: Allocator<D> + Allocator<D, D>,
+{
+    /// Computes the LDL^T factorization.
+    ///
+    /// The input matrix `p` is assumed to be hermitian c and this decomposition will only read
+    /// the lower-triangular part of `p`.
+    pub fn new(p: OMatrix<T, D, D>) -> Option<Self> {
```
geo-ant: this is absolutely a me-problem, but is there a reason why the matrix is called `p`, rather than `mat` or `a`? `p` always makes me think of a permutation matrix like in `AP = QR`... :sweat_smile:

src/linalg/ldl.rs
```diff
@@ -0,0 +1,102 @@
...
+                    l[(i, j)] = l_ij / d[j].clone();
+                }
+            }
+        }
+
+        Some(Self { l, d })
+    }
+
+    /// Returns the lower triangular matrix as if generated by the Cholesky decomposition.
+    pub fn cholesky_l(&self) -> OMatrix<T, D, D> {
```
geo-ant: I'd rather not have this in there since it's [easy to google](https://sites.math.duke.edu/~jdr/2021f-218/materials/week11.pdf) how to obtain an LL^T decomposition from an LDL^T decomposition. I don't have super strong feelings about this, but the naming is a bit awkward and the use case seems pretty fringe. But if you tell me this absolutely needs to be in there, I'll concede.

tests/linalg/ldl.rs
```diff
@@ -0,0 +1,108 @@
...
+            mod $module {
+                #[allow(unused_imports)]
+                use crate::core::helper::{RandScalar, RandComplex};
+                use crate::proptest::*;
+                use proptest::{prop_assert, proptest};
+
+                proptest! {
+                    #[test]
+                    fn ldl(m in dmatrix_($scalar)) {
+                        let m = &m * m.adjoint();
```
geo-ant, line 102: If I understood correctly what you said in the comment, then this implementation of the LDL^T decomposition requires the matrix to be symmetric positive *definite* (not *semi* definite), right? However, using `M M^T` will only give us pos. semidefinite, so we should expect some spurious failures here. I've actually run into the same problem in `nalgebra-lapack` and you can see my solution [here](https://github.com/dimforge/nalgebra/blob/main/nalgebra-lapack/tests/linalg/cholesky.rs). I've added a proptest function that gives a spd matrix in a bit of a hacky, but sound way: The idea is to use `M' = M M^T + alpha * Id`, where alpha is chosen suitably large to make we don't get into numerical trouble. You should be able to easily adapt the functions to return complex matrices.

src/linalg/ldl.rs
```diff
@@ -0,0 +1,102 @@
...
+
+                for k in 0..j {
+                    l_ij -= l[(j, k)].clone().conjugate() * l[(i, k)].clone() * d[k].clone();
+                }
+
+                if d[j] == T::zero() {
+                    l[(i, j)] = T::zero();
+                } else {
+                    l[(i, j)] = l_ij / d[j].clone();
+                }
```
geo-ant: is it actually correct to do that or should we return `None` on division by zero?

src/linalg/ldl.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        for j in 0..n {
+            let mut d_j = p[(j, j)].clone();
+
+            if j > 0 {
+                for k in 0..j {
+                    d_j -= l[(j, k)].clone() * l[(j, k)].clone().conjugate() * d[k].clone();
+                }
+            }
+
```
geo-ant: I've thought about the algorithm a bit and I'm wondering if you'd mind implementing the algorithm in https://github.com/dimforge/nalgebra/issues/762#issuecomment-694735837. The reason is, that algorithm should be a  more efficient in terms of memory usage (and that might make it more cache friendly, but I'm not sure) than your implementation. The reason is that the linked algorithm works in place, while your algorithm works out of place. Your interface wouldn't even have to change because you're taking the OMatrix by value already. Your original matrix A would then contain the LDL decomposition where L is in the stricly lower triangular part and d is on the diagonal.

The only problem I see then, is that exposing access to L and D would require new allocations. Is that why you implemented the algorithm out of place in the first place?

What I mean is this: is the primary use of LDL^T decomposition to get explicit access to L and D, or would it also be used to solve linear systems. In that case we could expose a solve method that uses the triangular solvers internally and deals with D through scaling.

geo-ant (commented):
thanks for the great work, see some comments in the review. 

You said that this algorithm _should_ also exist for positive **semi** definite matrices, but your implementation only covers positive definite ones. As long as we document this, I don't see a problem here.

tests/linalg/ldl.rs
alexhroom: just wanted to weigh in - i'm a little confused actually... the 'point' of LDL is that it's a generalisation of the Cholesky algorithm to **all** symmetric/Hermitian matrices, with no requirements on positive definiteness/semi-definiteness. If it only works for definite/semidefinite matrices then there's no benefit to using LDL over Cholesky, which is a simpler decomposition _and_ a more efficient algorithm.

src/linalg/ldl.rs
alexhroom: just want to add to make sure we clarify that `D` is **strictly diagonal** - people also use LDL to refer to a different algorithm which decomposes a Hermitian matrix into a lower unit-triangular matrix L and a **block-diagonal** matrix `D` composed of 1x1 and 2x2 blocks. 

This is the algorithm implemented in the [*HETRF family of algorithms in LAPACK](https://netlib.org/lapack/explore-html/d8/d0e/group__hetrf_ga9e7136edefc49152fe75a2c2bc410bce.html), which are [wrapped by SciPy's `ldl` function in Python](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.ldl.html) as well as [MATLAB's `ldl` function](https://www.mathworks.com/help/matlab/ref/ldl.html), so users may be expecting the block-diagonal algorithm and be confused. (the idea of the block-diagonal algorithm is that there are bounds on the size of the elements of `D` which do not exist for the strictly diagonal algorithm)

 The Rust package `faer` has both versions - they call the strictly diagonal one LDLT and the block-diagonal one LBLT.

> Thinking about this some more, do you think this decomposition should be called LDL or LDLT? I'm unsure...

for some comparisons summarising info above, both [Eigen (C++)](https://libeigen.gitlab.io/eigen/docs-nightly/classEigen_1_1LDLT.html) and [`faer` call it LDLT](https://docs.rs/faer/latest/faer/mat/type.Mat.html#method.ldlt), [SciPy calls it LDL](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.ldl.html), as does [MATLAB](https://www.mathworks.com/help/matlab/ref/ldl.html).

tests/linalg/ldl.rs
ajefweiss: Sorry if the above text is slightly confusing. But no, the LDL/LDL^T decomposition only requires the matrix to be symmetric. It does NOT require it to be positive definite. This is stated in both textbooks referenced here.

The problem is that, in both cases, the algorithms that are shown DO require the matrix to be positive OR negative definite. I'd have to look deeper into this, but I think this is because in the semi-positive/negative definite cases the solution is not unique. I think there is a choice for the particular column where the corresponding value in the D matrix is zero.

src/linalg/ldl.rs
ajefweiss: We can change this, currently the naming is just taken from the code from the UDU factorization.

src/linalg/ldl.rs
ajefweiss: See comment above about the choice for semi-positive/negative definite matrices.

src/linalg/ldl.rs
ajefweiss: Thanks for the comments. 

Correction:  My implementation of the algorithm covers semi-positive (and semi-negative) definite matrices. The referenced algorithms in the textbooks DO NOT.

Regarding the naming convection LDL/LDLT. As @alexhroom pointed out, both conventions are used. Given the fact that we already have the UDU factorization implemented as UDU (and not UDU^T), I think it would make sense to keep it as LDL.

src/linalg/ldl.rs
ajefweiss: Maybe a better name for this function would be `lsqrtd`?

One of the main uses cases of this factorization is to get the same result as from Cholesky but without the same constraints.

src/linalg/ldl.rs
ajefweiss: We could do this, but we should then change the structure a bit. LDL could then just be a new type of a single matrix without exposing the underlying field as L*D by itself does not have a use for anything. The current implementation for UDU uses the exactly same set up.

geo-ant: I'd prefer that, unless you say the main value of the LDL(T) decomposition is access to the L and D matrix. I'd still expose access to those matrices via dedicated methods that allocate, because who know what people might use this for. I'd also add a solve method for solving linear systems in a least squares sense, that takes advantage of the internal structure.

src/linalg/ldl.rs
geo-ant: sorry for being a bit dense here (no pun intended), but are you saying that your algorithm works for all symmetric positive or negative *semi*definite matrices? What if you changed the algorithm to use the in-place algorithm as suggested below? That algorithm requires A to have an LU decomp, which I believe precludes *semi*-definiteness...

src/linalg/ldl.rs
ajefweiss: Code has ben re-factored to make the calculations in-place, I'm still using the same algorithm just not allocating new matrices.

tests/linalg/ldl.rs
awinick: Thanks for working on this! Having an LDL-like factorization in nalgebra would be genuinely useful.

As written, this looks to me like the standard no-pivot, strictly diagonal LDLᵀ/LDLᴴ recursion, rather than the more general symmetric-indefinite factorization many users may expect from an `LDL` API.

To handle symmetric/Hermitian-indefinite factorizations, you have to use the Bunch–Kaufman family (or equivalent): symmetric pivoting/permutations, with `D` allowed to have 1×1 and 2×2 blocks. That extra machinery matters for general Hermitian/symmetric indefinite matrices.

In particular, without pivoting and with `D` forced to be strictly diagonal, this does not cover all symmetric matrices. For example, [[0, 1], [1, 0]] does not admit a factorization of the form `A = L D Lᵀ` with unit-lower-triangular `L` and diagonal `D` in that fixed ordering.

Because of that, I’d be hesitant to expose this as a general-purpose `LDL` for Hermitian matrices unless the contract is made very explicit in the docs.
I do think this could still be useful, but I think it would help to choose one of two directions explicitly:

1. Document it narrowly as a no-pivot diagonal LDLᵀ variant intended for semidefinite / covariance-style use cases, with naming/docs/tests that make that scope clear.
2. Or, if the goal is a general Hermitian/symmetric `LDL`, aim for a pivoted/block-diagonal factorization closer to Bunch–Kaufman.

geo-ant: Hey @awinick, thanks for weighing in! Do you have a source for the symmetric indefinite LDL decomposition that you mentioned?

ajefweiss: The original '77 paper is openly available here <https://www.ams.org/journals/mcom/1977-31-137/S0025-5718-1977-0428694-0/S0025-5718-1977-0428694-0.pdf>. Also see the comment here https://github.com/dimforge/nalgebra/pull/1515#discussion_r2554123336

Personally I would prefer continuing with option (1) for the time being as it seems to cover most practical use cases that I have run into. I can clean up the code / docs a bit more at some later time when I have more free time.

geo-ant: thank you!

awinick: Hey @geo-ant you might be interested in the PR I just opened.

ajefweiss:
I am planning to take a look at this again this weekend and clean up the remaining documentation. I believe that having both implementations should be useful as the strictly diagonal LDL version should be quicker if the assumed constraints hold (I plan on testing this).

There is only one thing I think that needs to be resolved regarding the naming convention. LAPACK has both LDL / UDU implemented incl. using the Bunch-Kaufman algorithm. But algebra nalready now has the strictly diagonal version for UDU already implemented. It might be confusing to call the BK implementation just "LDL" as it behaves differently than the existing UDU.

awinick:
@ajefweiss One thought on naming: since this is the strictly diagonal / no-pivot variant, would something like `Qdldl` make sense here? That seems to make the restricted contract clearer than plain `LDL`.

Also, would it make sense to detect breakdown explicitly instead of zeroing the column and continuing when a diagonal pivot is zero? The [[QDLDL implementation](https://github.com/osqp/qdldl/blob/master/src/qdldl.c)](https://github.com/osqp/qdldl/blob/master/src/qdldl.c) feels fairly close in spirit here and returns failure on zero pivots.

If semidefinite-ish cases are still important, maybe that behavior could be configurable via an API choice or pivot tolerance.

Curious what you think.

geo-ant:
Hey, @ajefweiss and @awinick, first of all sorry to both of you for not reviewing your implementations so far, life got in the way...

I'm happy you are already discussing the tradeoffs of having both implementations in nalgebra. In terms of naming, I think we could follow the suggestion of @alexhroom and do it like in `faer`, while also making sure that the documentation is very explicit about what the decomposition does and doesn't provide.

ajefweiss:
I have added extra documentation pointing out the limitation and also referred to the BK algorithm. Until that's merged I don't think I have a good way of creating a proper link to those types but that's always something we can clean up later.

Renamed d/l functions to have the same names as in the upcoming LBL implementation.

LDL::new now has the "expected" behavior and returns None if there is a zero pivot. 

I also added a LDL::new_unchecked for the special case with zero pivots. This decomposition may or may not return a valid LDL decomposition depending on the structure of the matrix, always works for covariance matrices. I have not added this function in the decomposition.rs file

PR #1522, opened 2025-06-07
Improve compile-time usability

TimTheBig:
Improve compile-time usability by adding const to as many functions as possible without `const_traits`


TimTheBig:
Should I add an entry to the change-log?

src/base/matrix.rs
```diff
@@ -400,10 +400,31 @@ where
...
+    ///  );
+    /// assert_eq!(const { MAT.shape_const() }, (3, 4));
+    /// ```
+    #[inline]
+    #[must_use]
+    pub const fn shape_const(&self) -> (usize, usize) {
+        (R, C)
+    }
+}
+
```
TimTheBig:
```suggestion
```

sebcrozet:
Thanks!

PR #1523, opened 2025-06-08
Move to rust 2024 edition

TimTheBig:
This patch moves all crates in the repo to rust 2024 edition.

## Benchmarks
Consistent improvement in large matrices especially asymmetric ones. I also saw a consistent improvement in 4x4.

PS. Sorry for the big PR


sebcrozet:
Thank you for this PR!
The CI failure is unrelated and the PR is big enough as it is so we’ll take care of it in a separate PR.

PR #1532, opened 2025-07-10
Fixes the Schur decomposition for a zero matrix

### LINKED ISSUE #1291: eigenvalues() hangs with all 0's
Running into an issue where when throwing in a matrix of all zeros into `.eigenvalues()`, the function hangs. To replicate, run the below:

```rust
use nalgebra::Matrix6;

fn main() {
    let test: Matrix6::<f64> = Matrix6::zeros();
    println!("{:?}", test.eigenvalues()); // This will hang
}
```
Looks like we're passing in by default `max_niter = 0` into the `Schur::do_decompose()` call, which causes the function to run indefinitely -- not sure if that's best practice. At the very least, this behavior should be called out in the doc comments.

alexhroom:
This PR fixes #1291, by adding a check to avoid a 0/0 division in `Schur.do_decompose()` which was causing the decomposition to hang. The decomposition now works correctly for zero matrices (giving the identity for $Q$ and the zero matrix for $T$).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Moved to Rust 2024 edition.
 - Several methods are now `const` whenever possible. See details in [#1522](https://github.com/dimforge/nalgebra/pull/1522).
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


rbenefo:
Thank you for the fix!

PR #1535, opened 2025-07-31
Release v0.34.0

sebcrozet:
### Added

- Add the `convert-glam030` feature to enable conversion from/to types from `glam` v0.30.
- Add the `defmt` cargo feature that enables derives of `defmt::Format` for all no-std types.

### Changed

- Bumped MSRV to 1.87.0.
- Updated rand dependency to 0.9.0.
- Renamed associated const `DimName::USIZE` to `DimName::DIM`.
- Moved to Rust 2024 edition.
- Several methods are now `const` whenever possible. See details in [#1522](https://github.com/dimforge/nalgebra/pull/1522).
- Features for conversion from/to types from `glam` (such as `convert-glam029`) no longer enable default features for
  `glam`, allowing use in `no_std` environments.

### Fixed

- Fix infinite loop when attempting to take the Schur decomposition of a 0 matrix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ documented here.
 
 This project adheres to [Semantic Versioning](https://semver.org/).
 
-## [Unreleased] (?)
+## [0.34.0] (31 July 2025)
 
 ### Added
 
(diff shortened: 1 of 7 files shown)


sebcrozet:
It’s unclear to me why the CI is failing on nalgebra-sparse. Running the slow tests locally passes every time.
I’ll leave it as is for now, but this will be worth looking into.

Issue #1540, opened 2025-09-06
nalgebra-lapack has failing tests for LU linear system solving


geo-ant:
When running the tests of `nalgebra-lapack`using 

```shell
cargo test --features "proptest-support"
```

I am seeing failing tests:

```shell
test linalg::lu::lu_solve_static ... FAILED
test linalg::lu::lu_solve ... FAILED
  ** On entry to DGEES parameter number  2 had an illegal value
```

Those pertain to solving a linear system with LU decomposition. For now, I believe there is an error in the logic in the test, but the lapack output might suggest that there are other bugs hidden in somewhere. For now, my working hypothesis is that the tests will have problems with rank deficient matrices and zero-rank matrices, but there might be other things awry.

It's not immediately obvious to me from looking at the LU implementation that something is wrong there. But I'll investigate and try to fix the root cause.

I've also noticed that the `nalgebra-lapack` tests are not run together with the `nalgebra` CI test suite. I'd like to enable those tests as part of CI, would that be okay?

geo-ant:
**Update 2**: I've discovered more errors in LU and in the Schur decomposition where e.g. incorrect characters are passed to the LAPACK routines. 

* In the LU decomposition those were logically incorrect (e.g. `'T'` was always passed for solving linear systems, when it should only have been passed when trying to solve the transposed system A^T x = b). This seems like c copy-paste error which can happen to the best of us. However, the tests were indeed failing so this should have been investigated.

* In the Schur decomposition we always pass `'T'` as the `SORT` argument to [`xGEES`](https://www.netlib.org/lapack/explore-html/d5/d38/group__gees.html), which accepts only `'S'` and `'N'` as per the docs. Also the authors clearly didn't know how to deal with the `SELECT` argument, which is fine but the way it was implemented it can lead to undefined behavior if `'S'` is passed to `SORT`. The authors did correctly check the LAPACK `info` variable, which should have reported the error but it seems something silently crashed and some tests weren't run.

I'll submit a PR with fixes this week, maybe today if time allows.

PR #1541, opened 2025-09-08
Fixes for nalgebra-lapack (faulty test, bugs in implementations, compile errors)

geo-ant:
this PR started by addressing failing tests in the LU decomposition of `nalgebra-lapack` and evolved into the following:

* fix bugs in LU implementation
* fix bugs in Schur decomposition
* fix failing tests in Cholesky decomposition that were caused by supplying tests matrices that were symmetric positive **semi** definite rather than strictly positive definite.
* fix failing compilation when using the serde-serialize feature due to faulty trait bounds

I've also added the now successful test suite into the CI pipeline, I hope that's okay. I have another PR in the works that adds a new feature (column pivoted QR decomposition) to this crate and I'd very much like this crate to be tested as part of CI.

Fixes https://github.com/dimforge/nalgebra/issues/1540

### DIFF
--- a/nalgebra-lapack/tests/linalg/cholesky.rs
+++ b/nalgebra-lapack/tests/linalg/cholesky.rs
@@ -3,7 +3,6 @@ use core::f64;
 use na::{Const, DMatrix, Matrix4, Matrix4xX};
 use nl::Cholesky;
 use proptest::prelude::*;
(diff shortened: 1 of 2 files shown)


sebcrozet (approved):
Looking great, thank you!

PR #1551, opened 2025-09-27
Feature/colpiv qr in nalgebra lapack

geo-ant:
This PR adds column-pivoted QR decomposition to `nalgebra-lapack`. This is a pretty useful feature, because while the "normal" `nalgebra` crate also has column-pivoted QR, it will only solve square linear systems. The `nalgebra-lapack` implementation can be used for both square and overdetermined linear systems, which should be a useful addition.

### DIFF
--- a/nalgebra-lapack/src/colpiv_qr.rs
+++ b/nalgebra-lapack/src/colpiv_qr.rs
@@ -4,7 +4,7 @@ use nalgebra::storage::RawStorageMut;
 use nalgebra::{DefaultAllocator, Dim, DimMin, DimMinimum, OMatrix, Scalar, allocator::Allocator};
 use num::float::TotalOrder;
 use num::{Float, Zero};
-use rank::{RankEstimationAlgo, calculate_rank};
+use rank::{RankDeterminationAlgorithm, calculate_rank};
 
 pub mod error;
 use crate::ComplexHelper;
@@ -26,7 +26,7 @@ pub enum Error {
     Backend(#[from] LapackErrorCode),
     #[error("Wrong matrix dimensions")]
     Dimension,
-    #[error("Solving underdetermined systems not supported")]
+    #[error("QR decomposition for underdetermined systems not supported")]
     Underdetermined,
     #[error("Matrix has rank zero")]
     ZeroRank,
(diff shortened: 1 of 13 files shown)


nalgebra-lapack/src/colpiv_qr.rs
```diff
@@ -0,0 +1,777 @@
...
+}
+
+/// Utility trait to add a thin abstraction layer over lapack functionality for
+/// column pivoted QR decomposition.
+#[allow(missing_docs)]
+pub trait ColPivQrScalar: ComplexField + QRScalar {
+    /// routine for column pivoting QR decomposition using level 3 BLAS,
+    /// see https://www.netlib.org/lapack/lug/node42.html
+    /// or https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-0/geqp3.html
+    fn xgeqp3(
```
im-0: I think that all functions of `ColPivQrScalar` must be `unsafe` because underlying lapack functions do not check the lengths of slices.

nalgebra-lapack/src/colpiv_qr.rs
```diff
@@ -0,0 +1,777 @@
...
+    ) -> Result<i32, LapackErrorCode>;
+}
+
+macro_rules! colpiv_qr_real_impl {
+    (
+        $type:ty,
+        xormqr = $xormqr:path $(,)?
+    ) => {
+        impl ColPivQrReal for $type {
+            fn xormqr(
```
im-0: This also looks unsafe...

nalgebra-lapack/src/colpiv_qr.rs
geo-ant: You're right. Frankly, I think all the lapack trait functions in this crate must be unsafe. That is something I already have on my agenda for later, but I will definitely update the stuff inside my current PR.

nalgebra-lapack/src/colpiv_qr.rs
```diff
@@ -0,0 +1,797 @@
...
+    xlapmr = lapack::dlapmr
+);
+
+/// Trait implemented by reals for which Lapack function exist to compute the
+/// column-pivoted QR decomposition.
+// @note(geo-ant) This mirrors the behavior in the existing QR implementation
+// without pivoting. I'm not 100% sure that we can't abstract over real and
+// complex behavior in the scalar trait, but I'll keep it like this for now.
+#[allow(missing_docs)]
+pub unsafe trait ColPivQrReal: ColPivQrScalar + QRReal {
```
im-0: I think that `unsafe` is not required here for a whole trait as merely implementing this trait does not make any safe code that does not use any `unsafe` methods of this trait itself unsafe.

Otherwise safety-related changes LGTM.

nalgebra-lapack/src/colpiv_qr.rs
```diff
@@ -0,0 +1,797 @@
...
+        let d = nrows.min(ncols);
+        let m = self.qr.generic_view((0, 0), (d, d));
+        m.upper_triangle()
+    }
+}
+
+/// Utility trait to add a thin abstraction layer over lapack functionality for
+/// column pivoted QR decomposition.
+#[allow(missing_docs)]
+pub unsafe trait ColPivQrScalar: ComplexField + QRScalar {
```
im-0: Same.

nalgebra-lapack/src/colpiv_qr.rs
geo-ant: Yeah, I went a bit overboard, but my thought was that I'd like to basically prevent people from implementing those traits themselves. I think what I'll actually do is make those traits sealed.

PR #1562, opened 2025-10-24
Use `wrapping_add()` for ptr addition in `Iterator::next()`

mtreinish:
When iterating over a view of a row major matrix and finishing a column during iteration it is possible for the pointer to be incremented outside the allocation when determining the current column has been finished. This is fine because the pointer never is actually reading from outside the allocation, the `inner_end` value used to track we've finished the column will match on the next call to `next()` and never actually attempt to read from outside the allocation for the matrix. However, this increment step is done using `add()` instead of `wrapping_add()` and per the `wrapping_add()` documentation [1] `wrapping_add()` should be used in situations where the pointer can be incremented outside the allocation. While using `add()` in these situations is immediately undefined behavior. This will also trigger miri failures when testing code that is building a `MatrixView` to a row major array and then iterating over it. This commit changes the use of pointer `add()` to `wrapping_add()` in the `Iterator::next()` method to fix this issue.

[1] https://doc.rust-lang.org/std/primitive.pointer.html#method.wrapping_add

### DIFF
--- a/src/base/iter.rs
+++ b/src/base/iter.rs
@@ -100,7 +100,7 @@ macro_rules! iterator {
                         // so this is not observable.
                         if self.size != 0 {
                             let stride = self.strides.0.value();
-                            self.ptr = self.ptr.add(stride);
+                            self.ptr = self.ptr.wrapping_add(stride);
                         }
 
                         Some(old)


Ralith:
It's legal and routine for `add` to result in a pointer that is the first address past the end of the allocation. Is that not what's happening here?

mtreinish:
That is what's happening here, but the documentation on pointers that I linked is pretty clear that using `add()` to advance the pointer outside the allocation is undefined behavior:

> Compared to [add](https://doc.rust-lang.org/std/primitive.pointer.html#method.add), this method basically delays the requirement of staying within the same allocation: [add](https://doc.rust-lang.org/std/primitive.pointer.html#method.add) is immediate Undefined Behavior when crossing object boundaries; wrapping_add produces a pointer but still leads to Undefined Behavior if a pointer is dereferenced when it is out-of-bounds of the object it is attached to

Additionally in the safety section in the `add` method's [documentation](https://doc.rust-lang.org/std/primitive.pointer.html#method.add) says: 

> If the computed offset is non-zero, then self must be [derived from](https://doc.rust-lang.org/std/ptr/index.html#provenance) a pointer to some [allocation](https://doc.rust-lang.org/std/ptr/index.html#allocation), and the entire memory range between self and the result must be in bounds of that allocation. In particular, this range must not “wrap around” the edge of the address space.

If you run `iter()` under miri it will error on the add call here and flag this as undefined behavior. `wrapping_add()` is the correct call to make if we need to increment the pointer outside the allocation like this.

Ralith:
Read the wording you quoted more carefully:

> the entire memory range between self and the result must be in bounds of that allocation

This is true when the result is the first address past the end of the allocation.

PR #1565, opened 2025-10-30
Implement absolute threshold for early deflation in Schur algorithm

### LINKED ISSUE #1528: Infinite loop when computing eigen values
There seems to be a bug with the eigen value finder.

The following code produces a tri-diagonal matrix with real eigenvalues:

```rs
pub fn legendre_nodes(size: usize) -> Vec<f64>
{
    let n = size;
    let mut matrix = nalgebra::DMatrix::<f64>::zeros(n, n);

    let a_fn = |i| (2. * i as f64 - 1.) / i as f64;
    let c_fn = |i| (i as f64 - 1.) / i as f64;
    for i in 1..=n
    {
        let a = a_fn(i);
        let an = a_fn(i + 1);
        let cn = c_fn(i + 1);

        let beta = (cn / (a * an)).sqrt();

        if i < size
        {
            matrix[(i, i - 1)] = beta;
            matrix[(i - 1, i)] = beta;
        }
    }

    println!("{:.2}", matrix);

    let evals = matrix.eigenvalues().unwrap();
    let mut res: Vec<_> = evals.iter().copied().collect();
    for e in &evals
    {
        println!("{}", e);
    }

    res
}
```

Or alternatively:

```
  ┌                     ┐
  │ 0.00 0.58 0.00 0.00 │
  │ 0.58 0.00 0.52 0.00 │
  │ 0.00 0.52 0.00 0.51 │
  │ 0.00 0.00 0.51 0.00 │
  └                     ┘
```

I traced the problem down to [`delimit_subproblem` ](https://github.com/dimforge/nalgebra/blob/main/src/linalg/schur.rs#L338). The start and the end get stuck at values 0 and 3 respectively and never update beyond that.

Toucan4Life:
Fix #1528.
The issue was that, during the deflation step, the diagonal element of the matrix were all zero so the deflation never triggered based on this line : 

` t[(n, m)].clone().norm1() <= eps.clone() * (t[(n, n)].clone().norm1() + t[(m, m)].clone().norm1())`

added a check to compare an off-diag element to an absolute very small value `(eps.clone() * eps.clone())`

the eigenvalues found are correctly :
Complex { re: -0.8611363115940531, im: 0.0 }, 
Complex { re: 0.8611363115940531, im: 0.0 }, 
Complex { re: -0.33998104358485637, im: 0.0 }, 
Complex { re: 0.33998104358485637, im: 0.0 }

### DIFF
--- a/src/linalg/schur.rs
+++ b/src/linalg/schur.rs
@@ -348,6 +348,9 @@ where
         DefaultAllocator: Allocator<DimDiff<D, U1>>,
     {
         let mut n = end;
+        // Equivalent to SMLNUM in LAPACK DLAHQR. Since SMLNUM depends on the machine
+        // precision, we use eps^2 here as best approximation.
+        // This is justified because the relative threshold use eps * ( t[n,n] + t[m,m] )
         let absolute_threshold = eps.clone() * eps.clone();
 
         while n > 0 {


alexhroom:
hey there, just approved the CI runs for you. I've got a pretty big stockpile of matrices that crash with the issue so next week I'll do some more robust checking that the issue's been sorted.

Thanks for figuring it out by the way, I had a go myself ages ago but couldn't quite figure out the problem...

Toucan4Life:
I looked at LAPACK DLAHQR subroutine, and they use SMLNUM as their absolute criteria : 
https://www.netlib.org/lapack/lapack-3.1.1/html/dlahqr.f.html.

SMLNUM is defined as the smallest possible value you can represent using the machine precision * number of rows of the matrix / epsilon. 

The issue is that depends on the machine precision and I didn't find an easy way to do that in rust. So I choose eps * eps, because the relative test for deflation was already eps * ( t[n,n] + t[m,m] )

alexhroom:
great - could I ask you to add that citation as a comment to the routine, and then we should be all good to go?

PR #1574, opened 2025-11-09
Cholesky decomposition: Proper check for positive definiteness

### LINKED ISSUE #1536: Cholesky erroneously produces (wrong) results for non-positive-definite complex matrices
See the following example:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=03236050db62b06a5a5a35b5438b0bad

The matrix here is
```
  ┌                                                                                                                               ┐
  │                   -4 0.000000000000000122 0.000000000000000122                    0                    0                    0 │
  │ 0.000000000000000122                   -4                    2                    0                    0                    0 │
  │ 0.000000000000000122                    2                   -4                    0                    0                    0 │
  │                    0                    0                    0                   -4 0.000000000000000122 0.000000000000000122 │
  │                    0                    0                    0 0.000000000000000122                   -4                    2 │
  │                    0                    0                    0 0.000000000000000122                    2                   -4 │
  └                                                                                                                               ┘
```
which is not positive definite. When the datatype of the matrix is `f64`, `Cholesky::new` correctly returns `None`. However, if I provide the same matrix but where the entries have the datatype `Complex<f64>`, `Cholesky::new` gives a result, which it shouldn't - if a matrix is not positive definite over the reals it certainly isn't positive definite over the complex plane. Furthermore, this result is not correct, as if I calculate $LL^*$, I get

```
  ┌                                                                                                                                                                                                                                                                                                                                                                                                                                                       ┐
  │                                                                     4+0i -0.000000000000000122+0.00000000000000000000000000000001494069094959771i -0.000000000000000122+0.00000000000000000000000000000001494069094959771i                                                                     0+0i                                                                     0+0i                                                                     0+0i │
  │ -0.000000000000000122-0.00000000000000000000000000000001494069094959771i                                                                     4+0i                                   -2+0.00000000000000024492935982947064i                                                                     0+0i                                                                     0+0i                                                                     0+0i │
  │ -0.000000000000000122-0.00000000000000000000000000000001494069094959771i                                   -2-0.00000000000000024492935982947064i                                                     6.000000000000001+0i                                                                     0+0i                                                                     0+0i                                                                     0+0i │
  │                                                                     0+0i                                                                     0+0i                                                                     0+0i                                                                     4+0i -0.000000000000000122+0.00000000000000000000000000000001494069094959771i -0.000000000000000122+0.00000000000000000000000000000001494069094959771i │
  │                                                                     0+0i                                                                     0+0i                                                                     0+0i -0.000000000000000122-0.00000000000000000000000000000001494069094959771i                                                                     4+0i                                   -2+0.00000000000000024492935982947064i │
  │                                                                     0+0i                                                                     0+0i                                                                     0+0i -0.000000000000000122-0.00000000000000000000000000000001494069094959771i                                   -2-0.00000000000000024492935982947064i                                                     6.000000000000001+0i │
  └                                                                                                                                                                                                                                                                                                                                                                                                                                                       ┘
```

which is not my original matrix.

I believe the problem here is that the [only way that we are checking for positive definiteness in the Cholesky implementation is by seeing whether non-zero entries on the diagonal produce a square root](https://github.com/dimforge/nalgebra/blob/8964bf28a8d7ca90e8886b81337585ceb6cd2fc4/src/linalg/cholesky.rs#L237-L260) - but for complex numbers this is always true!

PaulXiCao:
Closes #1536.

Proper check for positive definite matrix by checking diagonal entries of intermediate matrix. Added mentioned example from issue as test.

### DIFF
--- a/src/linalg/cholesky.rs
+++ b/src/linalg/cholesky.rs
@@ -235,10 +235,16 @@ where
             }
 
             let sqrt_denom = |v: T| {
-                if v.is_zero() {
+                // positive definiteness requires the diagonal to be real and strictly positive
+                if !v.clone().imaginary().is_zero() {
                     return None;
                 }
-                v.try_sqrt()
+                let re = v.real();
+                if re <= T::RealField::zero() {
+                    return None;
+                }
+                // fixme: could be improved by sqrt on real, e.g. `T::from_real(re.sqrt())`
+                Some(T::from_real(re).sqrt())
             };
 
             let diag = unsafe { matrix.get_unchecked((j, j)).clone() };
--- a/tests/linalg/cholesky.rs
+++ b/tests/linalg/cholesky.rs
@@ -11,6 +11,20 @@ fn cholesky_with_substitute() {
     assert!(na::Cholesky::new_with_substitute(m, 1e-8).is_some());
 }
 
+// 2x2 matrix filled with -1 is not positive definite and should not return a Cholesky
+// decomposition (regardless of element type).
+// See https://github.com/dimforge/nalgebra/issues/1536 .
+#[test]
+fn cholesky_check_positive_definiteness() {
+    let x = -1.0;
+    let m = na::Matrix2::new(x, x, x, x);
+    assert!(nalgebra::Cholesky::new(m).is_none());
+
+    let x = na::Complex::new(-1.0, 0.0);
+    let m = na::Matrix2::new(x, x, x, x);
+    assert!(nalgebra::Cholesky::new(m).is_none());
+}
+
 macro_rules! gen_tests(
     ($module: ident, $scalar: ty) => {
         mod $module {


src/linalg/cholesky.rs
```diff
@@ -235,10 +235,16 @@ where
...
-                if v.is_zero() {
+                // positive definiteness requires the diagonal to be real and strictly positive
+                if !v.clone().imaginary().is_zero() {
                     return None;
                 }
-                v.try_sqrt()
+                let re = v.real();
+                if re <= T::RealField::zero() {
+                    return None;
+                }
```
alexhroom, line 245: it would be clearer to write this as something like
```rust
let re = v.real();
let im = v.imaginary();

if re <= T::RealField::zero() || !im.is_zero() {
    return None;
}
```

alexhroom (changes requested):
thanks for the contribution! just a small tidy-up, and it seems the CI is failing with creating some `None`s that it isn't expecting.

alexhroom:
superseded by #1592

Issue #1422, opened 2024-06-28
New stack macro implementation triggers clippy::toplevel_ref_arg warning


dsemi:
[This](https://github.com/dimforge/nalgebra/blob/48c8f6a5053fac48e305f5e5e0004a332811d092/nalgebra-macros/src/stack_impl.rs#L29) line triggers the `clippy::toplevel_ref_arg` warning for users of `stack!`.

Example warning:
```
warning: `ref` on an entire `let` pattern is discouraged, take a reference with `&` instead
  --> src/main.rs:35:15
   |
35 |       let rhs = stack![ bp.cross(&bv) - ap.cross(&av);
   |  _______________^
36 | |                       cp.cross(&cv) - ap.cross(&av) ];
   | |_____________________________________________________^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#toplevel_ref_arg
   = note: `#[warn(clippy::toplevel_ref_arg)]` on by default
   = note: this warning originates in the macro `stack` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Andlon:
Uhm, does `clippy` normally give warnings from code generated by macros in external crates? That seems... excessive? Did you enable any particular options?

dsemi:
It seems so: https://github.com/rust-lang/rust-clippy/issues/702. I don't understand why that decision was made.

Andlon:
[This comment](https://github.com/rust-lang/rust-clippy/issues/702#issuecomment-546156574) suggests that it's due to our use of spans in the `stack!` macro to improve error messages. We should fix this by just plastering `#[allow(clippy::all)]` in the macro-generated code.

I'm not sure off the top of my head if we can put these attributes on code blocks though (not a function/impl block).

danieleades:
> We should fix this by just plastering `#[allow(clippy::all)]` in the macro-generated code.

I would absolutely not do that. That will suppress all future clippy lints, some of which might be useful.

i think the (reasonable) options are:

1. address the lint
2. suppress specifically this lint in the smallest possible scope
3. suppress all lints, but have CI in this repo that would catch new clippy warnings in similar code

I'm not even sure how option 3. would work to be honest.

Andlon:
[...]

PR #1580, opened 2025-12-19
Fix `clippy::toplevel_ref_arg` at stack! callsites

CattleProdigy:
`let x = &y` is preferred over `let ref x = y`

fixes https://github.com/dimforge/nalgebra/issues/1422


Ralith (approved):
LGTM, though I'm not familiar with this code so I'll let someone else merge.

Issue #1001, opened 2021-10-03
Unsound `Send` implementation for `SliceStorage`


steffahn:
This impl: https://docs.rs/nalgebra/0.29.0/nalgebra/base/struct.SliceStorage.html#impl-Send allows triggering UB in safe code e.g. as follows

```rs
#![allow(clippy::type_complexity)]

use std::{cell::Cell, thread};

use nalgebra::vector;

fn main() {
    create_data_race(unsound_share_non_sync);
}

fn unsound_share_non_sync<T: 'static + Send /* !Sync */>(
    t: T,
    callback1: fn(&T),
    callback2: fn(&T),
) {
    let v = vector![t];
    let v_ref_static = Box::leak(Box::new(v));
    let slice1 = v_ref_static.slice_range(.., ..);
    let slice2 = v_ref_static.slice_range(.., ..);
    let t1 = thread::spawn(move || {
        callback1(&slice1[0]);
    });
    let t2 = thread::spawn(move || {
        callback2(&slice2[0]);
    });

    t1.join().unwrap();
    t2.join().unwrap();
}

fn create_data_race(
    execute_on_two_threads: fn(
        Cell<&'static [&'static u8]>,
        fn(&Cell<&'static [&'static u8]>),
        fn(&Cell<&'static [&'static u8]>),
    ),
) {
    const LONG: &[&u8] = &[&42];
    const SHORT: &[&u8] = &[];
    let c = Cell::new(LONG);
    execute_on_two_threads(
        c,
        |c| loop {
            c.set(SHORT);
            c.set(LONG);
        },
        |c| loop {
            let r = c.get();
            if let Some(s) = r.get(0) {
                dbg!(**s);
            }
        },
    );
}
```
```
   Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 1.21s
     Running `target/debug/playground`
[src/main.rs:50] **s = 42
[src/main.rs:50] **s = 42
[src/main.rs:50] **s = 42
[src/main.rs:50] **s = 42
[src/main.rs:50] **s = 42
[src/main.rs:50] **s = 42
[src/main.rs:50] **s = 42
[src/main.rs:50] **s = 42
[src/main.rs:50] **s = 42
[src/main.rs:50] **s = 42
timeout: the monitored command dumped core
/playground/tools/entrypoint.sh: line 11:     8 Segmentation fault      timeout --signal=KILL ${timeout} "$@"
```
[...]

PR #1581, opened 2026-01-04
Fix T: Sync bound of impl Send for ViewStorage

197g:
Closes: #1001 

Note that this is a breaking change. The precondition differs between `ViewStorage` and `ViewStorageMut` just like they differ between `&mut T` and `& T`. While both allow reborrowing a `&T` from a shared reference to them, i.e. they have the same requirements for their `Sync` implementations, the formalism is different for the `Send` trait:
- a unique reference makes the value itself available and its creation and lifetime requires synchronization itself, hence `Send` where `T: Send`.
- a shared reference does not imply any synchronization point in its creation or during its lifetime so we must only make it available to parallel use (sendable to other threads) if the underlying type handles that parallelism, hence `Send` where `T: Sync`.

See: https://doc.rust-lang.org/stable/std/primitive.reference.html#trait-implementations-1

### DIFF
--- a/src/base/matrix_view.rs
+++ b/src/base/matrix_view.rs
@@ -33,14 +33,6 @@ macro_rules! view_storage_impl (
         #[deprecated = "Use ViewStorage(Mut) instead."]
         pub type $legacy_name<'a, T, R, C, RStride, CStride> = $T<'a, T, R, C, RStride, CStride>;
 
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Thanks! I won't merge this personally so it doesn't catch @sebcrozet by surprise, but we should definitely take this for the next breaking release.

PR #1591, opened 2026-03-16
feat: Add Bunch–Kaufman LDL Decomposition

awinick:
This PR adds an LDL decomposition for Hermitian matrices using the Bunch–Kaufman symmetric pivoting algorithm.

The factorization computed is

    Pᵀ A P = L D Lᴴ

where

- P is a permutation matrix from symmetric pivoting
- L is unit lower triangular
- D is Hermitian block diagonal containing 1×1 and 2×2 blocks

This corresponds to the class of algorithms implemented in LAPACK’s *HETRF family.

## Implementation notes

The factorization is stored in-place in a single matrix following the standard LAPACK layout:

- the strictly lower triangle stores L
- the diagonal and first subdiagonal store D
- pivot information is stored separately

## Tests

The tests are intentionally compact but exercise a wide variety of pivot behaviors.

During development I inspected pivot structures and found that matrices constructed as

    A = U D Uᴴ

where U is Haar-random unitary and D has alternating signs consistently generate rich mixtures of 1×1 and 2×2 pivot blocks.

Two spectra are used.

Alternating ±1 spectrum

    D = diag(1, -1, 1, -1, ...)

This produces a diverse set of pivot patterns across matrix sizes and verifies

- reconstruction accuracy
- determinant correctness
- linear solve correctness

Alternating geometric spectrum

    D = diag(1, -10, 100, -1000, ...)

This stresses the algorithm under extreme scaling, which is particularly important for

- pivot selection behavior
- numerical stability
- determinant accuracy

Empirically this spectrum produces a wide variety of pivot configurations while remaining numerically well-behaved.

## Determinant accuracy

During testing it was observed that computing determinants through this LDL factorization produces significantly more accurate real-valued results for Hermitian matrices than the current generic approach.

Because the decomposition explicitly respects Hermitian structure, the determinant is obtained as the product of real block determinants, which reduces complex roundoff artifacts.

## Benchmarking

Benchmarks were performed comparing this implementation against LAPACK (`zhetrf`) and the existing QR decomposition in nalgebra.

All results below correspond to **1000 factorizations**.

| Size | LAPACK zhetrf (Accelerate) | LAPACK zhetrf (OpenBLAS) | This implementation | nalgebra QR |
|-----|-----------------------------|--------------------------|--------------------|-------------|
| 100×100 | ~136 ms | ~117 ms | ~179 ms | ~529 ms |
| 200×200 | ~590 ms | ~877 ms | ~1274 ms | ~4249 ms |

The pure Rust implementation is typically within roughly a factor of two of hardware-accelerated LAPACK. Given that LAPACK implementations rely on highly tuned BLAS kernels and architecture-specific optimizations, this level of performance is reasonable for a portable pure-Rust implementation.

For comparison, QR factorization, the current best available method, is substantially slower. QR does not exploit Hermitian structure and therefore performs significantly more work than an LDL factorization.


src/linalg/ldl.rs
```diff
@@ -0,0 +1,450 @@
...
+            pivots,
+            zero_pivot,
+        }
+    }
+
+    /// The pivot data for this factorization.
+    ///
+    /// This uses the LAPACK `ipiv` convention:
+    /// a positive entry denotes a 1x1 pivot block, while a repeated negative entry
+    /// denotes a 2x2 pivot block. The stored pivot indices are 1-based.
```
geo-ant: I like LAPACK as much as the next guy, but to my mind this isn't going to feel intuitive for Rust users.

* Is there a reason to expose this API at all? 
* If so, would you mind abstracting it into a Pivot structure? This could expose an API for applying `P A P^T`, either as a function or using lazily overloaded operators depending how overengineered you'd like it to be... 😅 
* I'd much prefer 0-based indices, which is what `nalgebra` uses everywhere else.

src/linalg/ldl.rs
```diff
@@ -0,0 +1,450 @@
...
+    #[inline]
+    pub fn pivots(&self) -> &[isize] {
+        self.pivots.as_slice()
+    }
+
+    /// The first exactly zero pivot, if one was encountered.
+    ///
+    /// This follows the LAPACK convention and is therefore 1-based.
+    /// A value of `None` means no zero pivot was detected.
+    #[inline]
```
geo-ant: same comment as above, this could better be abstracted into a `Pivot` structure, if it's needed at all...

src/linalg/ldl.rs
```diff
@@ -0,0 +1,450 @@
...
+where
+    T::RealField: Copy,
+    DefaultAllocator: Allocator<N> + Allocator<N, N>,
+{
+    /// Compute the factorization of a complex Hermitian matrix using the Bunch-Kaufman
+    /// block-diagonal pivoting method:
+    ///
+    /// P A P^T = L * D * L^H
+    ///
+    /// where L is a product of permutation and unit lower triangular matrices, U^H is the
```
geo-ant: > where L is a product of permutation and unit lower triangular matrices, U^H is the

I think this is copied from [LAPACK's ?{he,sy}trf](https://netlib.org/lapack/explore-html/d8/d0e/group__hetrf_ga114675389727c72322841cef488c10dd.html#ga114675389727c72322841cef488c10dd) family of functions, but lapack uses the form `A = L  D  L^H`, whereas you give `P A P^T = L D L^H`. In your case `L` should indeed be _unit_ lower triangular, right?

src/linalg/ldl.rs
```diff
@@ -0,0 +1,450 @@
...
+    /// The first exactly zero pivot, if one was encountered.
+    ///
+    /// This follows the LAPACK convention and is therefore 1-based.
+    /// A value of `None` means no zero pivot was detected.
+    #[inline]
+    pub const fn zero_pivot(&self) -> Option<usize> {
+        self.zero_pivot
+    }
+
+    /// The permutation-aware factor of this decomposition.
```
geo-ant: Is this essentially `P^T L`? If so could you just state that in the comment?

geo-ant (commented):
Hey, I'm sorry that I didn't have time to take a super in-depth look yet. I need to brush up on the Bunch-Kaufman factorization.

Did you use LAPACKS `zhetrf` function as your reference? If not, what other references should I use to review the implementation. I'm aware of:

* [Bunch & Kaufmann 77](https://www.ams.org/journals/mcom/1977-31-137/S0025-5718-1977-0428694-0/S0025-5718-1977-0428694-0.pdf), which present a couple of algorithms. Which one should I be looking at?
* [Jones & Patrick 89](https://apps.dtic.mil/sti/tr/pdf/ADA211415.pdf), which present an algorithm in section 4 of their paper.

Let me know what would be a good reference for me to check your implementation. I see you've added tests which is already fantastic, I'd just like to be able to review your implementation as well.

I've made some comments in the code for some API related things. If you want to take care of them, could you also run `cargo fmt` on your code so it'll pass the lint?

src/linalg/ldl.rs
awinick: Yes LAPACK’s ?sytrf/?hetrf routines were the primary reference for the implementation.

I wrote the factorization in the form
P A P^T = L D L^H intentionally. In this formulation, L is unit lower triangular. LAPACK’s documentation writes A = L D L^H, but in practice the permutations are absorbed into the pivot representation, so the stored L is not strictly lower triangular in the original coordinate system.

Keeping the permutation explicit via P makes the structure a bit clearer: L is unit lower triangular in the permuted basis, and the pivoting is handled separately rather than implicitly encoded.

src/linalg/ldl.rs
awinick: That makes sense. I agree the LAPACK-style encoding is efficient but not very intuitive.

I’ve switched to a `(pivot_index, block_size)` representation with 0-based indices, so the semantics are explicit and consistent with nalgebra, without the sign/1-based encoding.

Right now I’m exposing the pivots mainly for inspection/testing rather than as a primary API. I didn’t introduce a dedicated `Pivot` type to keep things lightweight, but I’m happy to revisit that if you think it adds value.

I don’t expect most users to interact with `pivots` or `zero_pivot `directly, but exposing them gives some additional visibility for debugging (e.g. understanding how the factorization pivoted, or diagnosing singular/near-singular cases).

geo-ant: I don't have a super strong opinion here (just regular strength 😆), but if the API is exposed at all I'd like it to be simpler (meaning using some kind of wrapper type). But from what you said the API won't be touched by most users anyway so I'd say my preferred solution right now is to make it private and use it only in the tests. If people ask for it, there's always a chance to expose something that's easier to use. 

If you say that the pivot give us a way to understand whether a matrix is (near) singular, maybe that's the only thing that should be exposed for now.

src/linalg/ldl.rs
geo-ant: Okay then I understood your implementation correctly and I actually like the clarity. 

But then the comment needs to change, if I'm not missing something? Because the comment says the permutation is absorbed into L in your implementation, which it wouldn't be. 

Edit: I still want to check your implementation against the lapack impl.

awinick: > Okay then I understood your implementation correctly and I actually like the clarity.
> 
> But then the comment needs to change, if I'm not missing something? Because the comment says the permutation is absorbed into L in your implementation, which it wouldn't be.
> 
> Edit: I still want to check your implementation against the lapack impl.

Good catch 😅 I double-checked the docs, and they should be consistent and easier to understand now.

src/linalg/ldl.rs
awinick: On `zero_pivot`, I’d be a little careful about treating it as a general statement about singularity or conditioning. An exactly zero pivot means `solve`/`solve_mut` cannot proceed reliably, so it is useful diagnostic information, but the converse is not true: a badly conditioned or even singular matrix can still sometimes produce a result, and that result has to be interpreted with care.

So I see `zero_pivot` as information about how the factorization behaved, not as a full statement about conditioning or singularity. More generally, I think assessing conditioning is outside the responsibility of the decomposition itself. If nalgebra wants to expose that kind of information, I think it would make more sense as a separate method rather than through `zero_pivot`. I’d definitely find that useful myself when evaluating problems.

On the API, I’m open to making the pivot representation non-public. The main reason I exposed it was for inspection/testing, since I’ve been keeping tests in integration tests rather than unit tests, so private or `pub(crate)` internals are not accessible there.

Would hiding it from the public docs be sufficient here, or would you prefer a different approach?

awinick: Also, if a dedicated conditioning-related method would be useful, I’d be happy to open that as follow-up work.

src/base/matrix_view.rs
geo-ant, line 144: Hey, can you quickly explain why I'm seeing this change here?

geo-ant, line 167: Same question as above, why is there a change here...?

src/linalg/decomposition.rs
```diff
@@ -244,6 +244,7 @@ impl<T: ComplexField, R: Dim, C: Dim, S: Storage<T, R, C>> Matrix<T, R, C, S> {
 /// | -------------------------|---------------------------|--------------|
 /// | Hessenberg               | `Q * H * Qᵀ`             | `Q` is a unitary matrix and `H` an upper-Hessenberg matrix. |
 /// | Cholesky                 | `L * Lᵀ`                 | `L` is a lower-triangular matrix. |
+/// | LDL decomposition        | `Pᵀ * L * D * Lᴴ * P`    | `L` is unit lower-triangular, `D` is Hermitian block-diagonal, and `P` is a permutation matrix. |
```
geo-ant: I don't want to start bike-shedding here, but I like the comment by @alexhroom https://github.com/dimforge/nalgebra/pull/1515#discussion_r2554123336 over in the other PR. They point out that `faer-rs` has both the strictly diagonal decomposition `A = L D L^H`, where `D` is indeed diagonal (`ldlt`). In the block-diagonal case (as is the case here), they call the decomposition `lblt`: `P A P^T = L B L^H`.

If you would be open to renaming your decomposition to `lblt`/`LBLT`, then we have an elegant way of distinguishing it from the work at https://github.com/dimforge/nalgebra/pull/1515. The other advantage would be that this is consistent with how `faer-rs` does it and I do like users to have a good time when switching backends.

src/linalg/decomposition.rs
```diff
@@ -260,6 +261,18 @@ impl<T: ComplexField, D: Dim, S: Storage<T, D, D>> Matrix<T, D, D, S> {
         Cholesky::new(self.into_owned())
     }
 
+    /// Computes the LDL decomposition of this matrix.
+    /// The input matrix `self` is assumed to be Hermitian (symmetric) and the decomposition will
+    /// only read the lower-triangular part of `self`.
+    pub fn ldl(self) -> LDL<T, D>
```
geo-ant: Just in case you agree to rename this decomposition, this method would then have to be named `lblt`, but many more documentation and naming changes would be necessary. I won't comment on them anymore :D

src/linalg/symmetric_eigen.rs
geo-ant, line 222: I see the comment, but could you quickly explain the change? I can see below that it is related to https://github.com/dimforge/nalgebra/issues/1109, I'd just like a quick explanation of why this fixes it.

geo-ant, line 366: implement a fix AND write a test for it? You're spoiling us. Good job! ❤️ 

One nitpick is that you probably don't need a doc-comment on the test itself, an in-code comment is fine. I don't know if cargo doc will complain about this?? Meaning replacing `///` with `//` should do.

tests/linalg/ldl.rs
```diff
@@ -0,0 +1,122 @@
...
+    let mut rng = StdRng::seed_from_u64(0);
+
+    for n in 2..=10 {
+        let n_i32 = i32::try_from(n).unwrap();
+        let diag = DVector::from_iterator(n, (0..n_i32).map(|i| (-1.0).powi(i)));
+
+        for _ in 0..10 {
+            let matrix: DMatrix<Complex<f64>> = random_hermitian_from_diag(&diag, &mut rng);
+            let ldl = matrix.clone().ldl();
+            assert!(relative_norm(&matrix, &reconstruct(&ldl)) < 1e-12);
```
geo-ant: why can't you use `assert_relative_eq!` here? It's not that this looks super problematic, but I feel element wise differences are probably safer, even if the epsilon needs to be a little more generous. Otherwise a noticeable difference in one element could be offset by a large enough matrix norm. Please correct me, I might absolutely be missing something here.

tests/linalg/ldl.rs
```diff
@@ -0,0 +1,122 @@
...
+
+            assert_relative_eq!(
+                ldl.determinant(),
+                (-1.0).powi(n_i32 / 2),
+                max_relative = 1e-12
+            );
+
+            let b = random_isometry(Dyn(n), Dyn(n), &mut rng);
+            assert!((&matrix * &ldl.solve(&b).unwrap() - &b).norm() / matrix.norm() < 1e-12);
```
geo-ant: You want to check (in spirit) if $A X \approx B$, where $X$ is your calculated solution. What you're checking is $\lVert A X-B \rVert < \epsilon \lVert A \rVert$. So if $A$ is invertible, obviously we can check $A X \approx B$ within numerical accuracy. But if $A$ is singular, then this check won't work and maybe that's what you're trying to address with this?

If $A$ is singular, then the LBLT decomposition solves the system in a least squares sense $\min_{X} \lVert A X - B \rVert^2_2$. I think what you should do in that case is check whether the normal equations are satisfied: $A^H (A X -B) = \boldsymbol{0}$, where $\boldsymbol{0}$ is a matrix of all zeros. So I think the correct check to implement would be to use the element-wise assert relative eq on the normal equations. I think that's a stronger check than what you currently have. And it works whether A is singular or not. So you could check $A^H A X \approx A^H B$ using element wise equality. The fact that $A^H = A$ makes this even easier to implement :grin:

tests/linalg/ldl.rs
```diff
@@ -0,0 +1,122 @@
...
+    let mut rng = StdRng::seed_from_u64(0);
+
+    for n in 2..=10 {
+        let n_i32 = i32::try_from(n).unwrap();
+        let diag = DVector::from_iterator(n, (0..n_i32).map(|i| (-10.0).powi(i)));
+
+        for _ in 0..10 {
+            let matrix: DMatrix<Complex<f64>> = random_hermitian_from_diag(&diag, &mut rng);
+            let ldl = matrix.clone().ldl();
+            assert!(relative_norm(&matrix, &reconstruct(&ldl)) < 1e-12);
```
geo-ant: same comment as in the other function (element wise relative eq)

tests/linalg/ldl.rs
```diff
@@ -0,0 +1,122 @@
...
+
+            assert_relative_eq!(
+                ldl.determinant(),
+                (-10.0).powi(n_i32 * (n_i32 - 1) / 2),
+                max_relative = 10.0.powi(-11 + (n_i32 / 2))
+            );
+
+            let b = random_isometry(Dyn(n), Dyn(n), &mut rng);
+            assert!((&matrix * &ldl.solve(&b).unwrap() - &b).norm() / matrix.norm() < 1e-12);
```
geo-ant: same comment as in the other function (normal equations)

geo-ant (changes requested):
Hey, I haven't yet reviewed the implementation itself. But that's the last thing I need to do. I hope to do that soon (for some definition of soon). In the meantime I've taken a look at the other changes. This is shaping up really nicely, but I did spot some things in the tests that I believe should be addressed. Feel free to address the changes now (or whenever you have time) or to wait for the review of the implementation itself. And thanks for your continued work on this!!

src/linalg/symmetric_eigen.rs
awinick: I didn't touch this code. It looks like another PR is doing this..?

awinick: Not me

src/base/matrix_view.rs
awinick: Not me

awinick: Not me

tests/linalg/ldl.rs
awinick: We could certainly use `assert_relative_eq!` for the alternating-spectrum case. I’d lean toward leaving the reconstruction check normwise in both tests to keep the interpretation uniform, but I could change that one if you prefer.

For the geometric-spectrum case, though, I think a normwise check is the more appropriate criterion. This test is deliberately badly scaled, so a fixed absolute entrywise comparison is not very informative. In practice, making `assert_relative_eq` reliable there would require loosening it to roughly `epsilon = 1e-6`, which feels too weak to say much about the quality of the factorization.

More importantly, for Bunch–Kaufman the natural lens is normwise backward stability rather than uniformly small entrywise errors. What I want to check in this test is essentially that the computed factors reconstruct a nearby matrix with small relative error, i.e. that `‖A - LBLᴴ‖ / ‖A‖` is small. That is the stability notion usually associated with this factorization [1], whereas a fixed absolute elementwise tolerance is much more sensitive to the scaling of individual entries.

[1]: https://epubs.siam.org/doi/10.1137/S0895479896296921 "Accurate Symmetric Indefinite Linear Equation Solvers"

tests/linalg/ldl.rs
awinick: I think this is a good point in the abstract, but for these tests I don’t think the normal-equations check is the right replacement.

In the alternating-spectrum case, the matrix is nonsingular by construction, so checking `AX ≈ B` directly is the more natural condition, and it is strictly stronger than checking `Aᴴ(AX - B) ≈ 0`.

For the geometric-spectrum case, I agree the matrix can be extremely ill-conditioned in floating point, but this implementation also does not intentionally compute a least-squares solution for singular systems. It only reports failure when a pivot is exactly zero. So I don’t think the right interpretation of a successful solve here is “the least-squares normal equations are satisfied”; rather, the intent is still to check that the returned solution gives a small residual.

So my preference would be to keep the solve check as a direct residual / `AX ≈ B` check, and reserve the normal-equations condition for a solver that is explicitly intended to return least-squares solutions in the singular case.

src/linalg/decomposition.rs
awinick: Sure, `lblt` works for me. I’m pretty indifferent on the naming, and I agree it’s a cleaner way to distinguish this block-diagonal factorization from the strictly diagonal `ldlt` work in #1515. Consistency with faer-rs also seems like a reasonable choice here.

tests/linalg/ldl.rs
geo-ant: Hey, I'm really sorry I'm being so annoying here and I'm also reasonably confident that everything you're saying is correct, but I want to make sure we're not making a mistake here. You're right that the paper introduces the stability using $\lVert A - L D L^H\rVert$. However, not quite as you're using it here.

In Table 2.4 they introduce the **decomposition error** and the **solution error** as

$$\lVert A - L D L^H \rVert_1/(n \epsilon \lVert A \rVert_1)$$

and 

$$\lVert x - x^* \rVert / (\lVert x \rVert_1 \epsilon \kappa_1(A)),$$

With the condition number $\kappa_1 (A) = \lVert A\rVert_1 \lVert A^-1 \rVert_1$, $\epsilon$ as machine epsilon and the dimension $n$. They say this is only a problem If those values are much bigger than 1, since those are conservative bounds. They never explicitly define $\lVert . \rVert_1$ (or I missed it) but their definition of $\kappa_1$ suggests that this is the 1-norm as defined e.g. [here](https://en.wikipedia.org/wiki/Matrix_norm#p_=_1_or_%E2%88%9E). Since you're using the `Matrix::norm` function, you are calculating the **Frobenius Norm** of a the respective Matrix, which isn't the L2-norm (spectral norm of a matrix) despite the `nalgebra` documentation claiming that it is.

I think your hard coded bounds are fine, rather than using $epsilon$ and the condition number. I can't see them disclosing what much bigger than 1 means exactly, anyways. Do you think you could try the using the 1-norm here? Or do you think the way you did it is equivalent and I'm completely missing the mark here?

Also your test for the accuracy of the solution doesn't quite reproduce the criterion in the paper. You're testing $\lVert Ax^* - b\rVert$ is small whereas they are testing $\lVert x - x^*\rVert$ is small (for some measure of small, differences in norms notwithstanding). In spirit, this is of course the same test, but I am not sure that they will always mean the same in practice, especially considering you're using the F-Norm in the current impl... What do you think? If you went to single right hand sides, you'll be able to use the existing `Vector::norm`, since the F-norm and the 2-norm are the same for vectors. It'll still be nice to have a test in there for multiple right hand sides, but that might be a simple manually verified unit test, just to make sure the functionality works.

Please let me know what you think.

src/base/matrix_view.rs
geo-ant: this is likely related to the force push of main, which you didn't do. If you can, could you update your branch on top of the current main, but don't spend much time on this if it gets annoying.

src/linalg/lblt.rs
```diff
@@ -0,0 +1,502 @@
...
+pub struct LBLT<T: ComplexField, N: Dim>
+where
+    DefaultAllocator: Allocator<N> + Allocator<N, N>,
+{
+    matrix: OMatrix<T, N, N>,
+    pivots: OVector<Pivot, N>,
+    zero_pivot: Option<usize>,
+}
+
+impl<T: Copy + ComplexField, N: Dim> LBLT<T, N>
```
geo-ant, line 40: this looks good to me when compared to the literature reference you gave. I don't know if Bunch-Kaufmann factorization is universally implemented as Algorithm A from the paper, since they give other variations as well. Might be nice to stick it in the docs somewhere.

src/linalg/lblt.rs
```diff
@@ -0,0 +1,502 @@
...
+            }
+
+            let pivot_index: usize;
+
+            if diag_abs >= alpha * colmax {
+                // The diagonal dominates column k strongly enough to use A[k, k] as a 1x1 pivot
+                // without any row/column interchange.
+                pivot_index = k;
+            } else {
+                let imax = imax.unwrap();
```
geo-ant: That's a thing I stumbled over, but I think it's correct. In the `None` case for `imax` we'll also have `colmax == 0.` such that the if-branch above will always be hit. I'm wondering if there's another way to express this without having an unwrap there. The easiest way is probably to invert the branching logic and branch on `if let Some(imax) && diag_abs < alpha*colmax`... Or to make the tuple (imax,colmax) into type `Option<usize,T::Realfield>` and refactor accordingly. But tbh I don't know if this produce better or worse assembly and whether it'll help the legibility. If you don't like those alternatives, I'm fine with sticking a comment here as well... It's just that this unwrap might come back to bite iff there's ever a refactoring...

geo-ant (commented):
okay. I think we're in a good place. If you could respond to the other questions, this should be the last round.

tests/linalg/ldl.rs
awinick: Thanks again for the detailed feedback.

I’ve updated the tests to use the induced matrix 1-norm for both the reconstruction and solve checks, which aligns better with the analysis in the paper. While all matrix norms in finite dimensions are equivalent up to constant factors, for fixed tolerances that equivalence isn’t numerically neutral, so it makes sense to match the norm used in the bounds.

This did require adding a `OneNorm` implementation. We already had a local version of this internally for the matrix exponential code, so I moved it into the core norms API since it’s generally useful. I’m also planning to add `exp_multiply` and matrix logarithm functionality, which both rely on the 1-norm, so this felt like a natural addition rather than a one-off for these tests.

For the solve accuracy, I’ve kept the residual-based check (‖AX − B‖ small). In these tests we don’t have access to the true solution x*, so measuring ‖x − x*‖ directly isn’t possible in general. The residual is the standard practical proxy here, and using multiple right-hand sides (sampling B as an n×n isometry) exercises the solve across a well-conditioned basis rather than a single vector.

More generally, there isn’t a single canonical way to assess solver accuracy in this setting. There are several equivalent formulations depending on what is being emphasized. In this case, I don’t see a strong reason to prefer one beyond aligning with the analysis and using a practical, well-established proxy for the solve.

awinick: I didn’t add a dedicated unit test for OneNorm. The implementation is quite direct, and it’s already exercised indirectly by the updated tests as well as existing matrix exponential tests that rely on the same computation.

That said, I’m happy to add a small explicit test if you think it would be useful.

src/base/norm.rs
```diff
@@ -120,6 +126,45 @@ impl<T: SimdComplexField> Norm<T> for LpNorm {
...
+            .zip(m2.column_iter())
+            .map(|(c1, c2)| {
+                c1.zip_fold(&c2, T::SimdRealField::zero(), |acc, a, b| {
+                    acc + (a - b).simd_modulus()
+                })
+            })
+            .fold(T::SimdRealField::zero(), T::SimdRealField::simd_max)
+    }
+}
+
```
geo-ant, line 192: I'm in favor of having this in here, but I'll page the owner of the library privately, since I don't feel comfortable deciding this alone. The problem is that the current `LpNorm` implementation looks deceptively similar and the struct is poorly documented. What the current `LpNorm` calculates is the _elementwise_ Lp norm and there should be a comment explaining this and giving a formula. Your documentation is much better, but still relies on everyone understanding what _induced_ norm means. Would you be so kind to add some docs to both?

geo-ant (approved):
Thank you for your work. @awinick if you would update the comments on the norms I'd be very grateful. I'll also ping the owner of the library privately, because I'd like them to weigh in on the norms. Everything else looks great to me, thanks for your patience!

geo-ant:
I just saw that some of your changes are incompatible with the MSRV, could you quickly address those?

src/base/norm.rs
awinick: I've updated the docs for both as requested, actually updated docs for all the norms with warnings when needed.

There's a meaningful inconsistency in how they all behave on matrices vs. vectors. In addition to what you said about `LpNorm`, `UniformNorm` has the same issue. `EuclideanNorm` on a matrix computes the Frobenius norm, not the spectral norm.

Making them align with user expectation would be a breaking change but would be cool to see at some point maybe haha.

geo-ant:
Thank you @awinick for your contribution and your continued work on this! Was a pleasure working with you.

PR #1595, opened 2026-03-28
Numerically stable implementation of axis, angle and scaled_axis for Rotation3 and UnitQuaternion

### LINKED ISSUE #1382: Numecirally vulnerable axis calculation in Rotation3
### Abstract

`nalgebra::Rotation3::axis` relies on numerically vulnerable calculation. 
It may return `None` in the place where it should return a non-None value.

### Issue

In [the current implementation of Rotation3](https://github.com/dimforge/nalgebra/blob/a803815bd1d22a052690cfde3ff7fae26cd91152/src/geometry/rotation_specialization.rs#L844), the rotation axis is calculated in the following way. 

```rs
    #[inline]
    #[must_use]
    pub fn axis(&self) -> Option<Unit<Vector3<T>>>
    where
        T: RealField,
    {
        let rotmat = self.matrix();
        let axis = SVector::<T, 3>::new(
            rotmat[(2, 1)].clone() - rotmat[(1, 2)].clone(),
            rotmat[(0, 2)].clone() - rotmat[(2, 0)].clone(),
            rotmat[(1, 0)].clone() - rotmat[(0, 1)].clone(),
        );

        Unit::try_new(axis, T::default_epsilon())
    }
```

If I initialize `Rotation3` with a rotation of π radians around the y-axis, the code will look like this.

```rs
use nalgebra::{Rotation3, Vector3};

const PI: f64 = std::f64::consts::PI;

fn main() {
    let r = Rotation3::from_axis_angle(&Vector3::y_axis(), PI);
    println!("axis = {:?}", r.axis());
    println!("axis_angle = {:?}", r.axis_angle());
}
```

And it actually works. 

```rs
axis = Some([[0.0, 1.0, 0.0]])
axis_angle = Some(([[0.0, 1.0, 0.0]], 3.141592653589793))
```

However, this is very vulnerable, because this behavior relies on the rounding error of the internal matrix representation.

I added the line to print the axis value in the `axis` function.

```rs
    #[inline]
    #[must_use]
    pub fn axis(&self) -> Option<Unit<Vector3<T>>>
    where  
        T: RealField,
    {
        let rotmat = self.matrix();
        let axis = SVector::<T, 3>::new(
            rotmat[(2, 1)].clone() - rotmat[(1, 2)].clone(),
            rotmat[(0, 2)].clone() - rotmat[(2, 0)].clone(),
            rotmat[(1, 0)].clone() - rotmat[(0, 1)].clone(),
        );
          
        println!("intelnal axis value = {:?}", axis);
        Unit::try_new(axis, T::default_epsilon())    
    }
```

And it said

```
intelnal axis value = [[0.0, 2.4492935982947064e-16, 0.0]]
```

This is very close to `[0, 0, 0]`. If there was no rounding error, this code would return `None` in the place where it should return `[0, 1, 0]`.

### Why this happens

The matrix corresponding to the rotation of π radians around the y-axis is, without rounding error, 

```
-1  0  0
 0  1  0
 0  0 -1
```

So if we calculate `axis` based on the matrix above, it becomes `[0, 0, 0]`.
The same happens for the rotation around the x-axis, z-axis, or rotation of π radians around any axis. Actually, on my Ubuntu desktop, the code below printed `None`. 

```rs
use nalgebra::{Rotation3, Vector3, Unit};

const PI: f64 = std::f64::consts::PI;

fn main() {
    let x = Unit::new_normalize(Vector3::new(1., 2., 1.));
    let r = Rotation3::from_axis_angle(&x, PI);
    println!("axis = {:?}", r.axis());
}
```

### Possible solution 

One possible solution is to handle these singular value cases as explained in this paper, section 9.4.1.2 "Logarithm map".
[A tutorial on SE(3) transformation parameterizations and on-manifold optimization](https://ingmec.ual.es/~jlblanco/papers/jlblanco2010geometry3D_techrep.pdf#page=49.32)

Another solution is to replace the internal rotation representation in `Rotation3` from a 3x3 matrix to a unit quaternion. 
The conversion from a unit quaternion to a rotation vector is described in the same section of the paper. 
This will improve the overall performance including rotation multiplication, but requires a large amount of modifications.

nmayorov:
Hi and thank you for a great library!

In this PR I want to resolve couple of issues regarding axis-angle representation of rotations.

# Conceptual change

The conceptual idea is that we want to use "scaled_axis" aka "rotation vector" as the core representation, without splitting it into axis and angle in internal computations. This is the numerical stable way to work with such representation and it gracefully handles the case of zero or close to zero rotation angles without introducing special cases, singularities and epsilon tresholds.

# What issues does it fix

It fixes the following practical issues and bugs:

- Loss of precision when creating from `scaled_axis` , i.e creating rotation from `Vector3::new(5.0e-20, 1e-16, -1.0e-17)` and then querying `scaled_axis` representation will return this vector, not vector of zeros
- Possibility of `angle` returned as `nan` from `Rotation3` due to `acos` argument going out `[-1, 1]` bounds after some minor round-off error accumulation (the issue I encountered myself)
- Unstable computation of rotation axis from `Rotation3` for angles close to pi (example in issue #1382 )

Basically all the operations regarding `scaled_axis` and `angle` are 100% robust, the `axis` extraction can be controlled by the user (i.e. for how small angles to extract rotation axis).

# How does it achieve that

The core change is in `Quaternion::exp` and `UnitQuaternion::scaled_axis` where scaling of the form `sin(theta) / theta` and its inverse is implemented via `sinc` without any epsilon threshold checking. After that `UnitQuaternion` functionality is solid.

And `Rotation` methods are delegated to `UnitQuaternion` as `rotation matrix -> unit quaternion -> scaled axis`. There are not many reasons or savings doing it in any other way, because scaled axis / rotation vector is directly connected to quaternion representation.

# Code concerns

I had to change trait bound of `T` to `RealField` in `Rotation3::angle`. I'm not an expert in Rust and don't know exact implications of that, but it seems reasonable to do, because all other functions related to axis and angle have this specification.

---- 

I plan to add more tests on the functionality, meanwhile, hopefully, someone review and consider these changes.

Closes #1382.

### DIFF
--- a/src/geometry/quaternion.rs
+++ b/src/geometry/quaternion.rs
@@ -974,8 +974,8 @@ impl<T: RealField + AbsDiffEq<Epsilon = T>> AbsDiffEq for Quaternion<T> {
     #[inline]
     fn abs_diff_eq(&self, other: &Self, epsilon: Self::Epsilon) -> bool {
         self.as_vector().abs_diff_eq(other.as_vector(), epsilon.clone()) ||
-            // Account for the double-covering of S², i.e. q = -q
-            self.as_vector().iter().zip(other.as_vector().iter()).all(|(a, b)| a.abs_diff_eq(&-b.clone(), epsilon.clone()))
+        // Account for the double-covering of S², i.e. q = -q
+        self.as_vector().iter().zip(other.as_vector().iter()).all(|(a, b)| a.abs_diff_eq(&-b.clone(), epsilon.clone()))
     }
 }
 
--- a/src/geometry/quaternion.rs
+++ b/src/geometry/quaternion.rs
@@ -516,7 +516,13 @@ where
     #[inline]
     #[must_use]
     pub fn exp(&self) -> Self {
-        self.exp_eps(T::simd_default_epsilon())
+        let v = self.vector();
+        let v_norm = v.norm();
+        let exp = self.scalar().simd_exp();
+        Self::from_parts(
+            exp.clone() * v_norm.clone().simd_cos(),
+            v * exp * v_norm.simd_sinc(),
+        )
     }
 
     /// Compute the exponential of a quaternion. Returns the identity if the vector part of this quaternion
@@ -968,8 +974,8 @@ impl<T: RealField + AbsDiffEq<Epsilon = T>> AbsDiffEq for Quaternion<T> {
     #[inline]
     fn abs_diff_eq(&self, other: &Self, epsilon: Self::Epsilon) -> bool {
         self.as_vector().abs_diff_eq(other.as_vector(), epsilon.clone()) ||
-        // Account for the double-covering of S², i.e. q = -q
-        self.as_vector().iter().zip(other.as_vector().iter()).all(|(a, b)| a.abs_diff_eq(&-b.clone(), epsilon.clone()))
+            // Account for the double-covering of S², i.e. q = -q
+            self.as_vector().iter().zip(other.as_vector().iter()).all(|(a, b)| a.abs_diff_eq(&-b.clone(), epsilon.clone()))
     }
 }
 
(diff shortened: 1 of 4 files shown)


nmayorov:
I've cleaned up the commits and added a test where two very close rotations are multiplied (second inverted) to get a tiny rotation which on master branch returns angle as `nan` (which is obviously super nasty for reliable usage), it works correctly after the changes.

I think this MR is good to go and it would be really useful for reliable usage of rotations from nalgebra.

PR #1603, opened 2026-06-02
add exceptional shift

### LINKED ISSUE #611: possible bug, hanging on getting eigenvalues?
Hi,

I’m trying to find the roots of a polynomial, and nalgebra has overall worked great.

Recently, I’ve found that my calculations hang on certain inputs, for which numpy immediately returns a value.

You can see my code for finding roots at https://github.com/hwchen/roots.

A pathological input is in this test case, which you can also see in the repo's unit tests:
```rust
   #[test]
    fn hanging() {
        let input = &[
            -770.9700059294701,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            0.0,
            434.3959909081459,
            0.0,
        ];

        let input: Vec<f64> = input.into_iter().rev().cloned().collect();
        dbg!(&input);
        let rs = roots(input.as_slice());
    }
```
Thanks for any help.

Toucan4Life:
### Summary of Changes
This PR fixes some cases where the search for eigenvalue could hang. This was solved by using the "Exceptional Shift" method in Wilkinson, J. H., & Reinsch, C. (1971). 
Handbook for automatic computation (Vol. 2: Linear algebra). 

### Details
I first checked why the Lapack algorithm would converge and not hang. I looked at dlahqr procedure (https://netlib.org/lapack/explore-html/d8/df4/group__lahqr_ga100049ab080cef756846702a33e8e28f.html) saw the mention of the exceptional shift and the Ahues & Tisseur stopping criterion.

The last one was a red herring, I implemented it but the test matrix from #611 would still hang. I then implemented the exceptional shift and it worked. I tried tracing back the magical value (0.75 & -0.4375), and discovered https://www.netlib.org/eispack/hqr.f that led me to the original source:

Wilkinson, J. H., & Reinsch, C. (1971). 
Handbook for automatic computation (Vol. 2: Linear algebra).

You can view the orignal magical number and algorithm implementation here : https://www.scribd.com/document/568239788/Wilkinson-Reinsch1971-Book-HandbookForAutomaticComputatio#page=373
 
AI was used to find the original paper

### Testing
I implemented a test with the matrix from #611 that now pass successfully

Fixes #611

### DIFF
--- a/src/linalg/schur.rs
+++ b/src/linalg/schur.rs
@@ -135,6 +135,10 @@ where
 
         // Implicit double-shift QR method.
         let mut niter = 0;
+        let mut kdefl = 0usize;
+        let kexsh = 10usize;
+        let exceptional_shift_a: T::RealField = crate::convert(0.75);
+        let exceptional_shift_b: T::RealField = crate::convert(-0.4375);
         let (mut start, mut end) = Self::delimit_subproblem(&mut t, eps.clone(), dim.value() - 1);
 
         while end != start {
@@ -150,10 +154,25 @@ where
                 let h22 = t[(start + 1, start + 1)].clone();
                 let h32 = t[(start + 2, start + 1)].clone();
 
-                let hnn = t[(n, n)].clone();
-                let hmm = t[(m, m)].clone();
-                let hnm = t[(n, m)].clone();
-                let hmn = t[(m, n)].clone();
+                let (hnn, hmm, hnm, hmn) = if kdefl != 0 && kdefl % kexsh == 0 {
+                    let s = t[(start + 1, start)].clone().norm1()
+                        + t[(start + 2, start + 1)].clone().norm1();
+                    let h = t[(start, start)].clone()
+                        + T::from_real(exceptional_shift_a.clone() * s.clone());
+                    (
+                        h.clone(),
+                        h,
+                        T::from_real(exceptional_shift_b.clone() * s.clone()),
+                        T::from_real(s),
+                    )
+                } else {
+                    (
+                        t[(n, n)].clone(),
+                        t[(m, m)].clone(),
+                        t[(n, m)].clone(),
+                        t[(m, n)].clone(),
+                    )
+                };
 
                 let tra = hnn.clone() + hmm.clone();
                 let det = hnn * hmm - hnm * hmn;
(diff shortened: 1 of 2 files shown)


Toucan4Life:
@alexhroom still one case of eigen value search hanging that should be solved now :)

src/linalg/schur.rs
```diff
@@ -135,6 +135,10 @@ where
 
         // Implicit double-shift QR method.
         let mut niter = 0;
+        let mut kdefl = 0usize;
+        let kexsh = 10usize;
+        let exceptional_shift_a: T::RealField = crate::convert(0.75);
+        let exceptional_shift_b: T::RealField = crate::convert(-0.4375);
```
alexhroom, line 147: can you add a comment which documents the source for these magic values (as given in your PR description) as well as what `kdefl` and `kexsh` mean ?

src/linalg/schur.rs
```diff
@@ -150,10 +154,25 @@ where
                 let h22 = t[(start + 1, start + 1)].clone();
                 let h32 = t[(start + 2, start + 1)].clone();
 
-                let hnn = t[(n, n)].clone();
-                let hmm = t[(m, m)].clone();
-                let hnm = t[(n, m)].clone();
-                let hmn = t[(m, n)].clone();
+                let (hnn, hmm, hnm, hmn) = if kdefl != 0 && kdefl % kexsh == 0 {
```
alexhroom, line 163: can you add comments on what you're doing here?

alexhroom (changes requested):
thanks for this! just a couple requests for more comments...

Toucan4Life:
Just added more comments, my bad I always forgot them. I will not resolve the conversations, I'll let you be the judge if they are good enough

src/linalg/schur.rs
```diff
@@ -135,6 +135,16 @@ where
 
         // Implicit double-shift QR method.
         let mut niter = 0;
+        // Number of QR steps since the active window last shrank due to a deflation.
+        let mut kdefl = 0usize;
+        // Apply Wilkinson's exceptional shift every kexsh stalled QR steps.
+        let kexsh = 10usize;
```
alexhroom: i'd prefer
```suggestion
        // Number of QR steps since the active window last shrank due to a deflation.
        let mut kdefl: usize = 0;
        // Apply Wilkinson's exceptional shift every kexsh stalled QR steps.
        let kexsh: usize = 10;
```
unless there's some reason for the way you're doing it that i've missed

alexhroom (approved):
looks good to me! thanks for making the changes.

PR #1609, opened 2026-06-30
feat: add component wise power and log10

Kempeter:
I don't know if it's the right way to do it.

New:
- component-wise abs test
- component-wise power function and tests
- component-wise log10 function and tests

### DIFF
--- a/src/base/componentwise.rs
+++ b/src/base/componentwise.rs
@@ -1,6 +1,6 @@
 // Non-conventional component-wise operators.
 
-use num::{Signed, Zero};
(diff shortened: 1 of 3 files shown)


Ralith:
You should use `map` or `apply` to achieve this in your downstream code without requiring changes to nalgebra. We don't need component versions of every operation you might conceivably perform on a real/complex.

Kempeter:
```// TODO: add other operators like component_ln, component_pow, etc. ?```

I added the functions because of this TODO, but I understand your point

PR #1619, opened 2026-09-01
Clarify the docs about which Euler angle convention is used.

de-vri-es:
The current documentation calls the angles "roll", "pitch" and "yaw", but it doesn't explain what those are. Roll, pitch and yaw are conventions from aerospace, but there is no authoritative definition on which axes they correspond to, in which order they should be applied, and whether they are intrinsic or extrinsic rotations.

As a math library that uses X, Y and Z for the principal axes, it is more helpful if the documentation sticks to that convention.

This PR adjust the documentation to explain the angles in terms of X, Y and Z, while still mentioning that the angles are often called roll, pitch and yaw.

The PR also adds code snippets that shows how to go from Euler angles to a rotation without using `from_euler_angles()`. So even if the text explanation isn't clear enough for everyone, the code snippet should make sure there is no ambiguity.


de-vri-es:
Polite ping :) Any thoughts on this?

